PR #1, opened 2022-12-09
chore: remove unnecessary config step for user

khuezy:
The fewer steps, the better for the user.


fwang:
Awesome! Thanks @khuezy. Release in v0.1.0.

khuezy:
Thanks @fwang! Looking forward to the v1.0.0 release!

PR #16, opened 2022-12-27
Set `serverless-http` and `@aws-sdk/client-s3` as peerDependencies.

Shiroy:
These two packages are not specifically required to run open-next.

However, they are required in order for the lambda functions to be bundled correctly. Therefore, it makes sense to require them to be installed as dependencies of the Next.js application we are trying to bundle and not as dependencies of open-next itself.

This enables open-next to be used with stricter package managers such as `pnpm`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@aws-sdk/client-s3": "^3.234.0",
     "@changesets/changelog-github": "^0.4.4",
     "@changesets/cli": "^2.22.0",
     "@tsconfig/node18": "^1.0.1",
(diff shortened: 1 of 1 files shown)


fwang:
Hey @Shiroy, I make some progress on this over the holidays. I didn't want to have the users install packages "OpenNext" uses. I will share what I have later today, just have to iron out some kinks.

Shiroy:
My point is that package managers such as pnpm won't install open next dependencies as the application dependencies.

This prevents the lambda bundler from finding the package required by your lambda functions.

You don't have this issue with npm as everything is merged together.

Let me know if I can help you.

Best,

fwang:
Got makes sense!

OpenNext used to move the adapters to user's app `.open-next/.build` and perform the build there, which break pnpm b/c the two dependencies won't be found as u describe.

I just published v0.4.2. It will do the build inside the OpenNext folder, and write to `.open-next/.build`. This way, it will find the dependencies.

Can you see if v0.4.2 works for ur setup?

^ Here's the pr #20

Shiroy:
@fwang it works on my setup !

I haven't tried the result on Lambda yet, but now functions are bundled ant built correctly.

PR #30, opened 2023-01-15
fix: delete request body from server adapter

khuezy:
This PR fixes some issues:

1) Nextjs doesn't parse the body if it already exists on the request object: https://github.com/vercel/next.js/blob/canary/packages/next/src/server/api-utils/node.ts#L474

2) If a header has a space it in, the custom header ends up w/ `null` fields

3) `compress: false` has been fixed in the latest nextjs version.

EDIT: Do not merge yet

### DIFF
--- a/src/adapters/middleware-adapter.ts
+++ b/src/adapters/middleware-adapter.ts
@@ -67,6 +67,7 @@ function getMiddlewareRequestHeaders(response: any) {
   const headers: Record<string, string> = {};
   (response.headers.get("x-middleware-override-headers") || "")
     .split(",")
+    .filter(v => v)
     .forEach((key: string) => {
       headers[key] = response.headers.get(`x-middleware-request-${key}`)
     });
--- a/src/adapters/middleware-adapter.ts
+++ b/src/adapters/middleware-adapter.ts
@@ -67,7 +67,7 @@ function getMiddlewareRequestHeaders(response: any) {
   const headers: Record<string, string> = {};
   (response.headers.get("x-middleware-override-headers") || "")
     .split(",")
-    .filter(v => v)
+    .filter(Boolean)
     .forEach((key: string) => {
       headers[key] = response.headers.get(`x-middleware-request-${key}`)
     });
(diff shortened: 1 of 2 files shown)


timothymiller:
Hey khuezy,

This issue is affecting me as well. I would love to use next-auth with open-next.

How can I help?

timothymiller:
I would like to add I am getting a 'crypto is not defined' error when using next-auth middleware to provide per page authentication.

I am deploying with node 18.13.0, next-auth 4.18.6, sst 2.0.0-rc.41, and next 13.1.3

My middleware.ts code looks like this:

```ts
export { default } from "next-auth/middleware" // require authentication

export const config = { matcher: ["/:path*"] } // all pages
```

khuezy:
Another person on another ticket was experiencing the same crypto error as you... `crypto` comes w/ nodejs so that error is confusing.

timothymiller:
If memory serves me right, middleware functions run in a slimmed down version of node called 'edge runtime'

https://nextjs.org/docs/api-reference/edge-runtime

Could be relevant

khuezy:
As long as your `api/` functions aren't exporting the `config.runtime = 'edge'`, it shouldn't be using it.

khuezy:
Are you doing `pnpx sst deploy`?

timothymiller:
Yes. I am doing `pnpx sst deploy`

I am not exporting `config.runtime = 'edge'` either

khuezy:
Do `npx sst deploy` instead. I don't know why `pnpx` is causing issues.

timothymiller:
Doing that now. @khuezy That should fix the crypto issue?

khuezy:
Another person had the same issue and resolved it using `npx`

khuezy:
@timothymiller [related discussion](https://github.com/vercel/next.js/discussions/45384)

timothymiller:
I tried running the next-auth checks in `getServerSideProps` in place of middleware...

My code works on Vercel, but not with SST. The client does not redirect to Cognito as expected.

Can provide a code example and versions if necessary.

khuezy:
Did you monkeypatch the changes in this PR before deploying?

fwang:
Merging into master.

I reverted the `compress` setting as CloudFront should already compressed the response. Let's open a separate PR if it is still required.

revmischa:
You don't know what I went through tracking this problem down in cdk-nextjs lol

PR #38, opened 2023-02-02
fix: Fix next-auth middleware not working on Lambda

khuezy:
This PR allows `next-auth/middleware` to properly run in Lambda environment.

When using this w/ NextjsSitev2 you will probably get infinite redirects due to the way the CloudFront Distribution is set up. To get around this, you can make changes to NextjsSite, eg:
NOTE: there's a limit of 25 cache behaviors, so if you need more than that just request a quota upgrade and you should easily get several hundred. 

```typescript
    createCloudFrontDistributionForRegional() {
       ...
        const baseServerBehavior = {
            viewerProtocolPolicy: cloudfront.ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
            origin: new origins.HttpOrigin(Fn.parseDomainName(fnUrl.url)),
            allowedMethods: cloudfront.AllowedMethods.ALLOW_ALL,
            cachedMethods: cloudfront.CachedMethods.CACHE_GET_HEAD_OPTIONS,
            compress: true,

            cachePolicy: cdk?.cachePolicies?.serverRequests ??
                this.createCloudFrontServerCachePolicy(),
        };
        const middlewareBehavior = {
            ...baseServerBehavior,
            edgeLambdas: isMiddlewareEnabled ? [{
                eventType: cloudfront.LambdaEdgeEventType.VIEWER_REQUEST,
                functionVersion: middlewareFn.currentVersion,
                includeBody: true
            }] : undefined
            }
        const serverBehavior = {
            ...baseServerBehavior
        }
        const middlewareRoutes = this.getMiddlewareRoutes(middlewareBehavior)
       ...
       additionalBehaviors: {
                ...middlewareRoutes,
                "_next/data/*": serverBehavior,
                "_next/image*": imageBehavior,
                "_next/*": staticFileBehaviour,
                ...(cfDistributionProps.additionalBehaviors || {}),
            },
    }
    // create middleware behavior only for the matcher defined in middleware.ts
    getMiddlewareRoutes(behavior) {
        const middlewareManifest = path.join(this.props.path, '.next/server/middleware-manifest.json')
        const content = fs.readFileSync(middlewareManifest).toString('utf8')
        const obj = JSON.parse(content)
        const matcher = obj.middleware['/'].matchers
        const routes = {}
    
        matcher.forEach((reg) => {
          let path = reg.regexp
            .replace('^(?:\\/(_next\\/data\\/[^/]{1,}))?', '')
            .replace('(.json)?[\\/#\\?]?$', '')
            .replace('(?:\\/((?:[^\\/#\\?]+?)(?:\\/(?:[^\\/#\\?]+?))*))?', '/*')
            .replaceAll('\\', '')
            .replace(/^\//, '')
          routes[path] = behavior
        })
    
        return routes
    }
```

Or you can not use the `default` middleware and read the token manually:
NOTE: the current NextjsSite distribution will route *all* requests to middleware... this should be fixed in the future per logic above.

This will guard all your `/api` endpoints:
```typescript
import { NextRequest, NextResponse } from 'next/server'
import { getToken } from 'next-auth/jwt'

export async function middleware(req: NextRequest) {
     const pathname = req.nextUrl.pathname
     // /api/auth shouldn't be restricted or else user will never be able to login
     if (!pathname.startsWith('/api/auth/') && pathname.startsWith('/api/')) { // add your own path checks
       const token = await getToken({ req })
       if (!token) {
         return new NextResponse('Unauthorized', {
           status: 401,
           headers: { 'content-type': 'text/plain' }
         })
      }
    }
}
```

### DIFF
--- a/src/adapters/middleware-adapter.ts
+++ b/src/adapters/middleware-adapter.ts
@@ -1,13 +1,15 @@
 import type {
   CloudFrontRequestEvent,
   CloudFrontRequestResult,
-  CloudFrontHeaders
-} from "aws-lambda"
+  CloudFrontHeaders,
+} from "aws-lambda";
 
 // @ts-ignore
 const index = await (() => import("./middleware.js"))();
 
-export async function handler(event: CloudFrontRequestEvent): Promise<CloudFrontRequestResult> {
+export async function handler(
+  event: CloudFrontRequestEvent
+): Promise<CloudFrontRequestResult> {
   const request = event.Records[0].cf.request;
   const { uri, method, headers, querystring, body } = request;
   console.log(uri);
@@ -20,7 +22,7 @@ export async function handler(event: CloudFrontRequestEvent): Promise<CloudFront
   for (const [key, values] of Object.entries(headers)) {
     for (const { value } of values) {
       if (value) {
-        requestHeaders.append(key, value)
+        requestHeaders.append(key, value);
       }
     }
   }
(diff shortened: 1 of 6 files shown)


src/build.ts
```diff
@@ -306,6 +310,7 @@ function esbuildSync(options: BuildOptions) {
     format: "esm",
     platform: "node",
     bundle: true,
+    minify: true,
```
khuezy, line 313: This saves up to 60% of size

src/adapters/middleware-adapter.ts
```diff
@@ -57,8 +57,14 @@ export async function handler(event: CloudFrontRequestEvent): Promise<CloudFront
     return request;
   }
 
+  // In the case where `middleware.ts` response w/ a body, send that back to the client
```
khuezy, line 60: This may be useful to have if your middleware.ts returns data immediately to the client.

revmischa:
Just my opinion but having to request quota increases on every AWS account you want to deploy a small-medium sized nextjs project is not a great experience

khuezy:
Yes I agree, but we have to eat the 💩 sandwich due to the default limitation.

timothymiller:
🚀🚀🚀🚀 LFO Ship it!!

src/adapters/middleware-adapter.ts
```diff
@@ -57,8 +57,14 @@ export async function handler(event: CloudFrontRequestEvent): Promise<CloudFront
...
+  let responseBody
+  const state = Object.getOwnPropertySymbols(response).find(s => s.toString() === 'Symbol(state)')
```
revmischa, line 62: is this symbol not exported?

khuezy: For some reason, the symbol wasn't create via `Symbol.for('state')` so I couldn't access it by key.

src/adapters/middleware-adapter.ts
```diff
@@ -40,31 +42,30 @@ export async function handler(event: CloudFrontRequestEvent): Promise<CloudFront
...
 
-  // In the case where `middleware.ts` response w/ a body, send that back to the client
-  let responseBody
-  const state = Object.getOwnPropertySymbols(response).find(s => s.toString() === 'Symbol(state)')
-  if (state) responseBody = response[state]?.body?.source
-
   return {
     status: response.status,
-    body: responseBody,
+    body: response.body ? await response.text() : undefined,
```
khuezy, line 68: 👍

src/adapters/middleware-adapter.ts
khuezy: turns out it was much simpler to use `response.text()` to get the body.

src/adapters/middleware-adapter.ts
fwang: yeah.. I will make a separate PR for that today. I want to add a DEBUG flag, and when u do `DEBUG=true open-next build`, it will `console.log`.

PR #40, opened 2023-02-06
feat: optimize server lambda size

khuezy:
This PR optimizes the Server lambda size. For my use case, my WIP app totaled ~40MB in size. Removing the dev dependency reduced the size by ~20MB. Minifying the app itself saved ~60%.

40MB => 20MB => 13MB.

Lambda zipped size went from ~10MB to ~3MB.

The total reduction will depend on the size of your app, but 20MB in fixed savings and 60% app size reduction is significant.

NOTE: I've deployed my app w/ these changes and everything appears to be working. I'm not sure if some code somewhere is using the dev dependencies at runtime. Specifically the webassemblyjs module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "@tsconfig/node18": "^1.0.1",
     "@vercel/build-utils": "^5.7.0",
     "@vercel/next": "^3.3.2",
+    "minify-all-js": "0.1.9",
     "esbuild": "^0.15.18",
     "serverless-http": "^3.1.0",
     "yargs": "^17.6.2"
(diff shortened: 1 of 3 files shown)


fwang:
Released in v1.1.0

@khuezy The PR was forked a while back, and now has a few  merge conflicts. I manually copied over the changes to main, with a few changes:
- Only minify if the `OPEN_NEXT_MINIFY` env var is set (will make it the default once it's tested to be stable) https://github.com/serverless-stack/open-next#open_next_minimize
- I didn't find packages like `esbuild` inside server function's `node_modules`, so I didn't copy over the `removeNodeModules()` function
- `minify-all-js` prints out debug messages. I had to copied over the code and comment out the prints.

PR #43, opened 2023-02-10
Support redirects/rewrites in next.config.js

salvatore-fxpig:
This PR intends to add support to rewrites/redirects in next.config.js

It also handle rewrites from middleware (which does not seem to be really handled currently?)

Caveats: 
- I didn't test it for methods other than GET (but will do soon!)
- I added some support types here and there
- I added a "prepare" script to be able to handily import the lib from git repos

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   "devDependencies": {
     "@types/aws-lambda": "^8.10.109",
     "@types/node": "^18.11.17",
+    "node-fetch": "^3.3.0",
     "typescript": "^4.9.3"
   },
   "bugs": {
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   "scripts": {
     "build": "tsc",
     "dev": "tsc -w",
+    "prepare": "tsc",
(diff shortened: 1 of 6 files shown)


fwang:
Hey @salvatore-fxpig, thanks for putting this together! 

Is this still WIP?

salvatore-fxpig:
Hello @fwang after some tweaks I've tested it out a bit and it works.

I wanted to make time to handle the whole set of rewrite configurations by next.js but it's quite a bit beyond the scope of the project I'm using this for, and I don't think I will have a lot of time to spend on it in the near future, so I would think it's pretty GTG.

fwang:
Hey @salvatore-fxpig, thanks for contribution! The folks in the community ran into quite a few issues caused by middleware and the `minimalMode` flag, including rewrite and redirect. We ended rearchitected how middleware is deployed in v0.8.0. Details are shared here — https://github.com/serverless-stack/open-next#will-my-nextjs-app-behave-the-same-as-it-does-on-vercel

It's a lot more robust now. Give v0.9.0 a try!

PR #49, opened 2023-03-03
CLI install command and disable minimal mode (rewrites and redirects works)

albertehc:
Workaround to send a custom install command (yarn workspaces focus --production) for example and to disable minimalMode to get redirects and rewrites working

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,10 @@ OpenNext aims to support all Next.js 13 features. Some features are work in prog
    ```bash
    npx open-next@latest build
    ```
+  
+   CLI Options:
+   installCommand: Send a custom install command, For example: yarn workspaces focus --all --production
+   disableMinimalMode: Disable minimal mode
 
    This will generate an `.open-next` directory with the following bundles:
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ OpenNext aims to support all Next.js 13 features. Some features are work in prog
    ```
   
    CLI Options:
-   installCommand: Send a custom install command, For example: yarn workspaces focus --all --production
+   installCommand: Send a custom install command, For example: yarn workspaces focus --production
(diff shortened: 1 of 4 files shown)


revmischa:
Honestly I would make disabling middleware in lambda@edge (if minimalMode is disabled) the default for reasons I elaborated on discord
https://discord.com/channels/983865673656705025/1027265626085019769/1080924408824209470

khuezy:
Yea I'm on board w/ having the middleware be handled by nextjs as the default behavior.

khuezy:
Since middleware is running w/in the nextjs server, it needs to set the hostname of itself. Please add:
```
const requestHandler = new NextServer.default({
   hostname: "localhost",
   port: Number(process.env.PORT) || 3000,
```

jdpst:
Is anything blocking this from being merged in soonish? Without this we are blocked from going to prod. We'd either have to (1) use middleware or (2) rewrite all our redirects/rewrites using `getServerSideProps`, neither of which we want to do. 

FWIW, I agree with @revmischa that middleware/minimalMode should be opt-in since the current default is broken if you don't have middleware set up and have a bunch of stuff in your `next.config.js`–not a very good experience.

@albertehc @khuezy @fwang

khuezy:
I would suggest you apply a pnpm patch on the open-next package. You only need to edit 2 lines
1) set flag to fair
2) remove the function call to generate the middleware folder

jdpst:
We ended up forking and proceeding with these changes. 

One thing to note about these changes: the way the options are passed is quite unusual. It took me a few minutes to realise that it's just `disableMinimalMode` and not `--disableMinimalMode`. Was there a reason that you departed from the usual `--flag` syntax? @albertehc

revmischa (approved):
I really need this!

PR #54, opened 2023-03-13
Add hotfix for avoiding relative paths

Townsheriff:
Hi,

I'm setting up appDir feature with open-next and serverless lambda deployments.

I have encountered multiple issues for setting things up.

1. When using open-next and lambda `req.url` is a relative url, but next.js have prevents from using them therefore it requires absolute path.

2. When response status code is not present in appDir + route.ts, there is no default value set for statusCode, therefore making lambda return incorrect response.

This PR serves only as reference for issues I've encountered, it is not desired way of implementation.

PS.

While having full endpoint path solves "not having relative path", next.js fails to handle absolute path (what a coincidence). [Here](https://github.com/Townsheriff/next.js/commit/78b16ba8af70f0c6b8f54a0433f82a22b710a34e) is my current solution to make things work.

Cheers,
Arturs

### DIFF
--- a/src/adapters/server-adapter.ts
+++ b/src/adapters/server-adapter.ts
@@ -37,6 +37,13 @@ const requestHandler = new NextServer.default({
 // Create a HTTP server invoking the NextServer
 const server = slsHttp(
   async (req: IncomingMessage, res: ServerResponse) => {
+
+    if (process.env.OPEN_NEXT_REQ_BASE_URL) {
+      // https://google.com + /nextjs/route
+      req.url = `${process.env.OPEN_NEXT_REQ_BASE_URL}${req.url}`;
+    }
+    
+
     await requestHandler(req, res).catch((e: any) => {
       console.error("NextJS request failed.");
       console.error(e);
@@ -99,7 +106,10 @@ export async function handler(
   );
   response.headers = { ...response.headers, ...middlewareResponseHeaders };
 
-  debug("response headers", response.headers);
+  // new beta appDir does not set statusCode and lambda fails to correctly send response
+  if (!response.statusCode) {
+    response.statusCode = 200;
+  }
 
   return response;
 }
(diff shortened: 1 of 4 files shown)


src/adapters/server-adapter.ts
```diff
@@ -99,7 +106,10 @@ export async function handler(
   );
   response.headers = { ...response.headers, ...middlewareResponseHeaders };
 
-  debug("response headers", response.headers);
+  // new beta appDir does not set statusCode and lambda fails to correctly send response
```
khuezy, line 109: This appears to be a bug w/ app dir routes. The root route has the statusCode set but all nested routes are undefined.

khuezy:
This is an unfortunately situation w/ open-next's architecture. Overall, it's a good one but it leads to feature deviation from `next server`. By setting `minimalMode` to `true`, features like redirect/rewrite, middleware, app routes, etc... are broken or unsupported.
There's been discussion on the discord but Frank has been busy w/ the main `sst`.

Townsheriff:
Also I had noticed that `<Link></Link>` from next.js was preloading all the pages, but when I click on the link it would load again the page, instead of just replacing the view.

When disabling minimalMode this issue goes away.

Townsheriff:
@khuezy I added the changes as you described, let me know if they are correct, if so I will clean up and make PR for them.

khuezy:
@Townsheriff  the changes are correct, but there's a PR that adds an option to configure that here: https://github.com/serverless-stack/open-next/pull/49.

We're waiting for Frank to respond and make the proper changes.

As for the `Link` issue, `minimalMode: true` breaks the prefetch functionality of the app, ie it would hard refresh on navigation.

Ultimately though, I (and others) believe that setting `minimalMode` to false and removing the `middleware` is the correct decision. Next Server handling the middleware solves a lot of problems that the original design created.

PS: if you're using `next/link`, be careful as it will make the network call on component mount... potentially DDOSing your server and overloading the device CPU if you have many links on screen (nav bar or grid view)

src/adapters/server-adapter.ts
khuezy: might be fixed: https://github.com/vercel/next.js/pull/47096

PR #61, opened 2023-03-27
fixes import path for server function's real handler in monorepo with nested folders

rwlanup:
### Context
For monorepo with nested folders in windows operating system, the dummy `index.mjs` that re-exports the real handler has wrong export path.

In monorepo with nested folders, we might have configuration like:
`monorepoRoot` : `D:\\my-app`
`appPath`: `D:\\my-app\\packages\\frontend`
This will result in `packagePath` = `packages\frontend` and output following code for `index.mjs` in windows machine:
```
export * from "./packages\frontend/index.mjs
```
This should be:
```
export * from "./packages/frontend/index.mjs
```

### Fixes
Type: **BUG**
This code converts any path separator of windows and converts to posix.
Solves: https://github.com/serverless-stack/open-next/issues/60


fwang:
Thanks @rwlanup! Merged.

PR #62, opened 2023-03-29
Remove minimal mode

fwang:
Closes #49


packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -0,0 +1,262 @@
...
+    body: parser.body,
+    remoteAddress: parser.remoteAddress,
+  });
+  const res = new ServerResponse({ method: req.method });
+
+  // Process Next.js request
+  await processRequest(req, res);
+
+  // Format Next.js response to Lambda response
+  const statusCode = res.statusCode;
```
khuezy, line 146:
```suggestion
  const statusCode = res.statusCode ?? 200;
```

Nextjs app dir route does not return status code for some reason (it only returns status code of the root `app/route.ts`)

packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -0,0 +1,271 @@
...
+  headers: rawHeaders,
+  body,
+  isBase64Encoded,
+}: {
+  statusCode: number;
+  headers: Record<string, string | string[]>;
+  body: string;
+  isBase64Encoded: boolean;
+}) {
+  const headers: CloudFrontHeaders = {};
```
khuezy, line 253:
```suggestion
  const headers: CloudFrontHeaders = {};
  delete rawHeaders["content-length"]; // CF throws 502 if "content-length" is in headers
```

packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -0,0 +1,271 @@
...
+  const req = new IncomingMessage(reqProps);
+  const res = new ServerResponse({ method: reqProps.method });
+
+  // Process Next.js request
+  await processRequest(req, res);
+
+  // Format Next.js response to Lambda response
+  const statusCode = res.statusCode || 200;
+  const headers = ServerResponse.headers(res);
+  const isBase64Encoded = isBinaryContentType(headers["content-type"]);
```
khuezy, line 152:
```suggestion
  const isBase64Encoded = isBinaryContentType(Array.isArray(headers["content-type"]) ? headers["content-type"][0] : headers["content-type"]);
```

packages/open-next/src/adapters/server-adapter.ts
jdpst: Do you have a source for the comment? The CF docs explicitly recommend adding a content-length header:

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html

> We recommend that you configure your HTTP server to add a Content-Length header to prevent CloudFront from caching partial objects.

khuezy: I wish I did. Maybe the correct approach is the remove the `content-length` on specific request types? All my api requests (including next-auth) would throw a 502 error. I didn't research further on why that was happening though.

fwang: I read it here — https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/edge-functions-restrictions.html#function-restrictions-read-only-headers-origin-request

Issue #65, opened 2023-04-03
Image Optimization Lambda: Allow for defining a bucket key prefix


pinarruiz:
Something like an environment variable like `BUCKET_KEY_PREFIX` that will make the function search on `/process.env.BUCKET_KEY_PREFIX/...` if defined.

This allows for more complex architectures and rollover of the public folder.

pinarruiz:
I did not explain the favicon correctly, I have all the assets under `public/assets` so that will be `/assets` on the bucket, and then we configure it with the html tag and specify it is on `/assets/favicon.ico`, I do this so i do not need to set up any fallback and the assets all come from the same path, with the origin configured with a path that matches the prefix so both `/assets` and `/_next/static/*` get answered correctly.

fwang:
Ah I see, getting the deployments to be more atomic.

The origin path is only configured for the `/assets` and `/_next/static/*` behaviours, not the ssr/api/image routes right?

It'd be great if you can share your CloudFront configuration here. I think many team would be attracted to this architecture.

pinarruiz:
> The origin path is only configured for the S3 origin (ie. /assets and /_next/static/*), not the URL origins (ie. ssr/api/image) right?

Yeah, only configured on the S3 origin, the behaviours are as expected.

> It'd be great if you can share your CloudFront configuration here. I will keep this issue around. I'm sure many teams would be attracted to this architecture.

Sure, I use a python script to update the Cloudfront configuration just because I want some extra actions to be done and i found it easier this way, regardless of that, the configuration looks like:

```yaml
CloudFrontDistribution:
  Type: AWS::CloudFront::Distribution
  Properties:
    DistributionConfig:
      Enabled: true
      HttpVersion: "http2"
      DefaultCacheBehavior:
        TargetOriginId: "serverlambda"
        ViewerProtocolPolicy: "redirect-to-https"
        Compress: true
        CachePolicyId: ...
        AllowedMethods: ...
      CacheBehaviors:
        - TargetOriginId: "assetsbucket"
          PathPattern: "/_next/static/*"
          ViewerProtocolPolicy: "redirect-to-https"
          Compress: true
          CachePolicyId: ...
          AllowedMethods: ...

        - TargetOriginId: "imagelambda"
          PathPattern: "/_next/image"
          ViewerProtocolPolicy: "redirect-to-https"
          Compress: true
          CachePolicyId: ...
          AllowedMethods: ...

        - TargetOriginId: "serverlambda"
[...]

PR #66, opened 2023-04-04
Image Optimization Lambda: Add bucket key prefix support

pinarruiz:
Adds an environment variable (`BUCKET_KEY_PREFIX`) that enables a bucket key prefix for the image lambda.

When unset, it does not modify the key and when set, takes care of all the `/` that need to be there or dissapear for it to be a valid S3 key.

Edit: Related to #65.


tanvirstreame (approved):

pinarruiz:
Hey @fwang should I add a changeset file?

jayair:
@pinarruiz you might've already, but ping Frank in the #open-next channel?

fwang:
Added in v2.0.0

PR #81, opened 2023-04-12
open-next build no longer fail on missing public folder

edvinerikson:
This makes it possible to run open-next without a public folder created

### DIFF
--- a/.changeset/spicy-moose-brush.md
+++ b/.changeset/spicy-moose-brush.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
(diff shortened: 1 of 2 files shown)


edvinerikson:
Awesome! I think eventually we will have a public folder. But a use-case would be a next server that doesn't have anything except api routes (like graphql server for native app).

This fix makes the behavior aligned with next build/vercel deployment and I think that's good for consistency reasons. 

Apart from that the reason I discovered this was that I tried the open-next PR available for cdk-nextjs-standalone project and run into this.

PR #82, opened 2023-04-12
WIP support v1 format

edvinerikson:
Add support for payload format v1 which is used by api gateway rest apis. 

This is still WIP, but I'm open to feedback on the approach.

### DIFF
--- a/.changeset/kind-bananas-poke.md
+++ b/.changeset/kind-bananas-poke.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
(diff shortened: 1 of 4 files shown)


fwang:
Hey @edvinerikson, what's ur use case for hooking up server function to API Gateway REST API, and using v1 payload?

fwang:
Merged. Love the `event-mapper` lib, and the naming of `convertTo` and `convertFrom` are super intuitive ❤️

PR #101, opened 2023-04-26
Update event-mapper.ts

edvinerikson:
v1 event contain all params in multiValueQueryStringParameters. Atm there is a bug where all params get duplicated in when sent to next.

eg an input path to the api gw: `/?step=1` becomes `/step=1&step=1` after we reconstruct the query params from the v1 event. The solution in this PR i to just check the multiValueQueryStringParameters object as that one will be populated even in cases where there isn't any multi-values.


fwang:
Thanks @edvinerikson! Released in v1.3.3

PR #102, opened 2023-04-27
Allow custom build command

pinarruiz:
Allow for custom build commands to be specified. `package.json`'s `build` keeps being the default in case the custom one is empty.

New usage would be `open-next build [custom build command]`.

For example, if we have the `customBuild` command in our `package.json`, we will run OpenNext like `open-next build customBuild`.

I have created the `BuildArguments` type so new arguments can be added down the line.

### DIFF
--- a/.changeset/tricky-beds-live.md
+++ b/.changeset/tricky-beds-live.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+Add custom build commands
--- a/README.md
+++ b/README.md
@@ -43,7 +43,9 @@ When calling `open-next build`, OpenNext **runs `next build`** to build the Next
 
 #### Building the Next.js app
 
-OpenNext runs the `build` script in your `package.json` file. Depending on the lock file found in the app, the corresponding packager manager will be used. Either `npm run build`, `yarn build`, or `pnpm build` will be run.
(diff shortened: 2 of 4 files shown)


pinarruiz:
> @pinarruiz would it work if OpenNext can be called programmatically? And you can do something like this:
> 
> ```ts
> import { build } from "open-next";
> 
> await build({
>   path: ".",
>   command: "pnpm build",
>   esbuild: {
>     external: ["pg"]
>   },
> });
> ```


Hey @fwang, i have just added an `appPath` argument, using:

```ts
import { build } from "open-next/dist/build.js";

await build({
  appPath: "./example/",
  buildCommand: "build",
});
```

will start the build of the example project, however I am still working on it as it cannot generate de `.next/standalone` directory, since I have no clue why this is happening some help would be appreciated.

fwang:
Hey @pinarruiz, in v1.4.0, you can override the build command like this:
```bash
npx open-next build --build-command "npm run custom:build"
```

More details here — https://github.com/serverless-stack/open-next#custom-build-command

---

Appreciate the PR! The change was made based on ur PR.. but I thought it might be easier to copy it over to a new pr instead of merging the conflict :)

pinarruiz:
Hi @fwang could you link the pr? I just want to see the implementation of it. Thanks!

PR #106, opened 2023-05-04
fix: fix next not using prebundled react library

khuezy:
https://github.com/vercel/next.js/pull/49168/files#diff-7550a61853cf6a2d455d1613bd428eb42c6fc2057ce87e8ba7adab4094abcb6e

Next 13.1 and below had a bug where it would use the precompiled next library. 13.2+ fixes it so open-next needs to explicitly set the `NEXT_PRIVATE_PREBUNDLED_REACT` flat to `"next"`

### DIFF
--- a/packages/open-next/src/adapters/server-adapter.ts
+++ b/packages/open-next/src/adapters/server-adapter.ts
@@ -30,8 +30,9 @@ debug({ nextDir });
 const requestHandler = new NextServer.default({
   hostname: "localhost",
   port: Number(process.env.PORT) || 3000,
-  conf: config,
+  conf: {...config, compress: false},
   customServer: false,
+  dev: false,
   dir: __dirname,
 }).getRequestHandler();
 
--- a/packages/open-next/src/adapters/server-adapter.ts
+++ b/packages/open-next/src/adapters/server-adapter.ts
@@ -15,16 +15,14 @@ import { debug } from "./logger.js";
 import type { PublicFiles } from "../build.js";
 import { convertFrom, convertTo } from "./event-mapper.js";
 
+setNodeEnv();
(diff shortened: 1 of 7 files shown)


packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -28,11 +30,8 @@ debug({ nextDir });
 const requestHandler = new NextServer.default({
   hostname: "localhost",
   port: Number(process.env.PORT) || 3000,
-  // Next.js compression should be disabled because of a bug in the bundled
-  // `compression` package — https://github.com/vercel/next.js/issues/11669
-  conf: { ...config, compress: false },
+  conf: config,
   customServer: false,
-  dev: false,
```
khuezy, line 35: Need to verify removing `dev` is okay.

packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -28,11 +30,8 @@ debug({ nextDir });
 const requestHandler = new NextServer.default({
   hostname: "localhost",
   port: Number(process.env.PORT) || 3000,
-  // Next.js compression should be disabled because of a bug in the bundled
-  // `compression` package — https://github.com/vercel/next.js/issues/11669
-  conf: { ...config, compress: false },
```
khuezy, line 33: compress appears to be fixed now.

khuezy: I lied, it doesn't work :(

khuezy: Actually I might be correct. The error was from latest 1.2.0 version.

fwang:
Fixed in v1.2.1

PR #108, opened 2023-05-08
fixing the spelling of adapter, relies and determines

aballerr:
fixing the spelling of adapter, relies and determines


fwang:
Ah nice catch! Thanks @aballerr

Issue #63, opened 2023-03-29
runs 'yarn run build' during build


theAndrewCline:
Not sure if this is a feature request or a question, but it would be awesome to be able to customize the yarn script that is run during the open-next build process. 

My use case is I need to build other things and not just next during the build process but I need open next to build first.

martg0:
+1 for this, I have two builds separated by site, **npm run build:lat** and **npm run build:bra** for each deploy. 
Thanks!

alebelcor:
I have a somewhat similar request.

Right now the `build` command (`open-next build`)—generally speaking—does two things:

1. Creates a Next.js production build by running either `npm run build`, `yarn build` or `pnpm build`
2. Creates and bundles the server function, image optimization function and assets

I'd love a way to skip step 1.

I think if the `build` command instead assumes that a Next.js build is present, instead of force-building one, it would solve all of our use cases. It could also be a new/different command, CLI flag, env var, or other.

laubonghaudoi:
+1 to this. It would be great if we can keep using yarn instead of switching to npm.

fwang:
In v1.4.0, you can override the build command like this:
```bash
npx open-next build --build-command "npm run custom:script"
```

You can also skip building the Next.js app like this:
```bash
npx open-next build --build-command "echo 'skip build'"
```

More details here — https://github.com/serverless-stack/open-next#custom-build-command

PR #110, opened 2023-05-11
fix: skip next build with environment variable

robertveloso:
This Pull request aims to solve the problem discussed in the issue: #63 

The fix only add a conditional checking for a environment variable. 

I have also expirience in my pipelines another issue that this solution can be aplied as a workaround. For some reason when I run the sst deploy and next build is runned as a child_process from open-next executed from sst deploy, my pipeline hangs in: "Creating an optimized production build..."

Captura de tela 2023-05-11 143816


fwang:
Hey @robertveloso, in v1.4.0, you can skip building the Next.js app by overriding the build command like this:

```bash
npx open-next build --build-command "echo 'skip build'"
```

More details here — https://github.com/serverless-stack/open-next#custom-build-command

PR #113, opened 2023-05-12
Fix: next now use node_modules react for pages and prebundled for app

### LINKED ISSUE #109: Error after upgrading from 1.2.0 to 1.2.1
Hello,

We encoutered this error after upgrading from the version `1.2.0` to `1.2.1`

Here are the logs of the lambda

```
2023-05-10T09:46:48.327Z	d577c19a-2182-4b2e-ad9b-cefe19179b13	ERROR	TypeError: Cannot read properties of null (reading 'useEffect')
    at exports.useEffect (/var/task/node_modules/next/dist/compiled/react/cjs/react.production.min.js:28:292)
    at MyApp (/var/task/.next/server/pages/_app.js:130:34)
```

It seems that it is related to `__NEXT_PRIVATE_PREBUNDLED_REACT`.  Any idea how would we fix the error, we're using `Next.js@13.4.1`?

conico974:
There is an issue with require resolution of import where some external react dependencies could not find the right react version to use. 
To solve this, we followed the approach that [next.js uses to resolve react dependencies](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/require-hook.ts), but we modified it to work on every request instead of only at startup. This way, we ensured that Next always used the correct react version for each dependency.
Fix #109


fwang:
Released in v1.3.2

PR #129, opened 2023-05-26
Fix react dependencies resolution

conico974:
This PR fix 2 issues:
- For users using monorepo without hoisting react at the root of the monorepo, `overrideDefault` was trying to resolve from the root
- `getMaybePagePath` has been replaced by some custom logic since it wasn't working for dynamic routes. (External dependencies could cause issue on these routes)

### DIFF
--- a/packages/open-next/src/adapters/server-adapter.ts
+++ b/packages/open-next/src/adapters/server-adapter.ts
@@ -7,6 +7,12 @@ import type {
   APIGatewayProxyEvent,
   CloudFrontRequestEvent,
 } from "aws-lambda";
+// We need to resolve those react deps before nextjs overrides them with prebundled one in case of app dir
+const node_react: [string, string][] = [
+  ["react", require.resolve(`react`)],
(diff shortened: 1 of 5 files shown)


packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -131,30 +136,47 @@ function initializeNextjsRequireHooks(config: any) {
...
+      // We need to add `/page` to the end of the route pattern because that's how it is in the appPathsManifest
+      routePattern = `${route.page}/page`;
+    }
+  });
+  if (routePattern) return routePattern;
+  routesManifest.dynamicRoutes.forEach((route) => {
+    if (new RegExp(route.regex).test(url)) {
+      routePattern = `${route.page}/page`;
+    }
+  });
```
fwang, line 155: Do we need to also handle `routesManiftest.dataRoutes` here?

conico974: Not useful here, we just check if it's an app and app doesn't have separate data route

PR #131, opened 2023-05-27
fix: group routes react resolution

conico974:
This PR should fix: 
- Next overriding the needed react resolution override
- Fix `isApp` for:
  - Group routes
  - Index page (including index page under group routes)

### DIFF
--- a/packages/open-next/src/adapters/server-adapter.ts
+++ b/packages/open-next/src/adapters/server-adapter.ts
@@ -19,9 +19,8 @@ import { isBinaryContentType } from "./binary.js";
 import { debug } from "./logger.js";
 import { convertFrom, convertTo } from "./event-mapper.js";
 import {
-  overrideDefault,
-  overrideReact,
-  applyOverride,
+  overrideHooks as overrideNextjsRequireHooks,
+  applyOverride as applyNextjsRequireHooksOverride,
 } from "./require-hooks.js";
 import type { WarmerEvent, WarmerResponse } from "./warmer-function.js";
 
@@ -39,15 +38,16 @@ const publicAssets = loadPublicAssets(OPEN_NEXT_DIR);
 // Generate a 6 letter unique server ID
 const serverId = `server-${generateUniqueId()}`;
 
-// Need to override the require hooks for React before Next.js server
-// overrides them with prebundled ones in the case of app dir
+// WORKAROUND: Set `__NEXT_PRIVATE_PREBUNDLED_REACT` to use prebundled React — https://github.com/serverless-stack/open-next#workaround-set-__next_private_prebundled_react-to-use-prebundled-react
+// Step 1: Need to override the require hooks for React before Next.js server
+//         overrides them with prebundled ones in the case of app dir
+// Step 2: Import Next.js server
+// Step 3: Apply the override after Next.js server is imported since the
+//         override that Next.js does is done at import time
 overrideNextjsRequireHooks(config);
-
 // @ts-ignore
 import NextServer from "next/dist/server/next-server.js";
-// We need to apply the override after Next.js server is imported
-// since the override that Next.js does is done at import time
-applyOverride();
+applyNextjsRequireHooksOverride();
 
 const requestHandler = new NextServer.default({
   hostname: "localhost",
@@ -104,7 +104,7 @@ export async function handler(
   debug("IncomingMessage constructor props", reqProps);
   const req = new IncomingMessage(reqProps);
   const res = new ServerResponse({ method: reqProps.method });
-  setNextjsPrebundledReact(req, config);
+  setNextjsPrebundledReact(internalEvent.rawPath, config);
   await processRequest(req, res);
 
   // Format Next.js response to Lambda response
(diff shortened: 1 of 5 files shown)


khuezy (approved):
LGTM! I have deployed and app & pages works. Thanks!

packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -150,13 +158,13 @@ function setNextjsPrebundledReact(req: IncomingMessage, config: any) {
   // WORKAROUND: Set `__NEXT_PRIVATE_PREBUNDLED_REACT` to use prebundled React — https://github.com/serverless-stack/open-next#workaround-set-__next_private_prebundled_react-to-use-prebundled-react
 
   // Get route pattern
-  const route = [
-    ...routesManifest.staticRoutes,
-    ...routesManifest.dynamicRoutes,
-  ].find((route) => new RegExp(route.regex).test(req.url ?? ""));
+  const route = routesManifest.find((route) =>
+    new RegExp(route.regex).test(req.url ?? "")
```
khuezy, line 162: The one thing that I would change is to map all the routes' regex and join it into a single regex at the very top to optimize. Creating a new regex for each route can be computationally expensive.
Perhaps the new js engine is better at regex but I like to minimize them as much as possible.

conico974: I agree that we should compute the regex beforehand. Maybe I'm missing something, but how would you find the route if you merge all the regex at the top?

khuezy: Hmm, it'll be pretty nasty. I think this could work:
1) Loop through `routesManifest` to map the "regex" string AND wrap the regex inside parenthesis, also store its "page" in another array.
2) Join the array from above as: `.join('?|')`, this would give us `(\/pageA)?|(\/pageB)`, etc...
3) Use `req.url.replace(regexFromAbove, function(original, ...args)
4) the "...args" would be an array that correspond to the position of the regex group, which we can use to map its index to the "page" array we saved in step 1.

khuezy: I think `req.url` would need to strip out its domain and grab it's path, since the regex in the manifest has "^"

khuezy: I'll tinker about to see if it works.

conico974: > I think `req.url` would need to strip out its domain and grab its path, since the regex in the manifest has "^"

`req.url` is already without the domain here, it's the rawPath with queryParams: https://github.com/serverless-stack/open-next/blob/7f085fad656b1ebaa37dc1c123b499dbbb2fd6d4/packages/open-next/src/adapters/event-mapper.ts#L97
But you've raised a valid point we should probably use `internalEvent.rawPath` instead even if, for now the regex handles query params. Who knows what vercel decides to break next ? :joy:

conico974: > Hmm, it'll be pretty nasty. I think this could work:
> 
> 1. Loop through `routesManifest` to map the "regex" string AND wrap the regex inside parenthesis, also store its "page" in another array.
> 2. Join the regex array from above as: `.join('|')`, this would give us `(\/pageA)|(\/pageB)`, etc...
> 3. Use `req.url.replace(regexFromAbove, function(original, ...args)`
> 4. the "...args" would be an array that correspond to the position of the regex group, which we can use to map its index to the "page" array we saved in step 1.
> 5. if there's a match, we return that "page" to use in `appPathsManifestKeys.includes(page)`

Is it worth it though ? I mean how likely will this have an impact on performance ? 
Even if you have like 1000 different routes (and every dynamic routes will count only for 1 regex), how many ms will you gain realistically ?

khuezy: It's a good question, I don't honestly don't know. Would be pretty interesting to see the the performance. But since lambdas aren't paid by memory used (rather its allocation), then I don't think it would ultimately be worth it.

khuezy: Feel free to use or not use this:
```
function createRoutesRegex(routes: RouteDefinition[]): [string[], RegExp] {
  const paths: string[] = routes.map(r => r.page)
  const regexes = routes.map(r => `(${r.regex})`).join('|')
  return [paths, new RegExp(regexes)]
}

// create the regex at the top
const [routePages, routeRegex] = createRoutesRegex(routesManifest)

// update the param to use internalEvent
function setNextjsPrebundledReact(internalEvent: InternalEvent, config: any) {
  const page = internalEvent.rawPath.replace(routeRegex, (og, ...args) => {
    const ray = args.slice(0,-2) // args contains a couple of extra data at the end of array
    const i = ray.findIndex(e => !!e)
    return routePages[i]
  })

  const isApp = appPathsManifestKeys.includes(page ?? "");

```

I've deployed this change in my app and it seems to be working... but again, I haven't done any testing to see if this is worth it.

khuezy: I've created some random testing, 100 routes trying to find route50 (half way)

https://jsbench.me/7ili6iogv0/2

In general, creating the regex in the loop was ~20% slower, but creating smaller regex might be faster than 1 big one. But if you create one big one to be cached at the top, then that cost would only happen once.
🤷

conico974: Nice work, will this work reliably for overlapping routes ? Static routes have to take precedence over dynamic ones.

I ran your jsbench and for me it was even slower, like 49% slower, but even in a worst case scenario with 1000 routes and the route is the last one, we are still under 0.5 ms

My personal advice would still be to keep it simple here, i knew what it was supposed to do and i still had a little trouble figuring out what was going on in the `setNextjsPrebundledReact` part.

khuezy: Yes, static routes works for /app and /pages. I agree, this optimization isn't worth it for 99% of cases.

PR #141, opened 2023-06-22
Fix isr not found

conico974:
Fix support for `notFound` and `redirect` in getStaticProps


khuezy (commented):
Nice. Can you replace the `console.log`s with debug please?

PR #143, opened 2023-06-26
fix ISR when used in addition to middleware rewrite

### LINKED ISSUE #142: ISR not working with middleware rewrite
I have come accross than issue when you use ISR in combination with middleware rewrite.
Here is the setup:

in middleware rewrite the page `/middleware-rewrite-with-isr` to -> `/middleware-rewrite-with-isr-destination`
create a page `/middleware-rewrite-with-isr-destination.js` that uses ISR. In that case the page is created once but never revalidated and updated. (see [this commit](https://github.com/serverless-stack/open-next/commit/158c4cb3a6458ae4f9cc3dedb363500f14cfe314))

The issue comes from the way nextjs treats revalidation requests ([docs](https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration#on-demand-revalidation))

> "Middleware won't be executed for On-Demand ISR requests. Instead, call revalidate() on the exact path that you want revalidated. For example, if you have pages/blog/[slug].js and a rewrite from /post-1 -> /blog/post-1, you would need to call res.revalidate('/blog/post-1')."

But Open Next uses the revalidate api to trigger the revalidation of the page ([see code](https://github.com/serverless-stack/open-next/blob/8db7e076a5dc4716b564dee0fe1f3e00bd36d1fd/packages/open-next/src/adapters/revalidate.ts#L31C1-L44C8))

So basically open next is sending a request to revalidate a page `/middleware-rewrite-with-isr` instead of the correct page `/middleware-rewrite-with-isr-destination`

The way to solve this properly would be to check if a request has been rewritten before using the function `revalidateIfRequired` ([see code](https://github.com/serverless-stack/open-next/blob/8db7e076a5dc4716b564dee0fe1f3e00bd36d1fd/packages/open-next/src/adapters/server-adapter.ts#L134C1-L138C5))

instead of 
```typescript
  await revalidateIfRequired(
    internalEvent.headers.host,
    internalEvent.rawPath,
    headers
  );
```
  
  We should have 
```typescript
  await revalidateIfRequired(
    internalEvent.headers.host,
    rewrittenUrl,
    headers
  );
```

Now the tricky thing is I didn't find a good way to get the rewritten URL and the only thing that seems to be accessible is to use 
 `Symbol.for('NextInternalRequestMeta')._nextRewroteUrl` 
 But it's really not great (see that [comment](https://github.com/vercel/next.js/blob/913b365d85011473c0ec595791afa6d924a988c4/packages/next/src/server/request-meta.ts#L10C1-L12C1) saying it should be fixed)

mathisobadia:
this closes #142 

It also updates the example repo but maybe it should not be in that PR and is not strictly necessary

### DIFF
--- a/packages/open-next/src/adapters/server-adapter.ts
+++ b/packages/open-next/src/adapters/server-adapter.ts
@@ -1,4 +1,5 @@
 import path from "node:path";
+import crypto from "node:crypto";
 import type {
   APIGatewayProxyEventV2,
   APIGatewayProxyEvent,
@@ -271,10 +272,12 @@ async function revalidateIfRequired(
   // your page will need to have a different etag to bypass the deduplication window.
   // If data has the same etag during these 5 min dedup window, it will be deduplicated and not revalidated.
   try {
+    const hash = (str: string) => crypto.createHash('md5').update(str).digest('hex')
+
     await sqsClient.send(
       new SendMessageCommand({
         QueueUrl: REVALIDATION_QUEUE_URL,
-        MessageDeduplicationId: `${rawPath}-${headers.etag}`,
+        MessageDeduplicationId: hash(`${rawPath}-${headers.etag}`),
         MessageBody: JSON.stringify({ host, url: rawPath }),
         MessageGroupId: "revalidate",
       })
--- a/packages/open-next/src/adapters/server-adapter.ts
+++ b/packages/open-next/src/adapters/server-adapter.ts
@@ -131,9 +131,22 @@ export async function handler(
 
   fixCacheHeaderForHtmlPages(internalEvent.rawPath, headers);
   fixSWRCacheHeader(headers);
+
+  // we are sending a revalidation request that will bypass the middleware https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration#on-demand-revalidation
+  // so we need to use the rewritten url if it was rewritten
+  let revalidateUrl = internalEvent.rawPath;
+  if (req[Symbol.for("NextInternalRequestMeta")]?._nextDidRewrite) {
+    const rewrittenUrl =
+      req[Symbol.for("NextInternalRequestMeta")]?._nextRewroteUrl;
+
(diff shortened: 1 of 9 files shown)


packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -131,9 +131,22 @@ export async function handler(
...
+
+  // we are sending a revalidation request that will bypass the middleware https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration#on-demand-revalidation
+  // so we need to use the rewritten url if it was rewritten
+  let revalidateUrl = internalEvent.rawPath;
+  if (req[Symbol.for("NextInternalRequestMeta")]?._nextDidRewrite) {
+    const rewrittenUrl =
+      req[Symbol.for("NextInternalRequestMeta")]?._nextRewroteUrl;
+
+    // the rewritten url will does not include the /_next/data/ prefix so we need to add it back
+    revalidateUrl = internalEvent.rawPath.startsWith("/_next/data/")
```
khuezy, line 143: I'm confused by the comment and the code. Shouldn't it be `!internalEvent.rawPath...`? ie if it doesn't start w/ `/_next/data/`, the prepend it.

mathisobadia: No so basically from my testing we have two separate events / requests for each page with ISR:
1. First request we have `internalEvent.rawPath === '/some-path'` and `rewrittenUrl === '/some-path-rewritten'`
2. 2nd Request  we have `internalEvent.rawPath === '/_next/data/buildId/some-path.json'` but the rewrittenUrl is still the same and does not include the _next/data/ thing: `rewrittenUrl === '/some-path-rewritten'`. We could potentially send both requests to validate directly when we receive the first request but to make it consistent with the way things work without revalidation I kept is like that.
I probably can add more comments there to make it more easily understandable

khuezy: gotcha, thanks for the update.

fwang:
Thanks @mathisobadia!

PR #145, opened 2023-06-27
Fix: Add support for revalidateTag

conico974:
Right now, `revalidateTag` and `revalidatePath` from `next/cache` is not working. This PR aims to fix this.
It is still a work in progress

### How it works
- We add a dynamoDB table where all the tags are stored with their associated path and their last revalidated time
- Tags are written at the end of the `set` method of our incremental cache

### Requirements
- A dynamoDB table with the correct permissions for server-function lambda
- A GSI named revalidate
- A `CACHE_DYNAMO_TABLE` env variable 
If you're using sst for the time being you could do 
```typescript
const testTable = new Table(stack, "testCacheTable2", {
  fields: {
    path: "string",
    tag: "string",
    revalidatedAt: "number",
  },
  primaryIndex: { partitionKey: "tag", sortKey: "path" },
  globalIndexes: {
    revalidate: { partitionKey: "path", sortKey: "revalidatedAt" },
  }
});
const site = new NextjsSite(stack, "site", {
  buildCommand: 
    "/path/to/open-next/packages/open-next/dist/index.js build",
  waitForInvalidation: false,
  environment: {
    CACHE_DYNAMO_TABLE: testTable.tableName,
  }
});
site.attachPermissions([testTable]);
```

### Caveats
- ~~It seems that right now `revalidateTag` is not revalidating immediately but just setting the associated page to `STALE`. That's the way it is done on vanilla next.js, might be different on vercel~~
- ~~On write we override the revalidatedAt with `Date.now()` for every tags~~
- The fetch cache is not working right now because our solution for ISR is using on demand revalidate under the hood. Right now fetch cache is disabled for on demand revalidate. (There is a workaround that the user can do, see docs)
- All these caveats might change, and some may appears, as it relies on internal mechanism of next.js and on api that are still unstable

### Potential Improvements
- We could move some part that are reliant on S3 cache right now to dynamoDB (This should be faster)


khuezy:
Things to consider: if the NextjsSite is deployed on the edge, then we'd probably want to replicate the DDB to the same edge to reduce the latency. We'd have to setup latency records to allow the Lambdas to hit the closest DDB (this is pretty easy to setup).

khuezy:
Here's a gist of how to setup the closest DDB lookup: https://gist.github.com/khuezy/ce2a2fd6e6b3e6d435d2b45f00201d28

README.md
```diff
@@ -383,6 +411,19 @@ This cost estimate is based on the `us-east-1` region pricing and does not consi
 
 ## Limitations and workarounds
 
+#### WORKAROUND: Patch fetch behaviour for ISR. Only for next@13.5.1+
+
+In order to make fetch work as expected with ISR, you need to patch the `fetch` function in your app. Just add this lines to your root layout component:
+
+```ts
+const asyncStorage = require("next/dist/client/components/static-generation-async-storage.external");
```
khuezy, line 419: We should include the function here to make it clear that this code should be inside the function, not at the top of the file.

khuezy (approved):

PR #152, opened 2023-07-03
fix: missing html files for non blocking fallback page

conico974:
For pages using `fallback:true`, we need to keep the html files. This PR fix this

### DIFF
--- a/packages/open-next/src/build.ts
+++ b/packages/open-next/src/build.ts
@@ -343,6 +343,7 @@ function createCacheAssets(monorepoRoot: string) {
       file.endsWith(".js.nft.json") ||
       (file.endsWith(".html") && htmlPages.has(file))
   );
+
   // Copy fetch-cache to cache folder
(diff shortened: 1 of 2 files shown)


khuezy (approved):

packages/open-next/src/build.ts
```diff
@@ -507,7 +508,12 @@ function removeCachedPages(outputPath: string, packagePath: string) {
           file.endsWith(".json") ||
           file.endsWith(".rsc") ||
           file.endsWith(".meta") ||
-          (file.endsWith(".html") && !htmlPages.has(file))
+          (file.endsWith(".html") &&
+            // do not remove static HTML files
+            !htmlPages.has(file) &&
+            // do not remove HTML files with "[param].html" format
+            // b/c they are used for "fallback:true" pages
+            !isFallbackTruePage.test(file))
```
fwang, line 516: @conico974 I reworked these few lines. They are less performant than the initial version (okay since this is run at built time), but hopefully a bit more readable.

Will merge once u give green flag.

PR #155, opened 2023-07-06
Fix: react resolution for interception, parallel and group routes

conico974:
Right now with app dir, interception, parallel and group routes doesn't work correctly.
This PR should solve this

### DIFF
--- a/packages/open-next/src/adapters/routing.ts
+++ b/packages/open-next/src/adapters/routing.ts
@@ -123,8 +123,11 @@ export function handleRewrites<T extends RewriteDefinition>(
   }
 
   return {
-    rawPath: rewrittenUrl,
-    url: `${rewrittenUrl}${urlQueryString ? `?${urlQueryString}` : ""}`,
+    internalEvent: {
+      ...internalEvent,
+      rawPath: rewrittenUrl,
+      url: `${rewrittenUrl}${urlQueryString ? `?${urlQueryString}` : ""}`,
+    },
     __rewrite: rewrite,
     isExternalRewrite,
   };
@@ -134,12 +137,15 @@ export function handleRedirects(
   internalEvent: InternalEvent,
   redirects: RedirectDefinition[],
 ) {
-  const { url, __rewrite } = handleRewrites(internalEvent, redirects);
+  const { internalEvent: _internalEvent, __rewrite } = handleRewrites(
+    internalEvent,
+    redirects,
+  );
   if (__rewrite && !__rewrite.internal) {
     return {
       statusCode: __rewrite.statusCode ?? 308,
       headers: {
-        Location: url,
+        Location: _internalEvent.url,
       },
     };
   }
--- a/packages/open-next/src/adapters/routing.ts
+++ b/packages/open-next/src/adapters/routing.ts
@@ -4,8 +4,8 @@ import {
   RedirectDefinition,
   RewriteDefinition,
   RewriteMatcher,
-  RoutesManifest,
 } from './next-types.js';
+import { compile, match } from 'path-to-regexp';
 
 const redirectMatcher =
   (
@@ -16,11 +16,6 @@ const redirectMatcher =
   (redirect: RewriteMatcher) => {
     switch (redirect.type) {
       case 'header':
-        console.log(
-          'redirect-header',
-          headers?.[redirect.key.toLowerCase()],
-          redirect.value
-        );
         return (
           headers?.[redirect.key.toLowerCase()] &&
           new RegExp(redirect.value ?? '').test(
@@ -35,8 +30,9 @@ const redirectMatcher =
       case 'query':
         return query[redirect.key] && Array.isArray(redirect.value)
           ? redirect.value.reduce(
-              (prev, current) => prev || new RegExp(current),
-              true
+              (prev, current) =>
+                prev || new RegExp(current).test(query[redirect.key] as string),
+              false
             )
           : new RegExp(redirect.value ?? '').test(
               (query[redirect.key] as string | undefined) ?? ''
@@ -74,12 +70,21 @@ export function handleRewrites<T extends RewriteDefinition>(
   );
 
   const urlQueryString = new URLSearchParams(query).toString();
+  let rewrittenUrl = rawPath;
   if (rewrite) {
-    debug('rewrite', { rewrite, urlQueryString });
+    const toDestination = compile(rewrite?.destination ?? '');
+    const fromSource = match(rewrite?.source ?? '');
+    const _match = fromSource(rawPath);
+    if (_match) {
+      const { params } = _match;
+      const isUsingParams = Object.keys(params).length > 0;
+      if (isUsingParams) {
+        rewrittenUrl = toDestination(params);
+      }
+      rewrittenUrl = rewrite.destination;
+      debug('rewrittenUrl', rewrittenUrl);
+    }
   }
-  // Do we need to replace anything else here?
-  const rewrittenUrl =
-    rewrite?.destination.replace(/:[^\/]+/g, '[$1]') ?? rawPath;
 
   return {
     rawPath: rewrittenUrl,
--- a/packages/open-next/src/adapters/routing.ts
+++ b/packages/open-next/src/adapters/routing.ts
@@ -6,6 +6,7 @@ import {
   RewriteMatcher,
 } from './next-types.js';
 import { compile, match } from 'path-to-regexp';
+import { escapeRegex, unescapeRegex } from './util.js';
 
 const redirectMatcher =
   (
@@ -72,16 +73,19 @@ export function handleRewrites<T extends RewriteDefinition>(
   const urlQueryString = new URLSearchParams(query).toString();
   let rewrittenUrl = rawPath;
   if (rewrite) {
-    const toDestination = compile(rewrite?.destination ?? '');
-    const fromSource = match(rewrite?.source ?? '');
+    const toDestination = compile(
+      escapeRegex(rewrite?.destination ?? '') ?? ''
+    );
+    const fromSource = match(escapeRegex(rewrite?.source) ?? '');
     const _match = fromSource(rawPath);
     if (_match) {
       const { params } = _match;
       const isUsingParams = Object.keys(params).length > 0;
       if (isUsingParams) {
-        rewrittenUrl = toDestination(params);
+        rewrittenUrl = unescapeRegex(toDestination(params));
+      } else {
+        rewrittenUrl = rewrite.destination;
       }
-      rewrittenUrl = rewrite.destination;
       debug('rewrittenUrl', rewrittenUrl);
     }
   }
--- a/packages/open-next/src/adapters/routing.ts
+++ b/packages/open-next/src/adapters/routing.ts
@@ -1,0 +1,104 @@
+import { InternalEvent } from './event-mapper.js';
+import { debug } from './logger.js';
+import {
+  RedirectDefinition,
+  RewriteDefinition,
+  RewriteMatcher,
+  RoutesManifest,
+} from './next-types.js';
+
+const redirectMatcher =
+  (
+    headers: Record<string, string>,
+    cookies: Record<string, string>,
+    query: Record<string, string | string[]>
+  ) =>
+  (redirect: RewriteMatcher) => {
+    switch (redirect.type) {
+      case 'header':
+        console.log(
+          'redirect-header',
+          headers?.[redirect.key.toLowerCase()],
+          redirect.value
+        );
+        return (
+          headers?.[redirect.key.toLowerCase()] &&
+          new RegExp(redirect.value ?? '').test(
+            headers[redirect.key.toLowerCase()] ?? ''
+          )
+        );
+      case 'cookie':
+        return (
+          cookies?.[redirect.key] &&
+          new RegExp(redirect.value ?? '').test(cookies[redirect.key] ?? '')
+        );
+      case 'query':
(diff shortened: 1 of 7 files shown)


khuezy:
@conico974 This appears to be broken in latest 13.4.10 😢 
https://github.com/vercel/next.js/issues/50243 (see latest posts)

packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -164,16 +165,38 @@ function setBuildIdEnv() {
...
+      route.has[0].type === "header" &&
+      route.has[0].key === "Next-Url" &&
+      new RegExp(route.has[0].value ?? "").test(headers?.["next-url"] ?? "")
   );
+  let route: RouteDefinition | undefined;
+
+  // Get route pattern
+  if (rewrite) {
+    route = {
+      page: rewrite.destination,
```
khuezy, line 188: If you have dynamic routes, the destination is `/:dynamic`, while the routes manifest is `/[dynamic]`, and thus `isApp` will fail and cause 404 errors.

```
rewrite.destination.replace(/:([^\/]+)/g, '[$1]')
```

packages/open-next/src/adapters/routing.ts
```diff
@@ -0,0 +1,104 @@
...
+const redirectMatcher =
+  (
+    headers: Record<string, string>,
+    cookies: Record<string, string>,
+    query: Record<string, string | string[]>
+  ) =>
+  (redirect: RewriteMatcher) => {
+    switch (redirect.type) {
+      case 'header':
+        console.log(
```
khuezy, line 19: debug

packages/open-next/src/adapters/routing.ts
```diff
@@ -0,0 +1,104 @@
...
+          )
+        );
+      case 'cookie':
+        return (
+          cookies?.[redirect.key] &&
+          new RegExp(redirect.value ?? '').test(cookies[redirect.key] ?? '')
+        );
+      case 'query':
+        return query[redirect.key] && Array.isArray(redirect.value)
+          ? redirect.value.reduce(
```
khuezy, line 37: I'm confused by this reduce, `true` is initially passed in, so `true || x` will always be true?

packages/open-next/src/adapters/routing.ts
```diff
@@ -0,0 +1,104 @@
...
+        : true)
+  );
+
+  const urlQueryString = new URLSearchParams(query).toString();
+  if (rewrite) {
+    debug('rewrite', { rewrite, urlQueryString });
+  }
+  // Do we need to replace anything else here?
+  const rewrittenUrl =
+    rewrite?.destination.replace(/:[^\/]+/g, '[$1]') ?? rawPath;
```
khuezy, line 82: I think that should be good, I've checked the Vercel deployed app and they contain the `(.)` in the route (so we don't need to replace the groups; they may change that in the future for all we know)

khuezy (commented):
Great work! Fixes a few issues others have reported w/ the query stuff too 👍 
I'll test this in a couple hours.

packages/open-next/src/adapters/routing.ts
khuezy, line 82:
```suggestion
    rewrite?.destination.replace(/:([^\/]+)/g, '[$1]') ?? rawPath;
```
You forgot the grouping `()`, otherwise, it'll replace the word with a literal `$1`

conico974: Yeah i noticed that, testing this right now

packages/open-next/src/adapters/routing.ts
conico974: You're right, just a silly mistake

packages/open-next/src/adapters/routing.ts
conico974: This wasn't working finally, we need to replace the `rawPath` here, i used https://github.com/pillarjs/path-to-regexp. This is the lib they use internally inside next

khuezy: Hmm I'm getting 502's on prefetch now

conico974: It's likely that it comes from here https://github.com/conico974/open-next/blob/b3060a37960e4a1305fe82d125d5003c7e0897be/packages/open-next/src/adapters/routing.ts#L74-L85. Could you try adding some log here to see what could be the cause

khuezy: Yup I'll do that as soon as I get back from shopping.

khuezy: I added a bunch of logs after each line and it appears to crash here: https://github.com/conico974/open-next/blob/b3060a37960e4a1305fe82d125d5003c7e0897be/packages/open-next/src/adapters/routing.ts#L82C8-L82C8

conico974: We need the interception path here otherwise it will just redirect, but looking at the error, it might be this https://github.com/vercel/next.js/blob/da043b5535bf50f699c928d26052aba07cb243bc/packages/next/src/shared/lib/router/utils/path-match.ts#L65. Could you look what your params look like, and if they have number keys, maybe try to remove them

khuezy: I updated my comment w/ additional info.

conico974: I should have read the docs a little more... https://github.com/pillarjs/path-to-regexp#unnamed-parameters. So in `compile` it treats `(.)` as an unnamed parameter, and so it expects it to be present when we use it in `toDestination`. see https://github.com/pillarjs/path-to-regexp/blob/1cbb9f3d9c3bff97298ec45b1bb2b0beb879babf/src/index.spec.ts#L2892
We actually need the interception path (i.e. `(.)|(..)|(...)`) to be returned from our function so that it return the correct route. I think we need to escape those interception path for when we use `toDestination`  and then replace those escaped path with  `(.)|(..)|(...)` so that we return the expected path

khuezy: Hmm but isn't that (.) regex, and not directory paths?

khuezy: I added a util function in utils.ts:
```
const replacePathRegex = /(\(?\w+\)?\/)?\(\.{2}\)\/?/;
export function resolvePaths(path: string) {
  const rootIndex = path.indexOf("(...)");
  // If path contains (...), replace everything before it, including itself
  if (~rootIndex) {
    path = path.substring(rootIndex + 5);
  }
  // Remove (.), since that resolves to nothing
  path = path.replace(/\(\.\)/g, "");

  // Resolves each (..) path by consuming the path before it
  // eg: /page/(..)[dynamic] => /[dynamic]
  while (path.match(/\(\.{2}\)/)) {
    path = path.replace(replacePathRegex, "");
  }
  return path;
}
```

The in routing.ts, I resolved the source & destination:
```
    const toDestination = compile(resolvePaths(rewrite?.destination) ?? "");
    const fromSource = match(resolvePaths(rewrite?.source) ?? "");
    ```

Comment (khuezy) [reply packages/open-next/src/adapters/routing.ts]: Nvm, the modal's param.title is messed up for some reason, it has `%3Atitle` from (params.title)

File: packages/open-next/src/adapters/routing.ts [hunk 5]
```diff
@@ -0,0 +1,109 @@
...
+    const toDestination = compile(rewrite?.destination ?? '');
+    const fromSource = match(rewrite?.source ?? '');
+    const _match = fromSource(rawPath);
+    if (_match) {
+      const { params } = _match;
+      const isUsingParams = Object.keys(params).length > 0;
+      if (isUsingParams) {
+        rewrittenUrl = toDestination(params);
+      }
+      rewrittenUrl = rewrite.destination;
```
khuezy, line 84: This should be inside an `else` statement, otherwise the `toDestination` gets overriden.

conico974: Ho come on, how could i forget that..
Ok at least this way it makes more sense to me, i couldn't understand how your fix could resolve the issue, i'm pretty sure we need the `(.)` in the final url.
I've just tried something and it seems to work

I created these 2 functions 
```typescript
export function escapeRegex(str: string) {
  let path = str.replace(/\(\.\)/g, '%1');

  path = path.replace(/\(\.{2}\)/g, '%2');

  path = path.replace(/\(\.{3}\)/g, '%3');

  return path;
}

export function unescapeRegex(str: string) {
  let path = str.replace(/%1/g, '(.)');

  path = path.replace(/%2/g, '(..)');

  path = path.replace(/%3/g, '(...)');

  return path;
}
```
And for the rewrite part 
```typescript
const toDestination = compile(
      escapeRegex(rewrite?.destination ?? '') ?? ''
    );
    const fromSource = match(escapeRegex(rewrite?.source) ?? '');
    const _match = fromSource(rawPath);
    if (_match) {
      const { params } = _match;
      const isUsingParams = Object.keys(params).length > 0;
      if (isUsingParams) {
        rewrittenUrl = unescapeRegex(toDestination(params));
        console.log('rewrittenUrl1', rewrittenUrl);
      } else {
        rewrittenUrl = rewrite.destination;
      }
      console.log('rewrittenUrl2', rewrittenUrl);
```

Of course those escape character are pretty bad, but for testing it's ok, i need to figure out some escape that will not mess things up

khuezy: I have an extremely ugly hack to get the correct params from the modal:

In routing.ts:
```
  let params: { [key: string]: string } = {};

  if (rewrite) {
    const fromSource = match(resolvePaths(rewrite?.source) ?? "");
    const _match = fromSource(rawPath);
    if (_match) {
      params = _match.params as any;
      rewrittenUrl = rewrite.destination;
    }
  }

  return {
    rawPath: rewrittenUrl,
    url: `${rewrittenUrl}${urlQueryString ? `?${urlQueryString}` : ""}`,
    params,
    __rewrite: rewrite,
  };
  ```
In server-adapter:
```
  const { rawPath, url, params } = handleRewrites(
    internalEvent,
    routesManifest.rewrites
  );
....
  const body = ServerResponse.body(res)
    .toString(encoding)
    .replace(/%3A([^"]+)/g, (_, match: string) => {
      return params[match] || `%3A${match}`;
    });
```

But there's gotta be a correct way to do this...

Comment (khuezy) [reply packages/open-next/src/adapters/routing.ts:84]: Cool, let me give that a go.

Comment (khuezy) [reply packages/open-next/src/adapters/routing.ts:84]: Nice! Modal interception + client parameters works now :)

Comment (khuezy) [reply packages/open-next/src/adapters/routing.ts:84]: Would `fromSource` response ever need to be unescapeRegex'd?

Comment (conico974) [reply packages/open-next/src/adapters/routing.ts:84]: Yes otherwise `next-server` will not find the correct interception route path.
I'll change the escape character to something like`_µ1_`, it should be safe enough so that user will not have these in the url

**Comment by khuezy** [comment]
I tried rewrite and redirect. Redirect works but `rewrite` returns 404... I'll debug some more. My rewrite config is:
```
async rewrites() {
  return [{souce: 'about', destination: '/'}]
}
```
This works locally but 404s on deployed app.

**Comment by conico974** [comment]
> I tried rewrite and redirect. Redirect works but `rewrite` returns 404... I'll debug some more. My rewrite config is:
> 
> ```
> async rewrites() {
>   return [{source: '/about', destination: '/'}]
> }
> ```
> 
> This works locally but 404s on deployed app.

It might be that it's generated as `afterFiles` or as `fallback` (see https://github.com/conico974/open-next/blob/aec5166911b7887735d8546dbc108bd86bdf7a6f/packages/open-next/src/adapters/util.ts#L48). I'm not sure how we should handle that, especially when.

I tested mine this way and it works for me, even with dynamic path
```typescript
rewrites: () => ({
    beforeFiles: [
      {
        source: '/shouldBeRewritten',
        destination: '/',
      },
      {
        source: '/shouldBeRewritten/:path',
        destination: '/:path',
      },
    ],
  }),
```

khuezy:
Ah yes, it's generated as `afterFiles`

packages/open-next/src/adapters/util.ts
```diff
@@ -1,66 +1,103 @@
...
-    // Remove group route params and /page suffix
-    const cleanedKey = key.replace(/\/\(\w+\)|\/page$/g, "");
+    // Remove parallel route
+    let cleanedKey = key.replace(/\/@[^\/]+/g, '');
+
+    // Remove group routes
+    cleanedKey = cleanedKey.replace(/\/\((?!\.)[^\)]*\)/g, '');
+
+    // Remove /page suffix
+    cleanedKey = cleanedKey.replace(/\/page$/g, '');
```
khuezy, line 79: @conico974 I'm using `app/api/something/route.tsx`, but the key still contains `app/api/something/route`, should we also remove the `route`? The nextjs codebase deletes the last `/<thing>` from the key.

conico974: Technically we should, but I don't think it matters that much. This is used only to modify react resolution so that your page use the same version of react as next. I don't think they do anything related to react on route

khuezy:
This breaks the image optimization lambda b/c of missing "next" package... will look later.

packages/open-next/src/adapters/util.ts
conico974: There is actually a gotcha, https://nextjs.org/docs/app/building-your-application/routing/middleware#matching-paths. We should follow this execution order.
I think in order to do that we'll have to merge this with your middleware fix, i'll try to come up with something later today

khuezy:
https://github.com/sst/open-next/pull/208

PR #156, opened 2023-07-07
fix: use basePath when generating public-files.json 

### LINKED ISSUE #136: basePath / assetPrefix is not respected in generated public-files.json
basically, just what the title says. 

if you have a basePath or assetPrefix set in your next.config.js the generated public-files.json do not include said basePath or prefix.  

workaround is to manually modify the public-files.json to include said prefix. 

i can/am happy to create a PR to resolve this.

breningham:
- fixes #136

### DIFF
--- a/.changeset/shy-dogs-watch.md
+++ b/.changeset/shy-dogs-watch.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+Read the routes-manifest.json to determine a prefix for public files listed in public-files.json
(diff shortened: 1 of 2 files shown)


conico974:
I'm not sure that this PR still makes sense. `public-files.json` should not be used anymore, it should be handled by the cloudfront distribution directly. See https://github.com/sst/open-next/blob/348349c41ea2106250cbd67514318d0f257d94b1/packages/open-next/src/adapters/server-adapter.ts#L96-L99

breningham:
Closing this PR, as @conico974 say's it no longer makes sense... might want to update the links to the docs though!

PR #159, opened 2023-07-13
fix: missing react/jsx-runtime for require-hooks

conico974:
Should fix `Cannot find module /var/task/node_modules/react/jsx-runtime.js`

### DIFF
--- a/.changeset/moody-ways-knock.md
+++ b/.changeset/moody-ways-knock.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+Server: fix missing react/jsx-runtime for require-hooks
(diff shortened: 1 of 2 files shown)


revmischa:
Latest nextjs is not useable and throws this error, so any fix would be greatly appreciated

suil:
anyone can get this resolved and merged. our prod is being broken now. please fix it. thx

fwang:
Released in the latest version v2.0.5

gnllucena:
how do I upgrade SST to use the latest OpenNext, @fwang?

running `npx sst deploy --stage prod` is enough to use the latest OpenNext?

fwang:
The latest SST v2.22.1 has the latest OpenNext

PR #160, opened 2023-07-14
Additional framework for CDK with support for TS,Java,Go,Py,.NET

matusfaro:
Adding a CDK framework forked from jetbridge/cdk-nextjs to support all other JSII languages. I needed to get it packaged for Java and thought others may benefit from it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,8 @@ For more information, check out the SST docs: https://docs.sst.dev/start/nextjs
 
 The OpenNext community has contributed deployment options for a few other frameworks.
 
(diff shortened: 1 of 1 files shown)


jayair:
Very cool thanks for sharing! Just a bit of feedback for your readme. We call it OpenNext instead of OpenNEXT.

matusfaro:
> Very cool thanks for sharing! Just a bit of feedback for your readme. We call it OpenNext instead of OpenNEXT.

Ah thanks for letting me know, I'll change it. I was going by how the logo spells it.

PR #169, opened 2023-08-04
fix: 13.4.13 handling of middleware

khuezy:
fixes: #176
Context: Starting with nextjs 13.4.13-canary.0, the middleware handling was removed from next-server, so no middlewares were being executed. Obviously a breaking change if your app depended on middleware handling redirects/rewrites/etc...

TODO: 

- [x] figure out a plugin system to only apply this to nextjs 13.4.13+

We can use an esbuild plugin to inject the code conditionally.

@conico974 and I have been brainstorming a solution: https://discord.com/channels/983865673656705025/1137034343580172308

### DIFF
--- a/packages/open-next/src/adapters/plugins/13_4_13.ts
+++ b/packages/open-next/src/adapters/plugins/13_4_13.ts
@@ -1,6 +1,7 @@
+import type { PluginHandler, Options } from "../next-types.js";
+
 //#override imports
 import path from "node:path";
-import { PluginHandler, type Options } from "../next-types.js";
 import { IncomingMessage } from "../request.js";
 import { ServerResponse } from "../response.js";
 import { config, NEXT_DIR } from "../util.js";
--- a/packages/open-next/src/adapters/plugins/13_4_13.ts
+++ b/packages/open-next/src/adapters/plugins/13_4_13.ts
@@ -6,7 +6,7 @@ import {
   loadConfig,
   getMiddlewareMatch,
   loadMiddlewareManifest,
-  createRequestHandler,
+  requestHandler,
   NEXT_DIR,
 } from "../util.js";
 
@@ -20,7 +20,6 @@ const {
 } = require("next/dist/server/web/spec-extension/adapters/next-request");
 const config = loadConfig(NEXT_DIR);
 const middleMatch = getMiddlewareMatch(middlewareManifest);
-const requestHandler = createRequestHandler();
 
 export const handler: PluginHandler = async (
   req: IncomingMessage,
--- a/packages/open-next/src/adapters/plugins/13_4_13.ts
+++ b/packages/open-next/src/adapters/plugins/13_4_13.ts
@@ -2,13 +2,8 @@ import path from "node:path";
 import { PluginHandler, type Options } from "../next-types.js";
 import { IncomingMessage } from "../request.js";
 import { ServerResponse } from "../response.js";
-import {
-  loadConfig,
-  getMiddlewareMatch,
-  loadMiddlewareManifest,
-  requestHandler,
-  NEXT_DIR,
-} from "../util.js";
+import { config, NEXT_DIR } from "../util.js";
+import { requestHandler, getMiddlewareMatch, loadMiddlewareManifest } from "./util.js";
 
 const middlewareManifest = loadMiddlewareManifest(NEXT_DIR);
 
@@ -18,7 +13,6 @@ const { getCloneableBody } = require("next/dist/server/body-streams");
 const {
   signalFromNodeResponse,
 } = require("next/dist/server/web/spec-extension/adapters/next-request");
-const config = loadConfig(NEXT_DIR);
 const middleMatch = getMiddlewareMatch(middlewareManifest);
 
 export const handler: PluginHandler = async (
--- a/packages/open-next/src/adapters/plugins/13_4_13.ts
+++ b/packages/open-next/src/adapters/plugins/13_4_13.ts
@@ -5,7 +5,12 @@ import path from "node:path";
 import { IncomingMessage } from "../request.js";
 import { ServerResponse } from "../response.js";
 import { config, NEXT_DIR } from "../util.js";
-import { requestHandler, getMiddlewareMatch, loadMiddlewareManifest } from "./util.js";
+import {
+  requestHandler,
+  getMiddlewareMatch,
+  loadMiddlewareManifest,
+  setNextjsPrebundledReact,
+} from "./util.js";
 
 const middlewareManifest = loadMiddlewareManifest(NEXT_DIR);
 
@@ -30,7 +35,7 @@ export const handler: PluginHandler = async (
   // Middleware
   const ended = await handleMiddleware(req, res, rawPath);
   if (ended) return;
-
+  setNextjsPrebundledReact(rawPath);
   // Next Server
   return requestHandler(req, res);
 };
@@ -52,7 +57,9 @@ async function handleMiddleware(
   // but that was discarded for simplicity. The MiddlewareInfo type still has the original
   // structure, but as of now, the only useful property on it is the "/" key (ie root).
   const middlewareInfo = middlewareManifest.middleware["/"];
-  middlewareInfo.paths = middlewareInfo.files.map((file) => path.join(NEXT_DIR, file));
+  middlewareInfo.paths = middlewareInfo.files.map((file) =>
+    path.join(NEXT_DIR, file)
+  );
 
   const result = await run({
     distDir: NEXT_DIR,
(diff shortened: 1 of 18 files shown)


packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -210,13 +222,89 @@ function createRequestHandler() {
...
+  // the redirected url and end the response.
+  if (res.statusCode >= 300 && res.statusCode < 400) {
+    const location = result.response.headers
+      .get("location")
+      ?.replace("http://localhost:3000", `https://${req.headers.host}`);
+    res.setHeader("Location", location);
+    return res.end();
+  }
+  // Apply the headers from the middleware to the `req` to be processed by the server
+  for (let [key, value] of result.response.headers) {
```
khuezy, line 275: We may need to omit some headers.
TODO: double check

conico974: There is a `x-middleware-override-headers` that the middleware set that we may need to use. I have not tested this but the middleware return the overriden headers as `x-middleware-request-HEADER_NAME`, we might need to apply them ourselves.

I did this, not sure if it's necessary: 
```typescript
const toOverride: string =
  middlewareHeaders['x-middleware-override-headers'];
const overrideHeaders = toOverride.split(',').reduce(
  (acc, cur) => ({
    ...acc,
    [cur]: middlewareHeaders[`x-middleware-request-${cur}`],
  }),
  {}
);
```

conico974:
I might have an idea for the plugin system. We could write an esbuild plugin like this one https://github.com/zziger/esbuild-ifdef. Instead of just conditionally outputting the code, we could provide an override file which contains the part that we want to override.

We could use it like that
```typescript 
esbuild.build({
    entryPoints: ['./index.js'],
    bundle: true,
    target: 'es6',
    outfile: './out.js',
    plugins: [
        ourPlugin({
            overrideFilePath: "path/to/the/file.ts", 
            // ... plugin config
        })
    ]
});
```

And on our file and our override file we just override part surrounded like that 
```typescript 
/// _Override UNIQUE_ID
const partToOverride = 0;
/// _EndOverride
```
We could even make it possible to apply this plugin several times for different override

packages/open-next/src/adapters/server-adapter.ts
khuezy: Is there a good way to trigger that? I'm not seeing the `x-middleware-override-headers` from the `result.response.headers`.
In my middleware, I'm setting some test headers via `NextResponse.next({headers: {}})`

khuezy:
@conico974, as of the current situation, I think we can move the `async function processRequest` into its own file, then we can use the "`ourPlugin`" above to replace the implementation for the specific next version.

eg:
```typescript
        ourPlugin({
            overrideFilePath: getFilePathForNextVersion(nextVersion), 
            // ... plugin config
        })
```

If there's major changes in the future that requires a broader scope, we can update it then.

khuezy:
I tried moving `import NextServer from "next/dist/server/next-server.js";` into `util.js` but esbuild complains b/c it couldn't resolve that path... why doesn't work if that's moved outside the `server-adapter`?

conico974:
> I tried moving `import NextServer from "next/dist/server/next-server.js";` into `util.js` but esbuild complains b/c it couldn't resolve that path... why doesn't work if that's moved outside the `server-adapter`?

It might be the other functions that are built (i.e. revalidation function) You probably need to add next in external in esbuild for all the other function since utils is used there.

packages/open-next/src/adapters/server-adapter.ts
conico974: My bad it seems it's only for when we modify the request headers. https://nextjs.org/docs/app/building-your-application/routing/middleware#setting-headers. We still need to do it though

khuezy:
Great thanks, that was exactly it. I'll add `external: ['sharp', 'next']` to the base esbuild since those should always be external no matter what.

packages/open-next/src/adapters/plugins/13_4_13.ts
```diff
@@ -0,0 +1,127 @@
...
+    Nextjs will set `x-middleware-override-headers` as a comma separated list of keys.
+    All the keys will be prefixed with `x-middleware-request-<key>`
+
+    We can delete `x-middleware-override-headers` and check if the key starts with
+    x-middleware-request- to set the req headers
+  */
+  const responseHeaders = result.response.headers as Headers;
+  responseHeaders.delete("x-middleware-override-headers");
+  const xMiddlewareKey = "x-middleware-request-";
+  responseHeaders.forEach((value, key) => {
```
khuezy, line 106: cc @conico974 , I opted for this implementation instead of the reducer b/c it seemed easier to loop once and apply the `x-middleware-request-` prefixed headers + headers that were added outside the `request` object.

packages/open-next/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "declaration": true,
     "module": "esnext",
-    "moduleResolution": "nodenext",
+    "lib": ["DOM", "ESNext"],
```
khuezy, line 6: Nextjs uses the DOM lib for the Headers object, but I'm not sure what's the actual correct lib to load since it was still missing partial types for the iterators. But using "DOM.iterator" complains about other things... zzz

packages/open-next/src/plugin.ts
conico974, line 1: I've started looking more closely into the plugin system, and i think we need something that allow for more customization than this. Something that we could use like that:
On the `plugin/default` files
```typescript 
//this could be overriden in one of the replacement files
//#override id1
export const toReplace = () => {
  console.log('toReplace');
};
//#endOverride

//#override id2
export const toReplace2 = () => {
  console.log('toReplace');
};
//#endOverride

export const notReplaced = 0;
```
And for the replacement files something like that
```typescript
//import will be placed at top of the files
// It could be import or anything else actually

//#import
const REPLACEMENT_VALUE = 78;
//#endImport

//#override id1
export const replacement = REPLACEMENT_VALUE;
//#endOverride
```
Here is what i've come up with so far:
```typescript
import { readFile } from 'node:fs/promises';
import { Plugin } from 'esbuild';

export interface IPluginSettings {
  importPaths: string[];
}

const overridePattern = /\/\/#override (\w+)\n([\s\S]*?)\n\/\/#endOverride/gm;
const importPattern = /\/\/#import([\s\S]*?)\n\/\/#endImport/gm;

export default function openNextPlugin({
  importPaths,
}: IPluginSettings): Plugin {
  return {
    name: 'opennext',
    setup(build) {
      build.onLoad({ filter: /plugins\/default.js/g }, async (args) => {
        let contents = await readFile(args.path, 'utf-8');

        importPaths.map(async (fp) => {
          const replacementFile = await readFile(fp, 'utf-8');
          console.log(replacementFile);
          const matches = replacementFile.matchAll(overridePattern);

          const importMatch = replacementFile.match(importPattern);
          const addedImport = importMatch ? importMatch[0] : '';

          contents = `${addedImport}\n${contents}`;

          for (const match of matches) {
            const replacement = match[2];
            const id = match[1];
            const pattern = new RegExp(
              `\/\/#override (${id})\n([\\s\\S]*?)\n\/\/#endOverride`
            );
            contents = contents.replace(pattern, replacement);
          }
        });
        return {
          contents,
        };
      });
    },
  };
}
```

khuezy: Will apply your suggestion on Monday.

khuezy: I had to replace the `.map` with a regular `for loop` b/c `async` in maps doesn't properly await. Everything else works great but I need to deal w/ duplicate import statements from the overrides.

packages/open-next/src/adapters/plugins/README.md
```diff
@@ -0,0 +1,15 @@
+### Known issues
```
khuezy, line 1: @conico974 I spent a couple hours trying to figure out why the replacement stuff was broken... Looks like an esbuild bug removing the top comment when it omits the types, it also drops the head comment :|

conico974: I see, i didn't test with `import type`. But do we really need to put those inside the override ? esbuild will remove them anyway, can't we just put them at the top of the file ?

khuezy: Good point - yes we can. I'll update the known issue to note that `types` should be excluded from the plugin comments.

packages/open-next/src/plugin.ts
sanny-io: > I had to replace the .map with a regular for loop b/c async in maps doesn't properly await.

@khuezy I think this would work for you.

```javascript
await Promise.all(importPaths.map(async (fp) => {
   // ...
}))
```

khuezy: Thanks @sanny-io I forgot about that.

PR #170, opened 2023-08-05
feat(streaming): allow lambda image function to return streamed response

sladg:
This PR reacts to AWS's streaming. See: https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/. It's a minimalistic implementation purely for Image handler which we can continue building upon with other handlers.

In order to make this work, we need to do following:
- `invokeMode: InvokeMode.RESPONSE_STREAM` to be added here: https://github.com/serverless-stack/sst/blob/0529f6291e925aa3809abc746429964dcfce433b/packages/sst/src/constructs/NextjsSite.ts#L415C8-L415C8

Related to: https://github.com/serverless-stack/open-next/issues/79

Previous implementation is kept, it might be reasonable to remove it if all is fine with this PR.

### DIFF
--- a/packages/open-next/src/adapters/image-optimization-adapter.ts
+++ b/packages/open-next/src/adapters/image-optimization-adapter.ts
@@ -21,6 +21,8 @@ import {
 } from "next/dist/server/image-optimizer";
 import { loadConfig, setNodeEnv } from "./util.js";
 import { debug, error, awsLogger } from "./logger.js";
+import { RequestHandler } from "../types.js";
+import { streamError, streamSuccess } from "./streaming.js";
 
 // Expected environment variables
 const { BUCKET_NAME, BUCKET_KEY_PREFIX } = process.env;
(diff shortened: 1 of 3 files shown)


khuezy:
Nice, I'll test this on Monday.

khuezy:
Hi @sladg did you have to do anything special to get the images to work? I applied this commit to my local branch but I'm getting a transparent image 

image

sladg:
@khuezy I tested directly via API and Postman / browser to call `_next/image` route as well as given Function URL of image optimiser. Let me re-check with `<Image/>` to be sure.

I needed to manually change Lambda's Invocation mode in AWS console.

khuezy:
Oh my bad I didn't read the instructions. I didn't add the RESPONSE_STREAM in NextjsSite. Let me try that.

khuezy:
Nice work! 
image

I'll create a help thread in the discord for us to sync on bringing streaming to the server function. That'll enable to app to do SSE + any thing streaming

sladg:
Closing as this feature was released as part of other PR

PR #173, opened 2023-08-09
feat: use trustHostHeader to set req.url to host

khuezy:
Currently, `request.url` (eg inside a .route file) will be `https://localhost:3000/api/hi`.

This PR sets `trustHostHeader: true` which sets `request.url` to the actual url of the app, eg: `https://example.com/api/hi`


khuezy:
BLOCKER: This breaks next 13.4.12 and below, need to investigate.

khuezy:
This breaks 13.4.12 and below, this will be handled in https://github.com/sst/open-next/pull/169, which has a plugin system to modify code based off next version.

Issue #174, opened 2023-08-11
Codestyle guidelines


sladg:
👋 Looking at my experience as well as some other PRs (https://github.com/sst/open-next/pull/155/files) it would be wise to add Eslint config to the project to ensure we follow same formatting while developing.

khuezy:
Ditto, I had to wrestle w/ my VSCode settings b/c there wasn't a lint config in this project.

sladg:
@khuezy I'm gonna open PR for this, should be fairly simple, we just need to pick right setup for existing codebase 🚀

PR #175, opened 2023-08-11
Feat: add support for streaming

conico974:
This PR aims to add support for streaming in open-next.

In order to test this you need to set an env variable `USE_STREAMING_RESPONSE: true`, you'll also need to set lambda invoke mode to `RESPONSE_STREAM`. https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html#config-rs-invoke-furls

## Caveats
There is a couple of quirks with how lambda streaming seems to be working: 
- For the lambda to start the streaming process, it seems that the event loop needs to be empty. 
- It also seems that it needs a minimum amount of data, or a certain amount of time before starting the stream.

### TODO
- [x] Right now the writing process is inside a `setInterval`, find a better way to do this
- [ ] Add this into the plugin system


conico974:
For a better understanding of some of the quirks of lambda streaming, try to deploy this as a lambda and play around with it.
For example, if you comment the `for` you'll notice that the streaming doesn't start before 1.5-2 seconds

```javascript
exports.handler = awslambda.streamifyResponse(
    async (event, responseStream, context) => {
        const httpResponseMetadata = {
            statusCode: 200,
            headers: {
                "Content-Type": "text/html",
                "X-Custom-Header": "Example-Custom-Header"
            }
        };

        responseStream = awslambda.HttpResponseStream.from(responseStream, httpResponseMetadata);
        
        for(let i=0; i<10; i++){
            responseStream.write("\r\n");
            await new Promise(r => setTimeout(r, 1));
        }

        responseStream.write("<p>First write!</p>");
        await new Promise(r => setTimeout(r, 1500));
        responseStream.write("<p>Second write4!</p>");

        responseStream.write("<h1>Streaming h1</h1>");
        await new Promise(r => setTimeout(r, 50));
        responseStream.write("<h2>Streaming h2</h2>");
        await new Promise(r => setTimeout(r, 500));
        responseStream.write("<h3>Streaming h3</h3>");
        await new Promise(r => setTimeout(r, 500));
  
        const loremIpsum1 = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque vitae mi tincidunt tellus ultricies dignissim id et diam. Morbi pharetra eu nisi et finibus. Vivamus diam nulla, vulputate et nisl cursus, pellentesque vehicula libero. Cras imperdiet lorem ante, non posuere dolor sollicitudin a. Vestibulum ipsum lacus, blandit nec augue id, lobortis dictum urna. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae; Morbi auctor orci eget tellus aliquam, non maximus massa porta. In diam ante, pulvinar aliquam nisl non, elementum hendrerit sapien. Vestibulum massa nunc, mattis non congue vitae, placerat in quam. Nam vulputate lectus metus, et dignissim erat varius a.";
        responseStream.write(`<p>${loremIpsum1}</p>`);
        await new Promise(r => setTimeout(r, 1000));
  
        responseStream.write("<p>DONE!</p>");
        responseStream.end();
    }
);
```

sladg:
Heya! I opened issue on this (#174), any chance you could change local setting for prettier, hard to distinguish between style nad actual changes

packages/open-next/src/adapters/http/responseStreaming.ts
```diff
@@ -0,0 +1,181 @@
...
+  writeHead(
+    statusCode: number,
+    headers?: http.OutgoingHttpHeaders | http.OutgoingHttpHeader[] | undefined
+  ): this;
+  writeHead(
+    statusCode: unknown,
+    statusMessage?: unknown,
+    headers?: unknown
+  ): this {
+    if (!this._wroteHeader) {
```
sladg, line 43: can we go the other way around? aka. not to nest when not necessary?

I would suggest going `if(this._wroteHeader) { return this }`

packages/open-next/src/adapters/http/responseStreaming.ts
```diff
@@ -0,0 +1,181 @@
...
+        debug('writeHead', this[HEADERS]);
+      } catch (e) {
+        this.responseStream.end();
+        error(e);
+      }
+    }
+    return this;
+  }
+
+  // WORKAROUND: It seems that lambda streaming require the event loop to be free
```
sladg, line 67: when does this happen? I wasn't able to find any note neither within Lambda code nor documentation. Is this on our end or AWS's?

conico974: I've looked at this file https://gist.github.com/conico974/13afd708af20711b97df439b910ceb53#file-index-mjs-L871. That's the only explanation i found as to why this happens, but honestly i don't really understand what's going on here. It might be something on the other end of the internal request that lambda is doing

At first i tried writing directly to the stream, but the streaming never started, it was just waiting for the end to send all the data like a regular request.
What i did is almost the same as now:
- You just comment the listener part
- Everytime there is a `this[BODY].push(chunk)` i called `this.responseStream.write(chunk)`
- And on the `end` method you call `this.responseStream.end()`

There is some really funky stuff going on here https://github.com/sst/open-next/pull/175#issuecomment-1674926950

packages/open-next/src/adapters/http/response.ts
```diff
@@ -3,33 +3,19 @@
 // Licensed under the MIT License
 
 // @ts-nocheck
-import http from "node:http";
+import http from 'node:http';
+import { getString, headerEnd } from './utils.js';
 
-const headerEnd = "\r\n\r\n";
+const headerEnd = '\r\n\r\n';
```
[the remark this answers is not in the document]
conico974: I don't know much about SSE, but with response streaming new line seems to be `\r\n` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked_encoding.
On mozilla docs there is only one new line, but two from next

khuezy:
Have you tried using `process.nextTick` instead of `setImmediate?` If it works, we'd probably prefer `nextTick` b/c it cuts the line in front of the event loop.

conico974:
> Have you tried using `process.nextTick` instead of `setImmediate?` If it works, we'd probably prefer `nextTick` b/c it cuts the line in front of the event loop.

I've tried that already, unfortunately it doesn't work

conico974:
Replaced by #213

Issue #118, opened 2023-05-18
RFC: Add unit testing


khuezy:
At a minimum, we should have some unit testing of some utils function to ensure robustness of open-next.

RFC: Decide what's the best testing framework to use w/in context of the open-next framework.

khuezy:
Cool, sounds like a plan. I'll try to get something up by next week. Still need to figure how deploying would work (via SST would make sense but would need @fwang to set that up).

For the e2e integration, we should deploy: https://app-router.vercel.app/ and ensure all supported features are functional.

conico974:
> Cool, sounds like a plan. I'll try to get something up by next week. Still need to figure how deploying would work (via SST would make sense but would need @fwang to set that up).
> 
> For the e2e integration, we should deploy: https://app-router.vercel.app/ and ensure all supported features are functional.

We might need to modify it, for a good e2e we need to have page dir in it too. 
In a perfect world we'd have 3 different app, app dir only, page dir only and page + app dir. And testing against different version of next might be useful too.

khuezy:
In that case, we can use the `example` in here. Create 3 types of apps: `appDir` only, `pages` only, both. And deploy each as a subdomain: `https://app|pages|both.open-next.js.org/`

devmanbr:
I think it's excellent. it's not sustainable to monitor compatibility every time next.js updates without automated testing.

PR #178, opened 2023-08-16
Feat: Add unit and e2e testing

khuezy:
fixes: #118 

This PR adds unit/e2e testing and `@open-next` packages for modular development and testing, but currently, `open-next` does not import the `@open-next/utils` package b/c doing so would change the `dist` output and I didn't want to deal w/ that yet.

You can test this locally by running: `pnpm run dev` to run the next apps and then `pnpm run e2e:test`
You can test against a deployed app by setting the proper env variable:
```
/examples/sst

export APP_DIR_ONLY_URL=$(jq -r '.["e2e-example-AppDirOnly"].url' .sst/outputs.json)
export PAGES_ONLY_URL=$(jq -r '.["e2e-example-PagesOnly"].url' .sst/outputs.json)
export APP_DIR_AND_PAGES_URL=$(jq -r '.["e2e-example-AppDirAndPages"].url' .sst/outputs.json)
```
The current Nextjs Version is 13.4.12 (last stable version). But once the open PRs are merged to address the issues, we'll want to set the version to `latest`

todo:

- [x] Add AppDir + Pages project
- [x] Add more scenarios

@fwang 

- [x] Add SST credentials to project for deployment
- [x] Validate that the e2e.yml is properly configured


khuezy:
I'm having trouble setting up the aws credentials :/

fwang:
> I'm having trouble setting up the aws credentials :/

I think it's b/c the PR is created from an external repo. It seems to work when PR is created from the local branch — https://github.com/sst/open-next/pull/196

fwang:
This marks the beginning of OpenNext's stability 🫡.

A few thoughts:

1. What is `example_sst` going to be?
  We had `example` which the purpose was to deploy it to Vercel and AWS on every push. And at any time ppl can compare the two sites and determine something is a bug in OpenNext. And over time we can add more pages in there (ie. I recently added a NextAuth page)
  I like `example_sst` has both App and Pages sites. Should `example_sst` replace `example`?

2. Layout inside `packages`
  Just want to brainstorm on the structure. Does something like this make sense:
  ```
  packages/
    open-next/
      tests/           <-- move `tests_unit` here since the tests is scoped to the `open-next` package
    example-appdir/    <-- move `app-dir` here, make it an SST app with its `sst.config.ts`
    example-pages/     <-- move `pages-only` here, make it another SST app with its `sst.config.ts`
    core/              <-- it seems `core` is only used in `app-dir`, does it belong here or just put it in `app-dir`?
    utils/             <-- functions here are helpers for `open-next`, does it belong here?
    tests-e2e/
  ```

sladg:
Conflicts seems to be oddly resolved as changes shown does not reflect latest master (aka. showing changes which are already on master). My suggestion would be `git checkout main; git pull; git checkout my/branch; git rebase main`. This way we keep file/line history without overriding.

sladg:
@fwang having internal "examples" inside `packages/` dir might be confusing to differentiate between what's actual NPM package and what's not. I would suggest having `examples/` dir which would be part of workspaces. Then we can have plenty of examples and tests without polluting actual `packages/` dir.
What do you think?

mathisobadia:
>    I like `example_sst` has both App and Pages sites. Should `example_sst` replace `example`?

Yes please, anytime I tried to do something with the example folder to check if something is a bug or not I lost so much time because it was js and not ts  🥲

khuezy:
@sladg I like that setup, I'll make the changes. As for the merge/main issue, I merged the latest main but probably messed up somewhere.

@fwang I originally had the `tests-unit` as its own packages b/c it's one of turbo's design pattern. I do that in my personal project as it lets build/run the unit tests independently of the core app (modularity).

khuezy:
@fwang @sladg should we separate out each of the e2e tests in their own `yml` file?

fwang:
> @fwang @sladg should we separate out each of the e2e tests in their own `yml` file?

I'm not familiar w/ e2e setups. I will let @sladg chime in.

khuezy:
I've moved the internal stuff to the `examples/` folder. I have an `examples/sst` as the base SST project that contains the Nextjs Stacks for the e2e tests.

We could try to separate out each e2e tests into their own deployment and action, but for now, this should be a decent setup.

khuezy:
Yup, playwright runs ~5 in parallel (by default). Let's do the separate tests in a different PR after all this stuff settles. I think it's fine to do 1 deployment for now.
I'll create the next PR after we merge 13.4.13+ fixes (and streaming)

Issue #184, opened 2023-08-24
Project codebase improvements


sladg:
I have noticed couple of things which are not sub-optimal and messes up developer experience when working on project.

1. Use of Plain JS in example site. Considering SST is written mostly in TS as well as open-next, I would expect even tests and examples being done in same way.
2. `packages/open-next/assets` contains copy-pasted `node_modules` for Image optimisation Lambda. These shouldn't be part of repository, but should be dealt with in more reasonable way, such as: zip file, lambda layer, minified dist file, or similar.
3. No automated tests or deployments are in place. Considering we already have example site with lot of cases identified by @fwang. It would make sense to setup automated testing so any PR gets bundled, deployed and tested. This could also include daily tests for specific Next versions so we can ensure compatibility with wide-range of projects. Considering the serverless nature of open-next, I should be rather straight forward to achieve.
4. We are relying on `tsc` to build files for this package. Uses `tsup` is generally best case practise as it allows for extra configuration such as multiple inputs, bundling and minifying the output.

jayair:
Thanks! Just to double check, you are in our Discord right?

sladg:
Additional thing we came across in regards to streaming.

Current copy of `serverless-http`'s `response` is rather finicky. We might want to clean it up properly and provide easy commands to call from our wrapper.

sladg:
Might be worth investigating https://github.com/awslabs/aws-lambda-web-adapter and trying to use it as translation layer between event and HTTP. It's as first-party as we can get.

conico974:
I consider this one mostly done. 
Feel free to reopen it if needed

PR #186, opened 2023-08-25
Update image lambda assets

### LINKED ISSUE #185: Update semver in bundled image lambda assets
The current semver version bundled in the sharp assets is affected by [CVE-2022-25883](https://www.cve.org/CVERecord?id=CVE-2022-25883) and is currently flagged by the cloudwatch security scanner.

 What is the process to update these checked in assets?

edvinerikson:
Fixes #185 
 Added a script `update-sharp-assets.sh` to update the assets automatically


edvinerikson:
Note: something isn't right with the dependencies. Since I used linux, the native arm binaries were replaced by linux ones. NOt sure how I can force npm to install arm binaries while still on linux..

conico974:
> Note: something isn't right with the dependencies. Since I used linux, the native arm binaries were replaced by linux ones. NOt sure how I can force npm to install arm binaries while still on linux..

You could try a command like that `npm install --target_arch=arm64 --target_platform=linux` or maybe with `--arch=arm64` instead. There is not a lot of docs around that.

Anyway, i think it would be a great idea to write a script to automate this process for us in the future. It would also be great to output a single zip file instead of all those files.

edvinerikson:
> You could try a command like that npm install --target_arch=arm64 --target_platform=linux or maybe with --arch=arm64 instead. There is not a lot of docs around that.

Thanks, I'll try that!

> Anyway, i think it would be a great idea to write a script to automate this process for us in the future. It would also be great to output a single zip file instead of all those files.

I agree, I will try to create a script for it. However I will wait with the zipfile since that is a breaking change

sladg:
forcing arm/amd from other platform is virtually impossible with node currently 😢 I guess that's the reason these were committed and part of repo in a first place. We might want to run these in pipelines in Github or make script for running this in docker in given platform

sladg:
Related to #184

edvinerikson:
> forcing arm/amd from other platform is virtually impossible with node currently 😢 I guess that's the reason these were committed and part of repo in a first place. We might want to run these in pipelines in Github or make script for running this in docker in given platform


`--arch=arm64` seems to have made a difference at least. Now there's linuxarm64 binaries instead of linuxamd.

conico974:
> I agree, I will try to create a script for it. However I will wait with the zipfile since that is a breaking change
The bundling is done here https://github.com/sst/open-next/blob/274d446ed7e940cfbe7ce05a21108f4c854ee37a/packages/open-next/src/build.ts#L240. You could unzip at this time

PR #188, opened 2023-08-27
Exclude sharp from library

sladg:
We can install Sharp just-in-time when bundling NextJS. Having over thousand files as part of repository is rather not ideal.
Related to #184.

For more resources:
- https://github.com/lovell/sharp/blob/main/docs/install.md#aws-lambda
- https://github.com/lovell/sharp/blob/main/docs/install.md#cross-platform
- https://github.com/sladg/nextjs-lambda/blob/release/v1.1.2/scripts/build-sharp-layer.sh

Related to: #186


sladg:
Deployed version of example page bundled with this PR: https://duzc9gdmxipd2.cloudfront.net

khuezy:
On your deployed site, the image looks like it failed to get optimized. Can you check the image optimization lambda for errors?
image

Notice the dimension of the image and the type, it should be `webp` if optimized.

sladg:
@khuezy had wrong architecture, fixed. You can check it out here: https://d2stw39djmiwg.cloudfront.net

khuezy (approved):
Nice!

khuezy:
Thanks, the repo is a lot lighter now :)

PR #193, opened 2023-08-28
ci(prettier): remove invalid prettier reference from pipeline

sladg:
Fixes wrong reference inside Github actions after changes made in #183

### DIFF
--- a/.github/actions/pnpm-setup/action.yml
+++ b/.github/actions/pnpm-setup/action.yml
@@ -1,0 +1,31 @@
+name: Install & setup pnpm
+description: Install's node, pnpm, restores cache
+
+runs:
+  using: 'composite'
+  steps:
+    # Install nodejs. https://github.com/actions/setup-node
+    - name: Setup Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16.x
+
+    # Install pnpm. https://github.com/pnpm/action-setup
+    - uses: pnpm/action-setup@v2
+      with:
+        version: 8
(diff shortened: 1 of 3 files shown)


khuezy (approved):

.github/actions/pnpm-setup/action.yml
```diff
@@ -0,0 +1,31 @@
...
+description: Install's node, pnpm, restores cache
+
+runs:
+  using: 'composite'
+  steps:
+    # Install nodejs. https://github.com/actions/setup-node
+    - name: Setup Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16.x
```
khuezy, line 11: Should we upgrade to 18 here?

khuezy (approved):

.github/actions/pnpm-setup/action.yml
sladg: I'm not sure if there might be somone on node 16. It's could be quite dramatic in theory. SST seems to use 16 in their workflows. "You'll need at least [Node.js 16.6](https://nodejs.org/) and [npm 7](https://www.npmjs.com/). You also need to have an AWS account and [AWS credentials configured locally](https://docs.sst.dev/advanced/iam-credentials#loading-from-a-file)." from SST docs

khuezy: Ok, let's just leave for now. Hopefully pnpm 8 doesn't complain

PR #213, opened 2023-09-02
feat: support for streaming

### LINKED ISSUE #79: Support Streaming
AWS Lambda just introduced streaming capabilities via Lambda URLs. Opening this issue to gather some thoughts/interests from the community.

Ref:
https://beta.nextjs.org/docs/data-fetching/streaming-and-suspense
https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/

conico974:
This PR aims to add support for streaming in open-next.

In order to test this you need to run open-next with a build args `npx open-next build --streaming`
You'll also need to set lambda invoke mode to RESPONSE_STREAM. https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html#config-rs-invoke-furls

Replace #175
Fix #79 

### Todo
- [ ] Still need to refactor this a little bit so that it is cleaner
- [ ] Write tests


packages/open-next/src/adapters/http/responseStreaming.ts
```diff
@@ -0,0 +1,148 @@
...
+          cb();
+        }
+        return true;
+      },
+    });
+
+    this.responseStream.on("error", (err) => {
+      this.emit("error", err);
+      this.responseStream.end();
+      error("error", err);
```
sladg, line 145: This might be where we move `error` logging few lines up, in case `this.responseStream.end();` fails, we not gonna get original message

PR #214, opened 2023-09-04
Introduce a custom build-output-path and app-path argument for more flexible monorepo support.

lorenzodejong:
This change introduces a custom `build-output-path` and `app-path` argument which makes it more flexible to support monorepo flavors with decentralised build options.

An example of such a monorepo is an [Nx integrated repository](https://nx.dev/getting-started/tutorials/integrated-repo-tutorial), where apps are placed under `apps/[appName]` and build artifacts are output to `dist/apps/[appName]`. This will allow you to build a more generic script which is being executed from the root of the monorepo.

Besides that a small change is included to also support single-version monorepos, where the `package.json` is exclusively maintained from the root of the monorepo.

These changes are made in a non-breakable manner, paths are being defaulted to the current `process.cwd()` just like before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,6 +616,26 @@ await build({
 });
 ```
 
+#### Custom app and build output paths
+
+OpenNext runs the `build` script from your current command folder by default. When running OpenNext from a monorepo with decentralised application and build output paths, you can specify a custom `appPath` and/or `buildOutputPath`. This will allow you to execute your command from the root of the monorepo.
+
+```bash
+# CLI
+open-next build --build-command "pnpm custom:build" --app-path "./apps/example-app" --build-output-path "./dist/apps/example-app"
+```
+
+```ts
+// JS
+import { build } from "open-next/build.js";
+
+await build({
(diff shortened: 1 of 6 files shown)


khuezy:
Hey @lorenzodejong thanks for putting this up! The nx folks will appreciate your efforts. I'll test this and cut a release after some of the 2.1.0 goes in in a few hours. I'll try to get 2.1.1 w/ this next.

khuezy:
Thanks @lorenzodejong, this will go out in 2.1.1 soon.

lorenzodejong:
Great to hear @khuezy, thanks for the review!

khuezy:
@lorenzodejong can you try out 2.1.1 w/ `nx` and let us know if that behaves as desired? Some people on the discord are having issues. https://discord.com/channels/983865673656705025/1027265626085019769/1148981157438951444
If there's some missing gaps, we can fix in 2.1.2

lorenzodejong:
@khuezy i will take a look and also update my results in the Discord when i get home.

Issue #236, opened 2023-09-13
Windows Image Optimization Build fails.


khuezy:
Introduced in: https://github.com/sst/open-next/pull/219

- the --prefix breaks when the user's name has a space it in.

sladg:
Let me take a look into it!

PR #237, opened 2023-09-13
fix: use env for sharp flag on Windows

khuezy:
- env variable was not properly set on windows
- using the `env` property doesn't work
- wrap the `prefix` in quote to handle paths with white spaces
fixes: #236


khuezy:
cc @sladg 
I've tested this on Mac and had a discord user help verify on windows https://discord.com/channels/983865673656705025/1151634089032556554/1151634089032556554

PR #238, opened 2023-09-18
fix(sharp): escape path and simplify env setting

### LINKED ISSUE #236: Windows Image Optimization Build fails.

sladg:
Fixes #236 
Replaces #237


khuezy (approved):
Thanks!

I forgot to spread `process.env` when I tried doing that.

khuezy:
This is working for me, if caching is obfuscating any issues - we'll address it in the future.
We can release this as 2.1.5 or wait for the 2.2.0 train.

PR #257, opened 2023-10-04
Fix : incremental cache fixes

conico974:
The incremental cache in nextjs is actually recreated on every request, everything inside the constructor of incremental cache is executed several times during the lifetime of a lambda.

This PR should solve several issues with incremental cache :

- An issue arise when the lambda stays alive long enough, you'll get a `EMFILE: too many open files` error. To solve this we replaced `loadBuildId` with `process.env.NEXT_BUILD_ID` and made s3Client and dynamoClient global to the lambda. It should solve this issue :crossed_fingers: 
- The dynamodb provider was not generated when there was no fetch cache, but if you don't use fetch cache but want to use `revalidatePath`, it would not have worked.
- There is 2 new flags, `--dangerously-disable-dynamodb-cache` and `--dangerously-disable-incremental-cache`

When you disable the incremental cache every request to the incremental cache will return null and won't perform any other actions. If you disable the dynamodb cache, `revalidateTags` and `revalidatePath` won't work

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -16,8 +16,7 @@ import {
 // import { getDerivedTags } from "next/dist/server/lib/incremental-cache/utils";
 import path from "path";
 
-import { loadBuildId } from "./config/util.js";
-import { awsLogger, debug, error } from "./logger.js";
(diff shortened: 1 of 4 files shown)


khuezy (approved):
LGTM

PR #259, opened 2023-10-05
Allow sharp version override in build

### LINKED ISSUE #243: Allow specification of which `sharp` version to use.
I'd like to use `open-next` for my job, but we have an internal NPM registry with only specific versions of packages available based on auto-scanning. Unfortunately, the version of `sharp` currently hard-coded in the install command of the image optimization build function (0.32.5) is, at least right now, blocked. It might make more sense to make `sharp` a peer dependency or to add to the build interface allowing us to specify which version of `sharp` to use.

This is the build command I'm referring to: https://github.com/sst/open-next/blob/main/packages/open-next/src/build.ts#L327

lasley:
Update build.ts to allow environment variable driven override of the sharp version, which fixes #243


khuezy (approved):
Thanks!

PR #261, opened 2023-10-07
fix cookie redirect matcher

khuezy:
When `cookie['key-does-not-exist-and-will-be-undefined']`, the cookie matcher logic would return `true` because `undefined !== ""`

See: `cookies?.[redirect.key] !== ""`

### DIFF
--- a/packages/open-next/src/adapters/routing/matcher.ts
+++ b/packages/open-next/src/adapters/routing/matcher.ts
@@ -22,14 +22,14 @@ const routeHasMatcher =
     switch (redirect.type) {
       case "header":
         return (
-          headers?.[redirect.key.toLowerCase()] !== "" &&
+          headers?.[redirect.key.toLowerCase()] &&
           new RegExp(redirect.value ?? "").test(
             headers[redirect.key.toLowerCase()] ?? "",
           )
         );
       case "cookie":
         return (
-          cookies?.[redirect.key] !== "" &&
+          cookies?.[redirect.key] &&
           new RegExp(redirect.value ?? "").test(cookies[redirect.key] ?? "")
         );
       case "query":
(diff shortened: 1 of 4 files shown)


packages/open-next/src/adapters/routing/matcher.ts
```diff
@@ -22,14 +22,14 @@ const routeHasMatcher =
     switch (redirect.type) {
       case "header":
         return (
-          headers?.[redirect.key.toLowerCase()] !== "" &&
```
khuezy, line 25: @conico974 Should this be updated too?

conico974: This probably needs to be updated as well, but i think we should do something like 
```ts
headers?.[redirect.key.toLowerCase()] && headers?.[redirect.key.toLowerCase()] !== ""
```

khuezy: `!!headers?.[redirect.key.toLowerCase()]` should take care of both those conditions right? `undefined` and empty string will both be false.

yarinsa:
https://github.com/sst/open-next/issues/263 Should solve this issue

packages/open-next/src/adapters/routing/matcher.ts
khuezy: @conico974 gonna merge this unless I should make your suggested changes.

yarinsa:
What is the ETA for merging this?

khuezy:
There's 3 issues we'd like to get in for 2.2.2, not sure when that will get released, perhaps this week.

yarinsa:
Thank you

yarinsa:
@khuezy Why was it reverted?

khuezy:
> @khuezy Why was it reverted?

If you click on the revert PR, there's a description there: https://github.com/sst/open-next/pull/267
The issue was fix and merged into `main` but if you're still having issues then something's up.

The test app: https://github.com/sst/open-next/blob/main/examples/app-router/next.config.js
The e2e test: https://github.com/sst/open-next/blob/main/packages/tests-e2e/tests/appRouter/config.redirect.test.ts

PR #273, opened 2023-10-10
fix: pass through headers on redirect

### LINKED ISSUE #272: Middleware redirect response ignoring headers
The following response works perfectly fine when running locally
```ts
return NextResponse.redirect("some_url", {
  headers: {
    "set-cookie": "random=123",
    "X-Test": "123",
  },
});
```
However, when deployed to AWS none of these headers are returned from cloudfront.

relsunkaev:
resolves #272

### DIFF
--- a/packages/open-next/src/adapters/routing/middleware.ts
+++ b/packages/open-next/src/adapters/routing/middleware.ts
@@ -86,23 +86,6 @@ export async function handleMiddleware(
     onWarning: console.warn,
   });
   res.statusCode = result.response.status;
-  // If the middleware returned a Redirect, we set the `Location` header with
-  // the redirected url and end the response.
-  if (res.statusCode >= 300 && res.statusCode < 400) {
-    const location = result.response.headers
-      .get("location")
-      ?.replace("http://localhost:3000", `https://${req.headers.host}`);
-    // res.setHeader("Location", location);
-    return {
-      body: "",
-      type: internalEvent.type,
-      statusCode: res.statusCode,
-      headers: {
-        Location: location,
-      },
-      isBase64Encoded: false,
-    };
-  }
 
   /* Apply override headers from middleware
     NextResponse.next({
@@ -132,6 +115,25 @@ export async function handleMiddleware(
       res.setHeader(key, value);
     }
   });
+
+  // If the middleware returned a Redirect, we set the `Location` header with
+  // the redirected url and end the response.
+  if (res.statusCode >= 300 && res.statusCode < 400) {
+    const location = result.response.headers
+      .get("location")
+      ?.replace("http://localhost:3000", `https://${req.headers.host}`);
+    // res.setHeader("Location", location);
+    return {
+      body: "",
+      type: internalEvent.type,
+      statusCode: res.statusCode,
+      headers: {
+        ...resHeaders,
+        Location: location,
+      },
+      isBase64Encoded: false,
+    };
+  }
 
   // If the middleware returned a Rewrite, set the `url` to the pathname of the rewrite
   // NOTE: the header was added to `req` from above
--- a/examples/app-pages-router/middleware.ts
+++ b/examples/app-pages-router/middleware.ts
@@ -7,7 +7,9 @@ export function middleware(request: NextRequest) {
   const protocol = host?.startsWith("localhost") ? "http" : "https";
   if (path === "/redirect") {
     const u = new URL("/redirect-destination", `${protocol}://${host}`);
-    return NextResponse.redirect(u);
+    return NextResponse.redirect(u, {
+      headers: { "set-cookie": "test=success" },
+    });
   } else if (path === "/rewrite") {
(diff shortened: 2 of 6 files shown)


khuezy (approved):
Thanks! LGTM

@conico974 @fwang 
Do we want to send ALL the headers back? There may be some headers that we want to omit?

packages/open-next/src/adapters/routing/matcher.ts
```diff
@@ -180,6 +180,7 @@ export function handleRedirects(
       type: event.type,
       statusCode: 308,
       headers: {
+        ...event.headers,
```
conico974, line 183: Unless i'm missing something, we don't want to pass `event.headers` here.
`handleRedirects` are for `next.config.js` redirects and `event.headers` in this case are request headers, not response headers.

packages/open-next/src/adapters/routing/middleware.ts
```diff
@@ -32,7 +32,7 @@ type MiddlewareOutputEvent = InternalEvent & {
 
 //    if res.end() is return, the parent needs to return and not process next server
 export async function handleMiddleware(
-  internalEvent: InternalEvent,
+  internalEvent: InternalEvent
```
conico974, line 35: There is a bunch of changes like that inside your PR.
It might be your local prettier messing with our eslint config. You can run `pnpm run lint:fix` to solve this

conico974 (changes requested):

packages/open-next/src/adapters/routing/middleware.ts
```diff
@@ -98,6 +98,7 @@ export async function handleMiddleware(
       type: internalEvent.type,
       statusCode: res.statusCode,
       headers: {
+        ...res.headers,
```
conico974, line 101: `res.headers` is empty at this point, you probably want to use `resHeaders` instead, but you'll need to do it after they've been set here https://github.com/sst/open-next/blob/2014a1b67f09049b0d8b47f47054adebe24c7a77/packages/open-next/src/adapters/routing/middleware.ts#L126-L135

conico974 (changes requested):
I've run the e2e test locally and it failed, did you manage to make it work this way ?

packages/open-next/src/adapters/routing/middleware.ts
khuezy: Good catch. I used the responseHeaders and that worked for me

khuezy: @relsunkaev I don't think we should send back all the headers, it will include all the AMZ geo data too... We should just return `result.response.headers.get('set-cookie') ?? ''`

conico974: That's why i think `resHeaders` is fine it's already what you would have get if it wasn't a redirect, we need to allow people to pass the headers they want, maybe they'd want to pass an api key or an auth header

khuezy:
You'll have to deploy your app and run your e2e test against your deployed example apps.
I'll update the docs w/ better instructions... but for now:
1) `cd examples/sst`
2) `npx sst deploy --stage e2e`
3) export the env
```
          export APP_ROUTER_URL=$(jq -r '.["e2e-example-AppRouter"].url' .sst/outputs.json)
          export PAGES_ROUTER_URL=$(jq -r '.["e2e-example-PagesRouter"].url' .sst/outputs.json)
          export APP_PAGES_ROUTER_URL=$(jq -r '.["e2e-example-AppPagesRouter"].url' .sst/outputs.json)
```
4) `cd packages/tests-e2e`
5) Add `test.only` to isolate your test
6) `npm run e2e:dev`

conico974:
> The tests I wrote with the "set-cookie" header passed but I now realize the way I ran them doesn't test anything. How do you run e2e tests on the output built by open next locally?

@relsunkaev you'll need to deploy with sst. 
Inside `examples/sst` run `pnpm sst deploy --stage test`. Make sure that open-next is properly build before running this command.
And then you'll have to run the e2e tests against the deployed examples. 
I have a little script to make things easier that needs to be run under `packages/tests-e2e`: 
```bash
APP_ROUTER_URL="https://xxxxxxxxxx.cloudfront.net"
echo "APP_ROUTER_URL=$APP_ROUTER_URL"
export APP_ROUTER_URL
PAGES_ROUTER_URL="https://xxxxxxxxxx.cloudfront.net"
echo "PAGES_ROUTER_URL=$PAGES_ROUTER_URL"
export PAGES_ROUTER_URL
APP_PAGES_ROUTER_URL="https://xxxxxxxxxx.cloudfront.net"
echo "APP_PAGES_ROUTER_URL=$APP_PAGES_ROUTER_URL"
export APP_PAGES_ROUTER_URL
pnpm run e2e:dev
```
Just replace the x with the deployed value from the sst deployment

khuezy:
Thanks I'll let @conico974 approve then we can merge.

conico974 (approved):
LGTM

PR #280, opened 2023-10-13
TrailingSlash using rewrite for a fetch call results on infinite loop

AlbertSabate:
Fixes: https://github.com/sst/open-next/issues/279

### DIFF
--- a/packages/open-next/src/adapters/routing/matcher.ts
+++ b/packages/open-next/src/adapters/routing/matcher.ts
@@ -176,11 +176,12 @@ export function handleRedirects(
     !event.rawPath.endsWith("/") &&
     !event.headers["x-nextjs-data"]
   ) {
+    const headersLocation = event.url.split('?');
     return {
       type: event.type,
       statusCode: 308,
       headers: {
-        Location: event.url + "/",
+        Location: `${headersLocation[0]}/${headersLocation[1] ? `?${headersLocation[1]}` : ''}`,
       },
       body: "",
       isBase64Encoded: false,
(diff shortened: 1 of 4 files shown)


packages/open-next/src/adapters/routing/matcher.ts
```diff
@@ -191,11 +192,12 @@ export function handleRedirects(
     event.rawPath.endsWith("/") &&
     event.rawPath !== "/"
   ) {
+    const headersLocation = event.url.split('?');
     return {
       type: event.type,
       statusCode: 308,
       headers: {
-        Location: event.url.replace(/\/$/, ""),
+        Location: `${headersLocation[0].replace(/\/$/, "")}${headersLocation[1] ? `?${headersLocation[1]}` : ''}`,
```
khuezy, line 200: Should the internalEvent.url in the final return below also split the `?`?

AlbertSabate: I code-reviewed that part, and it is already handled by handleRewrites fn.
On handleRewrite, we are doing it like this:
```
const urlQueryString = new URLSearchParams(convertQuery(query)).toString();
const queryString = urlQueryString ? `?${urlQueryString}` : "";
let rewrittenUrl = rawPath;
url: `${rewrittenUrl}${queryString}`,
```

This should be working as expected because we do not add or remove any slash. On my changes, I've gone for split('?') because there are fewer lines of code. But happy to match the style. I'm not sure if there are some code style guidelines somewhere :D

khuezy: Should be fine, doesn't have to match style.

khuezy (approved):
Thanks!

AlbertSabate:
Fixed eslint.

Issue #286, opened 2023-10-17
Trailing slash appended to robots.txt and sitemap.xml


leanderiversen:
Hi again,

I have found a new issue regarding trailing slashes, and can confirm that it is present in at least `2.2.3` and `2.2.4`.  The `robots.txt (app/robots.ts)` and `sitemap.xml (app/sitemap.ts)` files get a trailing slash appended, when they were previously left alone.

Happy to provide more details if required.

Thank you.

khuezy:
Are these special files that shouldn't have a trailing slash, even if `trailingSlash` is true?

leanderiversen:
@khuezy That is correct. They should "live" at root level, and not have a trailing slash, even if `trailingSlash` is true.

The documentation can be found [here](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap) and [here](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/robots).

khuezy:
Cool thanks, are there any other special files that need to behave like this?

leanderiversen:
It looks like there is [app/manifest.ts](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/manifest) as well, but I think that is the complete list.

PR #287, opened 2023-10-17
Exclude manifest.json, robots.txt and sitemap.xml from routing matcher

leanderiversen:
Fix for #286

### DIFF
--- a/packages/open-next/src/adapters/routing/matcher.ts
+++ b/packages/open-next/src/adapters/routing/matcher.ts
@@ -175,7 +175,7 @@ export function handleRedirects(
     NextConfig.trailingSlash &&
     !event.headers["x-nextjs-data"] &&
     !event.rawPath.endsWith("/") &&
-    !["/manifest.json", "/robots.txt", "/sitemap.xml"].includes(event.rawPath)
+    !event.rawPath.match(/[\w-]+\.[\w]+$/g)
   ) {
     const headersLocation = event.url.split("?");
     return {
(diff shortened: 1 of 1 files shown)


packages/open-next/src/adapters/routing/matcher.ts
```diff
@@ -173,8 +173,13 @@ export function handleRedirects(
 ): InternalResult | undefined {
   if (
     NextConfig.trailingSlash &&
+    !event.headers["x-nextjs-data"] &&
     !event.rawPath.endsWith("/") &&
-    !event.headers["x-nextjs-data"]
+    ![
```
khuezy, line 178: Do you need to apply this to the block below too?

leanderiversen: It shouldn't need the file checks, if that's what you mean, as the next block would only be triggered if `trailingSlash` is disabled.

khuezy: If there's linting issue, you can run `npm run lint:fix`

leanderiversen: Sorry about that. It should be fixed now.

conico974 (changes requested):
I've run some test with `next start` and it looks like next is removing trailing slash for every path that ends with an extension ( In fact they do it for everything that ends with `.SOMETHING`)
We should probably do something similar. A regex like that `/[\w-]+\.[\w]+$/g` should do the trick

leanderiversen:
@conico974 Thank you for the feedback. This should be updated now.

conico974 (approved):
LGTM

khuezy:
@leanderiversen do you need this release ASAP? If not, we'll include it in the next once the ~3 PRs are merged.

leanderiversen:
@khuezy No, I am happy to wait. Thank you for checking!

PR #291, opened 2023-10-20
Fix regression on s3 fetch caches

leonchabbey:
This aims to fix a regression from https://github.com/sst/open-next/pull/252.
Following that PR, it became impossible to list a key of type "fetch" as the prefix in listS3Object always added a "dot" at the end which is only useful for non-fetch objects.

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -127,9 +127,9 @@ export default class S3Cache {
     }
     const isFetchCache =
       typeof options === "object" ? options.fetchCache : options;
-    const keys = await this.listS3Object(key);
-    if (keys.length === 0) return null;
+    const keys = await this.listS3Object(key, isFetchCache);
     debug("keys", keys);
+    if (keys.length === 0) return null;
     return isFetchCache
       ? this.getFetchCache(key, keys)
       : this.getIncrementalCache(key, keys);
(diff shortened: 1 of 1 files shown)


khuezy:
Thanks! Is this related to a recent latency others and myself have been seeing?

leonchabbey:
> Thanks! Is this related to a recent latency increase others and myself have been seeing?

Most probably yes!

It couldn't find any fetch cache in S3 because the prefix never matched the structure of a fetch cache path (the path was missing `__fetch` and had an extra `.` at the end)

This problem was particularly noticeable in dynamic routes (no ISR) because it's as if there were no fetch cache at all.

conico974 (commented):
>It couldn't find any fetch cache in S3 because the prefix never matched the structure of a fetch cache path (the path was missing __fetch and had an extra . at the end)
>
>This problem was particularly noticeable in dynamic routes (no ISR) because it's as if there were no fetch cache at all.

Good catch and thanks for the PR. 

Unfortunately this PR will become useless with #295. Not sure when we'll merge #295, if we wait for `2.2.6`, we should merge this one for sure

leonchabbey:
> > It couldn't find any fetch cache in S3 because the prefix never matched the structure of a fetch cache path (the path was missing __fetch and had an extra . at the end)
> > This problem was particularly noticeable in dynamic routes (no ISR) because it's as if there were no fetch cache at all.
> 
> Good catch and thanks for the PR.
> 
> Unfortunately this PR will become useless with #295. Not sure when we'll merge #295, if we wait for `2.2.6`, we should merge this one for sure

Nice! I actually did today a bigger refactor than this PR, quite similar to yours, while trying to understand/optimise the requests after seeing how not great to fetch performances are between a Lambda and S3 (btw I posted a question on discord about this to see if other people have seen the same thing: https://discord.com/channels/983865673656705025/1027265626085019769/1165983372590460939).

I'll close this PR then and hope yours is merged soon!

PR #293, opened 2023-10-21
Double chunk DDB batch writes to not overwhelm DDB on load

dkershner6:
For larger pre-rendered sites, throwing thousands of requests at DDB immediately overwhelms the throughput.

This change mitigates this.

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -15,6 +15,7 @@ import {
 import path from "path";
 
 import { debug, error } from "./logger.js";
+import { chunk } from "./util.js";
 
 interface CachedFetchValue {
   kind: "FETCH";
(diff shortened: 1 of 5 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -420,7 +421,7 @@ export default class S3Cache {
     try {
       if (disableDynamoDBCache) return;
       await Promise.all(
-        this.chunkArray(req, 25).map((Items) => {
+        chunk(req, 25).map((Items) => {
```
khuezy, line 424: Use `MAX_DYNAMO_BATCH_WRITE_ITEM_COUNT` here instead of `25`?

khuezy (approved):
LGTM thanks!

packages/open-next/src/adapters/dynamo-provider.ts
```diff
@@ -31,45 +36,56 @@ export async function handler(
   }
 }
 
+/**
+ * Sending to dynamo X commands at a time, using about X * 25 write units per batch to not overwhelm DDB
+ * and give production plenty of room to work with. With DDB Response times, you can expect about 10 batches per second.
+ */
+const DYNAMO_BATCH_WRITE_COMMAND_CONCURRENCY = 4;
```
conico974, line 43: Is there a reason why it's hardcoded to 4 ?
We could add an env variable to modify this behaviour, but keep it at 4 by default.

PR #295, opened 2023-10-23
Perf Reduce s3 calls

conico974:
The goal of this PR is to reduce the number of S3 call that we need to make for the S3 incremental cache

Before this PR there was 1 `ListObjectsV2Command` and 2 to 3 `GetObjectCommand` for every `get` of the incremental cache, this has been reduced to a single `GetObjectCommand`
For the `set` part of the incremental cache, we go from 2-3 `PutObjectCommand` to a single one.

This also reduce the number of objects in the S3 cache

TODO: 
- [x] Update docs
- [ ] Potentially filter the key not found error

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -9,7 +9,6 @@ import {
   ListObjectsV2Command,
   PutObjectCommand,
   PutObjectCommandInput,
-  PutObjectCommandOutput,
   S3Client,
 } from "@aws-sdk/client-s3";
 import path from "path";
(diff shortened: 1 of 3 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -160,96 +177,59 @@ export default class S3Cache {
...
-    // after having been redirected, we'll get the page data
-    if (keys.includes(this.buildS3Key(key, "redirect"))) {
-      debug("get redirect cache", { key });
-      try {
-        const { Body, LastModified } = await this.getS3Object(
-          key,
-          "redirect",
-          keys,
-        );
+      } else if (cacheData.type === "redirect") {
```
khuezy, line 217: Is there a case for `rewrites`?

packages/open-next/src/adapters/cache.ts
```diff
@@ -258,37 +238,44 @@ export default class S3Cache {
...
+          meta: { status: data.status, headers: data.headers },
+        } as S3CachedFile),
+      );
     } else if (data?.kind === "FETCH") {
       await this.putS3Object(key, "fetch", JSON.stringify(data));
     } else if (data?.kind === "REDIRECT") {
-      // delete potential page data if we're redirecting
-      await this.deleteS3Objects(key);
-      await this.putS3Object(key, "redirect", JSON.stringify(data));
+      // // delete potential page data if we're redirecting
```
khuezy, line 270: This comment may be confusing, since the command is `PUT` and not `DELETE`. Perhaps you can provide more context?

packages/open-next/src/adapters/cache.ts
```diff
@@ -523,7 +508,7 @@ export default class S3Cache {
 
   private async deleteS3Objects(key: string) {
     try {
-      const regex = new RegExp(`\.(json|rsc|html|body|meta|fetch|redirect)$`);
+      const regex = new RegExp(`\.(fetch|cache)$`);
```
khuezy, line 511: I think this is handled in Derek's PR but the `\.` should have been `\\.` to escape the RegExp constructor. Since there's no dynamic values here, it might be better just to use the inline one, eg: `/\.(fetch|cache)$/`

khuezy (approved):
I deployed w/ this branch and everything looks good.
There are a few conflicts from Derek's PR though

packages/open-next/src/adapters/cache.ts
conico974: No this is only for `redirect` inside `getStaticProps` or inside an ISR or SSG app route

packages/open-next/src/adapters/cache.ts
conico974: My bad, i forgot to remove the comment

PR #299, opened 2023-10-27
Fix external rewrite

conico974:
This PR should fix a couple of external rewrite issue

- [x] External rewrite return empty data
- [x] External rewrite from middleware
- [x] Middleware response error on next 14
- [ ] External rewrite with `transfer-encoding: chunked`

### DIFF
--- a/packages/open-next/src/adapters/plugins/routing/default.replacement.ts
+++ b/packages/open-next/src/adapters/plugins/routing/default.replacement.ts
@@ -59,15 +59,16 @@ export const processInternalEvent: ProcessInternalEvent = async (
     internalEvent = middleware;
   }
 
-  let isExternalRewrite = false;
-  // First rewrite to be applied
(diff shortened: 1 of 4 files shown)


conico974:
There is still an issue with external rewrites when the target use streaming. Somehow it totally broke the encoding.
I've tried doing this code locally and on the lambda and i get totally different result: 
```ts
const result = await fetch("URL_TO_CHUNKED_ROUTE")
const text = await result.text()
console.log(text)
```
Locally i get the expected result, on the lambda it is just gibberish. 
This cause cloudfront to return 500 with `InternalFailure` as cause.

conico974:
@khuezy i think we should merge this PR even if there is still an issue with external rewrites with chunked encoding. What do you think ?

khuezy:
Yes that should be fine for now. I can help investigate after but this should unblock some people.

PR #300, opened 2023-10-27
fix: XML Malformed Error DeleteObjectsCommand

dkershner6:
https://discord.com/channels/983865673656705025/1167435351577604203

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -15,7 +15,7 @@ import {
 import path from "path";
 
 import { MAX_DYNAMO_BATCH_WRITE_ITEM_COUNT } from "./constants.js";
-import { debug, error } from "./logger.js";
+import { debug, error, warn } from "./logger.js";
 import { chunk } from "./util.js";
 
 interface CachedFetchValue {
@@ -517,10 +517,16 @@ export default class S3Cache {
 
   private async deleteS3Objects(key: string) {
     try {
-      const regex = new RegExp(`\.(json|rsc|html|body|meta|fetch|redirect)$`);
+      const regexString = `\.(json|rsc|html|body|meta|fetch|redirect)$`;
+      const regex = new RegExp(regexString);
       const s3Keys = (await this.listS3Object(key)).filter(
         (key) => key && regex.test(key),
       );
+
+      if (s3Keys.length === 0) {
+        warn(`No s3 keys matching ${regexString} found for ${key}`);
+        return;
+      }
 
       await this.client.send(
         new DeleteObjectsCommand({
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -100,7 +100,7 @@ type CacheExtension =
 /** Beginning single backslash is intentional, to look for the dot + the extension. Do not escape it again. */
 const CACHE_EXTENSION_REGEX = /\.(json|html|rsc|body|meta|fetch|redirect)$/;
 
-function hasCacheExtension(key: string) {
+export function hasCacheExtension(key: string) {
   return CACHE_EXTENSION_REGEX.test(key);
 }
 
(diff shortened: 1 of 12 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -517,11 +517,17 @@ export default class S3Cache {
 
   private async deleteS3Objects(key: string) {
     try {
-      const regex = new RegExp(`\.(json|rsc|html|body|meta|fetch|redirect)$`);
+      const regexString = `\.(json|rsc|html|body|meta|fetch|redirect)$`;
```
khuezy, line 520: The `\.` needs to be escaped: `\\.`

packages/open-next/src/adapters/cache.ts
```diff
@@ -517,11 +517,17 @@ export default class S3Cache {
...
+      const regex = new RegExp(regexString);
       const s3Keys = (await this.listS3Object(key)).filter(
         (key) => key && regex.test(key),
       );
 
+      if (s3Keys.length === 0) {
+        warn(`No s3 keys matching ${regexString} found for ${key}`);
```
khuezy, line 527: You can keep the origin `regex`, it should print out the regex string b/c the regex implements `toString()`

packages/open-next/src/adapters/cache.ts
dkershner6: I was actually thinking this...but I figured this wouldve come up before now...

packages/open-next/src/adapters/cache.ts
dkershner6: Our explicit vs implicit meet again, im gonna stick with explicit since it has always served me well from an understandability standpoint. If you feel strongly, we can poll some others.

I will leave this unresolved as a reminder, you may resolve or not as you see fit.

khuezy (approved):
LGTM, looks like the commit history has stuff from your old branch, not sure if that'll cause any weird history.

dkershner6:
> LGTM, looks like the commit history has stuff from your old branch, not sure if that'll cause any weird history.

I saw this too, but I had done a full merge.  I think it is a GitHub display issue between forks, but these are in fact already merged, so it should be fine.

packages/open-next/src/adapters/cache.ts
khuezy: imo `const regex = new RegExp(`\.(json|rsc|html|body|meta|fetch|redirect)$`);` is cleaner b/c it's 1 less line and it doesn't have the `\\.` which can be strange if you're not too familiar w/ regex.

You can still call the toString explicitly, eg `warn(`No s3 keys matching ${regex.toString()} found for ${key}`);`

khuezy: On that note, there's not any dynamic value in `new Regex`, so it coulda have been the RegExp literal, eg `/\.(json|...)$`

dkershner6: Ran some tests, and the single backslash was correct, but since (I think) you read `apostrophe` where was `backtick`, went to the drawing board and attempted a clear, consise, explicit compromise that no one can get confused over.  Resolving here, post another convo if you disagree.

khuezy: Thanks! LGTM. I got confused, I thought the original was using the regex literal.
When using `new RegExp`, you have to double escape the `\.`, eg
```
const regexString = `\.(json|rsc|html|body|meta|fetch|redirect)$`;
const regex = new RegExp(regexString);

console.log('should be true: ', regex.test('hello.json'))
console.log('should be false: ', regex.test('hello,json')) // this returns true because the regex is "." instead of "\."
```

dkershner6: Went even further. ;)

khuezy: lol thanks! It's not for me, it's going to be for everyone using open-next

PR #301, opened 2023-10-31
fix: copy favicon.ico from app dir

### LINKED ISSUE #158: Next 13 Favicon file
Next 13 allows you to create a `favicon.ico` file at `/app`. This will then automatically add a `<link>` to the head of the document for you. Docs [here](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons)

I have found that the favicon file is not included in the static assets folder of `.open-next` and is not being uploaded to the S3 bucket when using `npx sst deploy`.

annleefores:
Fixes #158 

This PR adds functionality to check for the existence of favicon.ico in the `src/app/` or `app/` directory and, if present, copies it to the `.open-next/assets` directory.

### DIFF
--- a/packages/open-next/src/build.ts
+++ b/packages/open-next/src/build.ts
@@ -377,7 +377,7 @@ function createStaticAssets() {
   // - .next/BUILD_ID => _next/BUILD_ID
   // - .next/static   => _next/static
   // - public/*       => *
-  // - src/app/favicon.ico => favicon.ico
+  // - app/favicon.ico or src/app/favicon.ico  => favicon.ico
   fs.copyFileSync(
     path.join(appBuildOutputPath, ".next/BUILD_ID"),
     path.join(outputPath, "BUILD_ID"),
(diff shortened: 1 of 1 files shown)


packages/open-next/src/build.ts
```diff
@@ -377,6 +377,7 @@ function createStaticAssets() {
   // - .next/BUILD_ID => _next/BUILD_ID
   // - .next/static   => _next/static
   // - public/*       => *
+  // - src/app/favicon.ico => *
```
khuezy, line 380: A technicality 🤓 , it's a single file so the "*" is incorrect.

annleefores: yeah, I just didn't think it through. I'll fix it

khuezy: Thanks! Maybe not relevant now, but will we need to do this for `robots.txt`? We can kick that can until someone needs it.

annleefores: Next.js already bundles all the metadata files, and they are even accessible through SST deployments. So, I don't think there's a need to do the same for robots.txt. However, I did have some trouble getting opengraph-images to work in the same way. If you are open to adding this feature, I'm down to figure out a solution.

packages/open-next/src/build.ts
```diff
@@ -389,6 +390,12 @@ function createStaticAssets() {
   if (fs.existsSync(appPublicPath)) {
     fs.cpSync(appPublicPath, outputPath, { recursive: true });
   }
+
+  const faviconPath = path.join(appPath, "src", "app", "favicon.ico");
```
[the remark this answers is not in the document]
annleefores: Thank you for bringing this up, I'll add this too

khuezy (approved):
Thanks!

PR #304, opened 2023-11-03
fix: setting the right tag values for fetch cache

mathisobadia:
Right now the tag property is always empty, I tried to check in the nextjs code base the reason and i found [this relevant code](https://github.com/vercel/next.js/blob/b8b1ec24a259ac1a42a513971259bda62738dd47/packages/next/src/server/lib/patch-fetch.ts#L479C1-L498C18) 
this code was introduced in [this PR](https://github.com/vercel/next.js/commit/2eef7754728b3db8df8b1f52def0a47c8f0baed5#diff-4c54e369ddb9a2db1eed95fe1d678f94c8e82c540204475d42c78e49bf4f223a), So I'm wondering how we should manage that for older versions of nextjs that don't have that change.

Also ideally I would like to add some tests to the open-next repo to make sure that revalidateTag works ok but I still don't understand fully the setup and how to write that test so this PR is probably a WIP

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -306,7 +306,7 @@ export default class S3Cache {
     // If we use an in house version of getDerivedTags in build we should use it here instead of next's one
     const derivedTags: string[] =
       data?.kind === "FETCH"
-        ? ctx?.tags ?? []
+        ? ctx?.tags ?? data?.data?.tags ?? [] // before version 14 next.js used data?.data?.tags so we keep it for backward compatibility
         : data?.kind === "PAGE"
         ? data.headers?.["x-next-cache-tags"]?.split(",") ?? []
         : [];
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -70,6 +70,14 @@ type IncrementalCacheValue =
   | CachedFetchValue
   | CachedRouteValue;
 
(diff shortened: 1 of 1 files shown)


mathisobadia:
@khuezy I can confirm that this fixes the fetch cache issue in my app with latest nextjs and open-next version after testing it. So I think it can me merged. It also corresponds with what the new docs say are the parameters according to the latest PR https://github.com/vercel/next.js/blob/63ffdea4ab66a69d67d24161c995fd45e7b276f2/docs/02-app/01-building-your-application/08-deploying/index.mdx so we can hope that this will stay stable for a bit

packages/open-next/src/adapters/cache.ts
```diff
@@ -298,7 +310,7 @@ export default class S3Cache {
     // If we use an in house version of getDerivedTags in build we should use it here instead of next's one
     const derivedTags: string[] =
       data?.kind === "FETCH"
-        ? data.data.tags ?? []
+        ? ctx?.tags ?? []
```
khuezy, line 313: Will this break backwards compatibility? Can we do `ctx?.tags ??data?.data?.tags ?? []

conico974: Yeah you're right, that's how it should be done

mathisobadia: I just updated my PR with this change

khuezy:
Thanks! The changes LGTM

PR #305, opened 2023-11-03
[V3] Rewrite Server

conico974:
This PR is my take on how we could rewrite the server for V3 along with some changes to the config.

### Proposition
- We replace our cli args with a single config file `open-next.config.ts`
- We create a single entrypoint for the open-next server which take an `InternalEvent` and output an `InternalResult` 
- Then we create a system of adapters to be able to plug an handler (for aws lambda, or docker for example) and a converter to convert event to the internal event and result
- We replace the `ServerlessResponse` and `StreamingServerResponse`, with a more elegant `OpenNextNodeResponse` which extends `Transform` stream which should work better and can be used both in streaming and non streaming case
- There is also a possibility to override the adapter and the handler via `open-next.config.ts`

### TODO
- [ ] Better handler for docker/node cases
- [ ] Generate dockerfile
- [ ] Write tests
- [ ] Write other adapter and converter
- [x] Find a way to make everything work on edges runtime (i.e. Cloudflare workers...)
- [x] Add override options for queue, incremental cache and revalidate cache
- [ ] Fix `open-next.config.ts` build
- [ ] Lot of cleanup to do
- [ ] Probably a lot of other things


packages/open-next/src/wrappers/docker.ts
```diff
@@ -0,0 +1,27 @@
+import { createServer } from "http";
+
+import { ResponseStream } from "../adapters/http";
+import { Wrapper } from "../adapters/types/open-next";
+
+const wrapper: Wrapper = async (handler, converter) => {
```
khuezy, line 6: Very cool. We should allow for clusters, eg: https://nodejs.org/api/cluster.html
It's just a matter of moving the block below into the `else` block.
In my testing, the vertical scaling of the multi-core added about 25% more throughput than an equivalent horizontal scale.

packages/open-next/src/wrappers/aws-lambda-streaming.ts
```diff
@@ -0,0 +1,49 @@
...
+        // We replace their write with ours which are inside a setImmediate
+        // This way it seems to work all the time
+        // I think we can't ship this code as it is, it could break at anytime if they decide to change the runtime and they already did it in the past
+        responseStream.setContentType(
+          "application/vnd.awslambda.http-integration-response",
+        );
+        const prelude = JSON.stringify(_prelude);
+
+        // Try to flush the buffer to the client to invoke
+        // the streaming. This does not work 100% of the time.
```
khuezy, line 29: Our latest code had the streaming working 100% of the time now.

packages/open-next/src/wrappers/docker.ts
conico974: Good idea, i haven't looked too much into it, is there any downsides ?

Anyway it should be really easy to customize this according to your needs. 
If you want to override this and use your own, you just have to create something that respect `Wrapper` interface and import it in `open-next.config.ts`.

And if/when we add splitting, each split could have different wrapper and converter.
Your api could be on docker, your ISR and SSG on lambda and your SSR on the edge.

packages/open-next/src/wrappers/aws-lambda-streaming.ts
conico974: I haven't tested streaming much, but it seems to work well for now.
The comment is from a part i copied from the `serverless-replacement.ts` file.

I haven't tested SSE yet, but the way it is done now is how it should be done.

packages/open-next/src/wrappers/docker.ts
khuezy: Not that I'm aware of, it just makes use of all the CPUs on the machine. All requests are distributed on the same port so there's no behavioral changes.

packages/open-next/src/wrappers/aws-lambda-streaming.ts
conico974: I had to add some fix, but now streaming is as reliable as it was before (Probably more robust since there is no more hack with `setImmediate`)
I just need to add compression again, but this should not be too hard

PR #306, opened 2023-11-04
fix: Just reducing error log spam a bit

dkershner6:


packages/open-next/src/adapters/cache.ts
```diff
@@ -490,7 +490,7 @@ export default class S3Cache {
       );
       return result;
     } catch (e) {
-      error("This error can usually be ignored : ", e);
+      warn("This error can usually be ignored : ", e);
```
khuezy, line 493: How often are you seeing this error? If it's usually ignorable, then should we just use `debug` or omit the log altogether?

dkershner6: It seems to be in weird scenarios, like when sustaining attacks with weird things on the end of URLs (\ is common).

All in all, not too much of an issue, I can ignore on our side, but I doubt we are the only website on the internet to get attacked by bots.

dkershner6: I also think `warn` is appropriate.  Something weird happened, but no need to get out of bed in the middle of the night (unlike error).

khuezy (approved):

packages/open-next/src/adapters/cache.ts
khuezy: Is it a denial of wallet attack? Is the S3 LIST adding up on your account?

dkershner6: Seems more like probing for now, but yes, I would assume it is to try and throw a wrench into some database call or something.

dkershner6:
I added one more warn.  This error log is hit during the same flow after a mocked response and null coalesce.

packages/open-next/src/adapters/cache.ts
conico974: We should probably modify the `awslogger` that we pass to S3 as well.
https://github.com/sst/open-next/blob/e32af92b30738a52066208b8ea39a71fdb3be67f/packages/open-next/src/adapters/server-adapter.ts#L35

dkershner6: > We should probably modify the `awslogger` that we pass to S3 as well.
> 
> https://github.com/sst/open-next/blob/e32af92b30738a52066208b8ea39a71fdb3be67f/packages/open-next/src/adapters/server-adapter.ts#L35

Done.

dkershner6:
Added an interface for downplaying AWS SDK errors, with just the one for now, at @conico974 's request.

PR #310, opened 2023-11-07
SDK maxAttempts now configurable through env, plugin logs now debug

dkershner6:
https://discord.com/channels/983865673656705025/1168586969543082237

### DIFF
--- a/packages/open-next/src/adapters/server-adapter.ts
+++ b/packages/open-next/src/adapters/server-adapter.ts
@@ -1,21 +1,20 @@
-/* eslint-disable unused-imports/no-unused-imports */
-import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
-import { S3Client } from "@aws-sdk/client-s3";
+import { DynamoDBClient, DynamoDBClientConfig } from "@aws-sdk/client-dynamodb";
+import { S3Client, S3ClientConfig } from "@aws-sdk/client-s3";
 
-// We load every config here so that they are only loaded once
-// and during cold starts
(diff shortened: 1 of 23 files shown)


packages/open-next/src/adapters/server-adapter.ts
```diff
@@ -24,20 +23,40 @@ declare global {
 
 const CACHE_BUCKET_REGION = process.env.CACHE_BUCKET_REGION;
 
+function parseS3ClientConfigFromEnv(): S3ClientConfig {
+  return {
+    region: CACHE_BUCKET_REGION,
+    logger: awsLogger,
+    maxAttempts: parseNumberFromEnv(process.env.AWS_SDK_S3_MAX_ATTEMPTS),
```
khuezy, line 30: Is the "defaults to 3" an AWS default config?

dkershner6: Yessir.  Effectively `undefined` === 3.  I could change the docs to be "defaults to whatever the SDK defaults to" if that is preferred, but I think that is the direction I would take.

khuezy (approved):

packages/open-next/src/adapters/server-adapter.ts
khuezy: Thx, that's fine, I just wanted to make sure we didn't miss the defaults.

conico974:
Frank was faster than you, he already merged this one https://github.com/sst/open-next/pull/311

PR #313, opened 2023-11-07
chore: update sst version

khuezy:


examples/sst/stacks/AppRouter.ts
```diff
@@ -7,6 +7,9 @@ export function AppRouter({ stack }) {
     bind: [],
     environment: {},
     timeout: "20 seconds",
+    experimental: {
```
conico974, line 10: We can use the default `NextJsSite` here, no need to override anything anymore

conico974 (approved):
Apart from my comment, it looks good to me

examples/sst/stacks/AppRouter.ts
khuezy: Oh yea good call, updating...

PR #317, opened 2023-11-17
Fix tags cache

conico974:
This PR should fixes 2 issues from using `revalidateTag`
- A regression was introduced in #295 which made tags not being generated at build time for any route
- fetch cache is never read on next@14+, they changed the interface again

This PR also adds a test for `revalidateTag`

### DIFF
--- a/.changeset/poor-ravens-attack.md
+++ b/.changeset/poor-ravens-attack.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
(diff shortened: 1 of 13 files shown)


khuezy (approved):

sladg:
Considering no change in `package.json` is lock file update correct/expect?

conico974:
> Considering no change in `package.json` in lock file update correct/expect?

Yes and no, I had to update next for the e2e tests. 
I'll push a change to pin next to 14.0.3 in `package.json`, we already install latest during e2e so it's fine

Issue #319, opened 2023-11-21
multi-value query parameters behave differently when using open-next


thoaltmann:
We have a few pages and API-routes in our application that use multi-value query parameters with the same key and after updating open-next, this doesn't work as expected anymore.

The NextJS `useSearchParams`-hook and `request.nextUrl.searchParams` in e.g. API routes convert query parameters to an array when they are in the form of `?paramKey=value1&paramKey=value2`

```js
// API route
request.nextUrl.searchParams.getAll("paramKey") // => ["value1", "value2"]

// hook
const searchParams = useSearchParams();
searchParams.getAll("paramKey") // => ["value1", "value2"]
```

However, due to probably https://github.com/sst/open-next/blob/main/packages/open-next/src/adapters/routing/middleware.ts#L67 such query parameters are now transformed to `?paramKey=value1,value2`. With this change, multiple query parameters with the same key are no longer regarded as arrays:

```js
// API route
request.nextUrl.searchParams.getAll("paramKey") // => ["value1,value2"]

// hook
const searchParams = useSearchParams();
searchParams.getAll("paramKey") // => ["value1,value2"]
```

Building and running the Next-app locally works as expected, but as soon as it's deployed, the rewriting of the querystring breaks our application.

khuezy:
Hi thanks for the report. CC: @pbaire 
We had this discussion I think in https://github.com/sst/open-next/pull/281#discussion_r1360901658

khuezy:
@thoaltmann can you put up a PR?

thoaltmann:
@khuezy I can certainly try, I'll start setting up open-next locally tomorrow morning and I'll see how far I can get 😄

conico974:
Fixed in https://github.com/sst/open-next/pull/320

PR #320, opened 2023-11-22
change querystring format for multi value parameters (#319)

thoaltmann:
Changes querystrings with multi-value parameters from `"?key=value1,value2"` to `?key=value1&key=value2` since NextJS understands this format and converts it to an array when reading search params.

### DIFF
--- a/packages/open-next/src/adapters/routing/matcher.ts
+++ b/packages/open-next/src/adapters/routing/matcher.ts
@@ -153,13 +153,11 @@ export function handleRewrites<T extends RewriteDefinition>(
     }
   }
 
-  const queryString = query ? `?${convertToQueryString(query)}` : "";
-
   return {
     internalEvent: {
       ...event,
       rawPath: rewrittenUrl,
-      url: `${rewrittenUrl}${queryString}`,
+      url: `${rewrittenUrl}${convertToQueryString(query)}`,
     },
     __rewrite: rewrite,
     isExternalRewrite,
@@ -237,7 +235,7 @@ export function fixDataPage(internalEvent: InternalEvent, buildId: string) {
       ...internalEvent,
       rawPath: newPath,
       query,
-      url: `${newPath}${query ? "?" + convertToQueryString(query) : ""}`,
+      url: `${newPath}${convertToQueryString(query)}`,
     };
   }
   return internalEvent;
--- a/packages/open-next/src/adapters/routing/matcher.ts
+++ b/packages/open-next/src/adapters/routing/matcher.ts
@@ -11,7 +11,7 @@ import {
   RouteHas,
 } from "../types/next-types";
 import { escapeRegex, unescapeRegex } from "../util";
-import { convertQuery, getUrlParts, isExternal } from "./util";
+import { convertToQueryString, getUrlParts, isExternal } from "./util";
 
 const routeHasMatcher =
   (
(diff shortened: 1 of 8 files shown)


khuezy (approved):
Thanks!

conico974 (changes requested):
4 E2E tests are failing for me: 

image

The redirect doesn't work because there is an additional `?` at the end.
Not sure why the multi query doesn't work

thoaltmann:
I may have an idea but I'm not sure why my change would cause the behavior:

in `getUrlParts`, the `?` character is replaced with `%3F` and then returned as part of the `pathname`.

So, some middleware functions add a parameter with e.g. `url.searchParams.set("rewritten", "true")`. Does that mean, that in `handleRewrites`, the `rewrittenUrl` is actually `/middleware-rewrite%3Frewritten=true` and when other parameters are added, it results in `/middleware-rewrite%F3rewritten=true?otherParam=hello` which somehow gets decoded somewhere so that the final pathname is `/middleware-rewrite?rewritten=true?otherParam=hello`?

But that would have been the behavior before too I think, so I'm not really sure if that's actually the problem. I'm not that familiar with which events contain which fields and when a request ends up in one or the other function, so I'm mostly guessing. Do you have any other ideas?

packages/open-next/src/adapters/routing/matcher.ts
```diff
@@ -154,7 +153,7 @@ export function handleRewrites<T extends RewriteDefinition>(
     }
   }
 
-  const queryString = urlQueryString ? `?${urlQueryString}` : "";
+  const queryString = query ? `?${convertToQueryString(query)}` : "";
```
conico974, line 156: The issue might be here, query here could be `{}`, this will evaluate to true.
This does not explain why `query.test.ts` fails though.
My bet is that it comes from here https://github.com/sst/open-next/blob/8bc075b9696857d60ffe520154b8395c3a730b00/packages/open-next/src/adapters/event-mapper.ts#L126.
You should probably try deploying and add a bunch of console.logs to try to understand why it fails

packages/open-next/src/adapters/routing/util.ts
```diff
@@ -54,13 +54,21 @@ export function convertRes(res: ServerlessResponse) {
...
 }
 
-export function convertQuery(query: Record<string, string | string[]>) {
-  const urlQuery: Record<string, string> = {};
-  Object.keys(query).forEach((k) => {
-    const v = query[k];
-    urlQuery[k] = Array.isArray(v) ? v.join(",") : v;
+/**
+ * Make sure that multi-value query parameters are transformed to
+ * ?key=value1&key=value2&... so that Next converts those parameters
```
khuezy, line 59: `URLSearchParams.toString()` won't automatically prefix the output with "?"

packages/open-next/src/adapters/routing/matcher.ts
```diff
@@ -233,18 +232,12 @@ export function fixDataPage(internalEvent: InternalEvent, buildId: string) {
...
-    });
+
     return {
       ...internalEvent,
       rawPath: newPath,
       query,
-      url: `${newPath}${
-        query ? `?${new URLSearchParams(urlQuery).toString()}` : ""
-      }`,
+      url: `${newPath}${query ? "?" + convertToQueryString(query) : ""}`,
```
khuezy, line 240: I think it might be cleaner to move the ternary logic to the `convertToQueryString` function itself.

`url.search = '?a=a'` would still work w/ the prefixed `?`

thoaltmann:
With the last commit I changed how the `query` for the `internalEvent` is built. It now results in a record like `multiValueQueryStringParameters` from the `APIGatewayProxyEvent` when converting from other events to ensure one unified structure. Now the internal event always has a query object that contains an array of values. e2e-tests are happy too

PR #327, opened 2023-12-01
[V3] Function splitting

conico974:
This PR add support for function splitting in open-next
It is based on the rewrite server PR, if you want to review start at a50cdf7

### Features
- Split next standalone output into multiple bundles.
- Each splitted function could use different wrapper and converter (i.e. 1 on docker, 1 on lambda with streaming, 1 on an ec2 instance or a vps)
- Experimental bundled NextServer. At the moment we use the standalone output, this introduces a lot of unnecessary files into the bundle. With bundled NextServer in my test app, one function could go from 29.9Mb unzipped to 6.3Mb, and that's including almost 1Mb for all the `@aws-sdk` packages
- Reduced cold start, with the decrease in sizes, it also means that cold start is way better. From 2.5 seconds to around 700ms in a best case scenario (A simple api with few deps)

### How to test

You'll need an `open-next.config.ts` file next to your `next.config.js`. A simple example of such a file:

```ts
import type { BuildOptions } from 'open-next/types/open-next'
const config = {
  default: {
    // debug: true,
    override: {
      wrapper: "aws-lambda-streaming",
    },
  },
  // functions: {},
  functions: {
    ssr: {
      routes: ["app/api/isr/route", "app/api/sse/route", "app/api/revalidateTag/route"],
      patterns: ['api/isr', 'api/sse', 'api/revalidateTag'],
      override: {
        wrapper: "aws-lambda-streaming",
      },
      experimentalBundledNextServer: true
    },
    pageSsr: {
      routes: ["pages/pageSsr"],
      patterns: [ 'pageSsr', "_next/data/BUILD_ID/pageSsr.json"],
      override: {
        wrapper: "node",
        converter: "node",
        generateDockerfile: true,
      },
    },
    edge: {
      // Edge runtime support only app dir routes and page
      runtime: "edge",
     // You could deploy edge runtime both "regional" or "global" (i.e. lambda@edge)
      placement: "regional",
      // Only one route at a time allowed for edge runtime
      routes: ["app/ssr/page"],
      patterns: ["ssr"],
      override: {}
    }
    
  },
  buildCommand: "echo 'hello world'"
} satisfies BuildOptions

module.exports = config;
export type OpenNextConfig = typeof config
```
You can now use the corresponding PR in sst to test this https://github.com/sst/sst/pull/3567

### TODO
- [x] External middleware/routing (Necessary for internal rewrites to work correctly with splitted function)
- [x] Support edge runtime (On aws you could deploy the edge runtime in lambda@edge or in regular lambda)
- [ ] Add tests 
- [x] Make warmer works with splitted function
- [x] Write corresponding PR in sst or in ion


packages/open-next/src/adapters/warmer-function.ts
```diff
@@ -27,49 +21,74 @@ const resolveWarmerInvoke = async () => {
...
-                ),
-              }),
-            );
-          } catch (e) {
-            error(`failed to warm up #${i}`, e);
-            // ignore error
-          }
-        }),
+      const { InvokeCommand, LambdaClient } = await import(
+        "@aws-sdk/client-lambda"
```
khuezy, line 25: 👍 nice, we should dynamically import the aws-sdk where possible.

packages/open-next/src/wrappers/aws-lambda-streaming.ts
```diff
@@ -6,14 +6,30 @@ import { StreamCreator } from "http/index.js";
...
+      resolve({ serverId, type: "warmer" } satisfies WarmerResponse);
+    }, event.delay);
+  });
+  return result;
+}
+
 const handler: WrapperHandler = async (handler, converter) =>
   awslambda.streamifyResponse(
     async (event: AwsLambdaEvent, responseStream): Promise<AwsLambdaReturn> => {
+      if ("type" in event) {
```
khuezy, line 27: Just a thought: This currently works, but it should have been designed as `event.type === 'warmer'` to be explicit.

packages/open-next/src/adapters/edge-adapter.ts
```diff
@@ -0,0 +1,73 @@
...
+      responseHeaders[key] = responseHeaders[key]
+        ? [...responseHeaders[key], value]
+        : [value];
+    } else {
+      responseHeaders[key] = value;
+    }
+  });
+  console.log("responseHeaders", responseHeaders);
+  const body = buffer.toString();
+  console.log("body", body);
```
khuezy, line 54: Remove logs?

packages/open-next/src/adapters/edge-adapter.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  return {
+    type: "core",
+    statusCode: response.status,
+    headers: responseHeaders,
+    body: body,
+    // Do we need to handle base64 encoded response?
+    isBase64Encoded: false,
```
khuezy, line 62: I think we should, better safe than sorry.

packages/open-next/src/adapters/warmer-function.ts
conico974: Yeah, the only one missing is in the image-optimization. The rest are already dynamically loaded

packages/open-next/src/wrappers/aws-lambda-streaming.ts
conico974: I think it should be `"type" in event && event.type === "warmer"`, otherwise typescript will complain

PR #329, opened 2023-12-05
Changes encoding on cache.body from utf8 to base64

dylanirion:
[Code generated icons](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons#generate-icons-using-code-js-ts-tsx) are cached as routes with a binary body. open-next currently merges the route meta and utf-8 encoded body into a single *.cache file, which can be problematic for binary png data. This PR simply changes the encoding on the cached route body from `utf8` to `base64` to preserve binary data.

Alternatively, we could preserve the utf-8 encoding on json routes, and only apply base64 when necessary according to `meta.headers.content-type`?

### DIFF
--- a/.changeset/wicked-comics-trade.md
+++ b/.changeset/wicked-comics-trade.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+Changes encoding on cache.body for binary data
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -226,7 +226,12 @@ export default class S3Cache {
           lastModified: LastModified?.getTime(),
           value: {
             kind: "ROUTE",
-            body: Buffer.from(cacheData.body ?? Buffer.alloc(0), "base64"),
+            body: Buffer.from(
+              cacheData.body ?? Buffer.alloc(0),
+              meta?.headers?.["content-type"]?.includes("application/json")
+                ? "utf8"
+                : "base64",
+            ),
             status: meta?.status,
             headers: meta?.headers,
           },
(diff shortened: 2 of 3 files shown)


khuezy:
Thanks!

khuezy (commented):
Did you get a chance to deploy the changes and verify things are working correctly?
My only concern is that we now default to `base64` and only set `utf8` under 1 specific condition (content type is `application/json`)

dylanirion:
> Did you get a chance to deploy the changes and verify things are working correctly? My only concern is that we now default to `base64` and only set `utf8` under 1 specific condition (content type is `application/json`)

icons and a dummy static route that returns json are working in my deployment. I hear you RE the greedy `base64` encoding. We could do the opposite and catch via `startsWith("image")` instead? I nearly did this but my thinking was it's cleaner to match just json over the countless other binary mime types that users might be generating. I considered `text/html` but figured that's what a page is for. I suppose matching `text/plain` might be necessary then too.

Would the better pattern be`utf8` as default and then specific cases for `base64`? Happy to re-commit!

khuezy:
Yes I think we should default as `utf8` and on do base64 for images. Sorry for the churn, but it's good that we have this discussion with regards to bandwidth... it does add up at scale.

khuezy (approved):
LGTM thanks!

conico974:
We should probably just use this function here https://github.com/sst/open-next/blob/ba6e176fe30c3654de24c34a0269cbf1ac4d8ada/packages/open-next/src/adapters/binary.ts#L59C27-L59C27
We already use this function in other places, we should have a single way to choose between `utf8` and `base64`

dylanirion:
> We should probably just use this function here https://github.com/sst/open-next/blob/ba6e176fe30c3654de24c34a0269cbf1ac4d8ada/packages/open-next/src/adapters/binary.ts#L59C27-L59C27 We already use this function in other places, we should have a single way to choose between `utf8` and `base64`

Hah! How'd I miss that? I'll drop this in with another commit

PR #330, opened 2023-12-07
Initial docs for V3

conico974:
Just some docs for open-next V3 release candidate

### DIFF
--- a/docs/pages/v3/config.mdx
+++ b/docs/pages/v3/config.mdx
@@ -23,15 +23,19 @@ const config = {
   },
   functions: { // here we define the functions that we want to deploy in a different server
     ssr: {
-      routes: ["app/api/isr/route", "app/api/sse/route", "app/api/revalidateTag/route"], // For app dir, you need to include route|page, no need to include layout or loading
-      patterns: ['api/*'], // patterns needs to be in a cloudfront compatible format, this will be used to generate the output
+      routes: [
+        "app/api/isr/route", "app/api/sse/route", "app/api/revalidateTag/route", // app dir Api routes
+        "app/route1/page", "app/route2/page", // app dir pages
+        "pages/route3" // page dir pages
+        ], // For app dir, you need to include route|page, no need to include layout or loading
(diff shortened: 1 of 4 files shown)


docs/pages/v3/config.mdx
```diff
@@ -0,0 +1,56 @@
...
+      queue: () => Promise.resolve({
+        send: async (message) => {
+          //Your custom code here
+        }
+      })
+    },
+  },
+  functions: { // here we define the functions that we want to deploy in a different server
+    ssr: {
+      routes: ["app/api/isr/route", "app/api/sse/route", "app/api/revalidateTag/route"], // For app dir, you need to include route|page, no need to include layout or loading
```
khuezy, line 26: "include route|page", is the `page` for the `pages-dir`? That comment might be confusing.

docs/pages/v3/config.mdx
```diff
@@ -0,0 +1,56 @@
...
+      patterns: ['api/*'], // patterns needs to be in a cloudfront compatible format, this will be used to generate the output
+      override: {
+        wrapper: "aws-lambda-streaming",
+      },
+      experimentalBundledNextServer: true // This enables the bundled next server which is faster and reduce the size of the server
```
khuezy, line 31: We might want to move this inside an `experimental` object if there's more than 1 in the future.

khuezy (approved):
LGTM, some minor comments.

docs/pages/v3/config.mdx
conico974: Good idea, but as of now there is only one `experimental`

docs/pages/v3/config.mdx
khuezy: ah ok, it might be a little clearer if there was an example of the page, eg:
```typescript
routes: [
   "app/api/isr/route", "app/api/sse/route", "app/api/revalidateTag/route", // API routes
   "app/route1/page", "app/route2/page", "app/route3/page"                    // Pages
]
```

conico974: I'll add this and merge

PR #336, opened 2023-12-18
fix: try to match errors, fall back to just adding raw key / value pare

JanStevens:
Hi,

Current logic for matching headers does not 100% match how nextjs deals with it. For example when you have a `Content-Security-Policy` header which contains urls like `https://google.com` then the matching will fail resulting in an Internal Server Error.

This PR adds a small bandage so non matched headers are just returned as-is ensuring no Internal Server Error is triggered.

Ideally we have same logic as Nextjs

### DIFF
--- a/.changeset/afraid-ducks-speak.md
+++ b/.changeset/afraid-ducks-speak.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+try to match errors, fallback to raw key/value pair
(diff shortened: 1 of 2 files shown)


khuezy (commented):
Does this break any functionality of your headers? We might want to port of Vercel's logic to handle the matching regex.

JanStevens:
@khuezy No I just had some basic security headers present which caused the parameter matching to trip up.

Link to the relevant discord discussion: https://discord.com/channels/983865673656705025/1186265613220446218

PR #353, opened 2024-01-30
fix(open-next): correctly set cache control for html pages

jadenv:
As the code reads in the comment right above the line I changed, NextServer will not respond with a cache-control header value for HTML pages.

Since the cache-control header value is never set upstream, open-next also never set its own cache-control header due to this truthy if check.

As HTML pages never got any cache-control value set, what should be some of the fastest pre-rendered pages actually become slower than SSR/ISR pages that do get a cache header set correctly. (In the case of `cdk-nextjs`, the value actually gets falls back to `no-cache`)

I tested this locally with a number of CloudFront events against a site with static HTML as well as other SSG/ISR page types and it helped a lot for static HTML pages. 🚀

### DIFF
--- a/.changeset/wet-brooms-cheer.md
+++ b/.changeset/wet-brooms-cheer.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+fix(open-next): correctly set cache control for html pages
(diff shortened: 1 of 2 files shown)


conico974 (approved):
LGTM

PR #364, opened 2024-02-16
Fix cannot write default config file b/c folder not created

fwang:
Happens when the `.open-next` folder does not exist.

Also includes a few trivial changes:
- Changed windows message to warning, since user can still continue
- Printing out a larger block of warning message and removed the 10s wait. Thinking being if it works for a windows user (miraculously), sucks having to wait 10s on each deploy.
- removed a few props from BuildOptions, that exists in both `config` and `options` that are being passed around.


conico974 (approved):
LGTM

PR #369, opened 2024-02-26
fix: prevent duplication of location header

### LINKED ISSUE #368: Duplicating location header
I'm using `next-intl`. Server function handles redirect from middleware, but there is an issue that it duplicates `location` header because of case mismatch.

```
"headers": {
    "location": "https://someurl.com/some-path",
    "Location": "https://someurl.com/some-path"
},
```

r34son:
Closes #368

### DIFF
--- a/packages/open-next/src/adapters/routing/middleware.ts
+++ b/packages/open-next/src/adapters/routing/middleware.ts
@@ -123,18 +123,13 @@ export async function handleMiddleware(
   // If the middleware returned a Redirect, we set the `Location` header with
   // the redirected url and end the response.
   if (res.statusCode >= 300 && res.statusCode < 400) {
-    const location = result.response.headers
-      .get("location")
-      ?.replace("http://localhost:3000", `https://${req.headers.host}`);
+    resHeaders.location = resHeaders.location?.replace("http://localhost:3000", `https://${req.headers.host}`);
     // res.setHeader("Location", location);
     return {
       body: "",
       type: internalEvent.type,
       statusCode: res.statusCode,
-      headers: {
-        ...resHeaders,
-        Location: location ?? "",
-      },
+      headers: resHeaders,
       isBase64Encoded: false,
     };
   }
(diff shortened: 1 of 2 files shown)


packages/open-next/src/adapters/routing/middleware.ts
```diff
@@ -123,18 +123,13 @@ export async function handleMiddleware(
   // If the middleware returned a Redirect, we set the `Location` header with
   // the redirected url and end the response.
   if (res.statusCode >= 300 && res.statusCode < 400) {
-    const location = result.response.headers
```
khuezy, line 126: iirc, we had to handle this case specifically for middleware `Redirects`. Can you verify that it works with this change?

r34son: Yeah, i`ve tested it by patching 2.3.6 version. Maybe I need to cover this case by some tests? Do you have any contributing docs?

khuezy: I believe we have this covered (middleware redirect test)[https://github.com/sst/open-next/blob/main/packages/tests-e2e/tests/appRouter/middleware.redirect.test.ts]

App: https://github.com/sst/open-next/blob/main/examples/app-router/middleware.ts#L8

packages/open-next/src/adapters/routing/middleware.ts
```diff
@@ -123,18 +123,13 @@ export async function handleMiddleware(
...
-      .get("location")
-      ?.replace("http://localhost:3000", `https://${req.headers.host}`);
+    resHeaders.location = resHeaders.location?.replace("http://localhost:3000", `https://${req.headers.host}`);
```
conico974, line 126: Not sure this will work on every case, what if you set the uppercase `Location` headers in the middleware ?

r34son: here i mentioned that it doesn't matter uppercase or lowercase, key will be lowercase always
https://github.com/sst/open-next/issues/368#issuecomment-1964869309

Issue #386, opened 2024-03-15
[Edge] Set-Cookie headers are concatenated


dg-lucasvieira:
#### Issue

When the Next.js App is deployed to AWS Lambda@Edge and there is more than one `set-cookie` header in the response, the CloudFront response concatenates the entries into one `set-cookie` item.

#### Example

Header: `set-cookie`: `test=1; Path=/; HttpOnly; Secure; SameSite=None, test=2; Path=/; HttpOnly; Secure; SameSite=None`

and it should be:

`set-cookie`: `test=1; Path=/; HttpOnly; Secure; SameSite=None`
`set-cookie`: `test=2; Path=/; HttpOnly; Secure; SameSite=None`

khuezy:
Hi thanks for the report. Is this specific to L@E or does this happen everywhere?

lucasvieirasilva:
I've only tested in the L@E, but looking at OpenNext's code the other places seem to be right.

I've raise a PR for that https://github.com/sst/open-next/pull/387

conico974:
Fixed in open-next@2.3.8 by https://github.com/sst/open-next/pull/387

PR #387, opened 2024-03-15
fix(open-next): parse cookies when converting response to cloudfront

lucasvieirasilva:
This PR parse the `set-cookie` header when converting the response to the CloudFront format.

#### Current Behavior

`set-cookie: test=1; Path=/; HttpOnly; Secure; SameSite=None, test=2; Path=/; HttpOnly; Secure; SameSite=None`

#### Expected Behavior

`set-cookie: test=1; Path=/; HttpOnly; Secure; SameSite=None`
`set-cookie: test=2; Path=/; HttpOnly; Secure; SameSite=None`

#### Tests

I've published it to my own npm org and tested it in my application:

https://www.npmjs.com/package/@sstlv/open-next/v/2.3.7-beta.0

#### Fixes

[#386]

### DIFF
--- a/.changeset/short-eels-compare.md
+++ b/.changeset/short-eels-compare.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+fix(open-next): parse cookies when converting response to cloudfront
--- a/packages/open-next/src/adapters/event-mapper.ts
+++ b/packages/open-next/src/adapters/event-mapper.ts
@@ -218,6 +218,17 @@ function convertToCloudFrontRequestResult(
   Object.entries(result.headers)
     .filter(([key]) => key.toLowerCase() !== "content-length")
     .forEach(([key, value]) => {
+      if (key === "set-cookie") {
+        const cookies = parseCookies(value);
+        if (cookies) {
+          headers[key] = cookies.map((cookie) => ({
+            key,
+            value: cookie,
+          }));
+        }
+        return;
+      }
+
       headers[key] = [
         ...(headers[key] || []),
         ...(Array.isArray(value)
--- a/packages/tests-unit/tests/event-mapper.test.ts
+++ b/packages/tests-unit/tests/event-mapper.test.ts
@@ -1,0 +1,114 @@
+import { CloudFrontRequestResult } from "aws-lambda";
+
+import { convertTo } from "../../open-next/src/adapters/event-mapper";
+
+describe("convertTo", () => {
+  describe("CloudFront Result", () => {
+    it("Should parse the headers", () => {
+      const response = convertTo({
+        body: "",
+        headers: {
+          "content-type": "application/json",
+          test: "test",
+        },
+        isBase64Encoded: false,
+        statusCode: 200,
+        type: "cf",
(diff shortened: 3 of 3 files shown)


khuezy:
Thanks for the PR, can you also add an e2e test to cover this?
eg: https://github.com/sst/open-next/blob/main/examples/app-router/middleware.ts#L39
set the `set-cookie` and validate that multiple cookies are set on the client.

lucasvieirasilva:
> Thanks for the PR, can you also add an e2e test to cover this? eg: https://github.com/sst/open-next/blob/main/examples/app-router/middleware.ts#L39 set the `set-cookie` and validate that multiple cookies are set on the client.

Do the e2e tests simulate the CloudFront request/response?

If not, this block is already covering multiple cookies: https://github.com/sst/open-next/blob/main/examples/app-router/middleware.ts#L62-L66

khuezy:
Cool, I wonder if we could just add `edge: true` here: https://github.com/sst/open-next/blob/main/examples/sst/stacks/AppRouter.ts or if we want to duplicate the stack to cover both scenarios... probably the latter.

lucasvieirasilva:
No, in the project that I'm using OpenNext locally always works, even without this fix, only when we deploy the issue start happening.

khuezy:
correct, I'm saying we should add coverage for the deployed e2e case with Lambda@Edge configured.

khuezy:
I'm thinking we can just duplicate this https://github.com/sst/open-next/blob/main/examples/sst/stacks/AppRouter.ts and name it `AppRouterEdge` and add `edge: true`.

Then add the new stack here: https://github.com/sst/open-next/blob/main/examples/sst/sst.config.ts

Then update the e2e configuration: https://github.com/sst/open-next/blob/main/.github/workflows/e2e.yml#L154-L159

Finally add some e2e that targets the edge app, eg a copy of this: https://github.com/sst/open-next/blob/main/packages/tests-e2e/tests/appRouter/middleware.cookies.test.ts but for the `appRouterEdge` version of the coverage.

conico974:
I'm not sure creating another e2e app to test only `edge: true` make that much sense
The only difference between `edge: true` and `edge: false` from an open-next standpoint is in the event mapper. 

The thing that i think we should do is to start incorporating unit test for this kind of things.
If we properly setup those test, most things should be covered, and we can add the missing test afterwards.
There is already a `tests-unit` package, we just need to create the tests and add them to the github actions.

And with V3 and splitting we won't have to move a whole app to the edge, we could only move a small part there for e2e.

khuezy:
Fair point, since V3 is "around the corner", a unit test would suffice. My only worry is that a unit test doesn't truly cover the actual deployed services so things can go through the cracks.

lucasvieirasilva:
@khuezy I've added some unit tests for the CloudFront event mapper.

khuezy (approved):
LGTM thanks!

lucasvieirasilva:
@khuezy can we have this PR merged and released?

conico974:
I can test, merge and release tomorrow

PR #397, opened 2024-03-31
Fix incorrect 200 with wrong buildId

conico974:
When requesting data for page dir with an incorrect `BUILD_ID`, we should return 404 with `{}` as the body.

This could cause some 500 on new deployment for page router, since running client could request data with the old `BUILD_ID`

Links to the discord thread : https://discord.com/channels/983865673656705025/1217895314741006387


khuezy (approved):
LGTM

PR #398, opened 2024-03-31
Feat add a static etag for Image

conico974:
At the moment, image from `_next/image` are only cached in the cdn.
We don't generate any `ETag` either for the image so we never get 304.
Every stale request actually needs to be recomputed again.

In order to improve perf and reduce cost, this PR introduces an env variable `OPENNEXT_STATIC_ETAG` that we can set on the image opt function.
When you set this env variable, we create an `ETag` that depends on some parameter that we know before performing the image opt: ( href, width, quality, buildId )
Stale request will then return a 304 for request where `if-none-match` header is equal to the computed `ETag`

**One important things to note**, when using this env variable, it means that all the images should be considered immutable (The source image should not change) and the only way to change that is to change the `BUILD_ID` (on new deploy)

Not sure if this should be enabled by default or not.

### DIFF
--- a/.changeset/hot-toys-push.md
+++ b/.changeset/hot-toys-push.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+Feat add a static etag for Image Optimization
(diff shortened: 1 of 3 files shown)


khuezy (approved):
LGTM

PR #400, opened 2024-04-01
Docs for V3

conico974:
This PR updates the docs for V3.
It should include all the features of V3 and keep all the docs for V2 under `/v2/...`

There are still some things that will need some work.
- The FAQ page is unchanged, not even sure if we should keep it
- Need to write some docs about the `edge` runtime in V3
- ISR/SSG is such a big topic, not sure if the way i've done it is the correct one
- I've used the term backend for all the functions (this could be a lambda, docker...), not sure if that makes sense

Also some things might seem obvious to me, not for others.

Should we merge this now ? Or should we wait for V3 to be officially out ?


khuezy (approved):
LGTM.

I wasn't 100% thorough in the review, too many words and if I missed anything it's not too big of a deal.

Releasing the v3 docs now is fine, we might find some typos after seeing it live.

conico974:
Ok i'll merge then

PR #409, opened 2024-05-09
Fix next version check

### LINKED ISSUE #406: Image optimization fails due to version mismatch
I'm using OpenNext (v2.3.7) through the SST `NextJsSite` construct. 

I recently had an issue where images loaded through the `<Image>` tag failed at the optimization step. After digging into the issue, I learned that Next changed the optimizer call signature in v14.1.1, hence #374 and #377. I was running v14.2.2.

Digging in further, I learned that the OpenNext image optimization plugin was not applied during my build step. This is because the plugin is [gated behind a version check for v14.1.1](https://github.com/chungweileong94/open-next/blob/98db0fc20f1515d9d10bfc598359ecff4c810ee6/packages/open-next/src/build.ts#L328). BUT when a Next version isn't specified in the initial options, [OpenNext attempts to extract it from package.json](https://github.com/sst/open-next/blob/826a37ede3cf2081db1aa7a93b07eeda9f46d0d2/packages/open-next/src/build.ts#L191).

However, the version specified in `package.json` is not necessarily the version that's installed. In my case, I had `"next": "^14.0.4"` in my `package.json`, but the actual installed version was v14.2.2. This resulted in a logic mismatch between versions, and the image optimizer function simply returned `Cannot read properties of undefined (reading '0')` every time it was invoked.

It seems like an appropriate solution would be to instead obtain the installed version number from `package-lock.json` or the module itself, but I defer to the original authors.

ETA: I was able to fix my original issue by bumping the `next` version in `package.json` to `^14.2.2`.

conico974:
Fix #406 
Instead of checking `package.json` for next version, we do a `require.resolve` in the appPath directory to find the `package.json` of next. Then we require this package and check for the version there.
This should work every time

### DIFF
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -25,19 +25,19 @@ export function normalizeOptions(config: OpenNextConfig, root: string) {
   );
   const outputDir = path.join(buildOutputPath, ".open-next");
 
-  let nextPackageJsonPath: string;
+  let appPackageJsonPath: string;
   if (config.packageJsonPath) {
     const _pkgPath = path.join(process.cwd(), config.packageJsonPath);
-    nextPackageJsonPath = _pkgPath.endsWith("package.json")
+    appPackageJsonPath = _pkgPath.endsWith("package.json")
       ? _pkgPath
       : path.join(_pkgPath, "./package.json");
   } else {
-    nextPackageJsonPath = findNextPackageJsonPath(appPath, root);
+    appPackageJsonPath = findNextPackageJsonPath(appPath, root);
   }
   return {
     openNextVersion: getOpenNextVersion(),
     nextVersion: getNextVersion(appPath),
-    nextPackageJsonPath,
+    appPackageJsonPath,
     appPath,
     appBuildOutputPath: buildOutputPath,
     appPublicPath: path.join(appPath, "public"),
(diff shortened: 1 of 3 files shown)


packages/open-next/src/build/helper.ts
```diff
@@ -209,9 +209,12 @@ export function getOpenNextVersion(): string {
   return require(path.join(__dirname, "../../package.json")).version;
 }
 
-export function getNextVersion(nextPackageJsonPath: string): string {
-  const version = require(nextPackageJsonPath)?.dependencies?.next;
-  // require('next/package.json').version
+export function getNextVersion(appPath: string): string {
+  // We cannot just require("next/package.json") because it could be executed in a different directory
+  const nextPackageJsonPath = require.resolve("next/package.json", {
```
Seanw265, line 214: There are two variables in this file named `nextPackageJsonPath` with different semantic meanings. This one appears to be the package.json path for the `next` package itself, while the other (L28) seems to represent the package.json path for the Nextjs app that OpenNext is running against.

There is also a function called `findNextPackageJsonPath` that seems to obtain the latter.

To someone unfamiliar with the codebase, this is a bit confusing. Might I suggest renaming these to clarify which is which? Maybe this instance could be `nextDependencyPackageJsonPath` or something along those lines? 

Happy to be corrected if there is anything I've missed.

conico974: Yeah you're right, they should be different. I think it would make more sense to rename the other one `packageJsonPath` or maybe `appPackageJsonPath`

PR #410, opened 2024-05-09
Handle partial failure in ISR revalidation

conico974:
This PR improve ISR revalidation. 

Right now if a revalidation of a page fails for any reason, the page will have to wait for the 5 min deduplication window before attempting any revalidation again. And in a batch if the first one fails with an exception, all the other route in the batch will also have to wait for this 5 min dedup window.

This PR handles partial failures, and if properly set in IAC, the failed route will automatically retry every 30 seconds.

This also fix an issue with HEAD request not having the fix applied for all the ISR headers in non streaming and so revalidation ISR request were cached in the CDN, which could also cause the 5 min dedup window issue


khuezy (approved):
LGTM

PR #411, opened 2024-05-09
Fix next rewrites

### LINKED ISSUE #263: Rewrites in next.config.js doesnt work
After updating Next.js to 13.5.3 to handle a bad error in production (500, attemp to render client in the server). The following stopped working: 
  redirects: () => {
    const SIGNIN_ROUTE = `/login`;
    const DEFAULT_ROUTE = `/dashboard`;
    return [
      {
        source: `/`,
        destination: DEFAULT_ROUTE,
        permanent: true,
        has: [{type: `cookie`, key: `next-auth.session-token`}],
      },
      {
        source: `/`,
        destination: DEFAULT_ROUTE,
        permanent: true,
        has: [{type: `cookie`, key: `__Secure-next-auth.session-token`}],
      },
      {
        source: `/`,
        destination: SIGNIN_ROUTE,
        permanent: true,
        missing: [{type: `cookie`, key: `__Secure-next-auth.session-token`}],
      },
      {
        source: `/`,
        destination: SIGNIN_ROUTE,
        permanent: true,
        missing: [{type: `cookie`, key: `next-auth.session-token`}],
      },
    ];

I am in a damn loop hole, for few weeks already. 
My original issue was Next trying to generated protected pages during ISG, and then caching a 500 result. 
Then I updated next.js and sst to the latest stable versions (13.5.4, sst 2.26.10). 
My folder setup is as following:
src
├── app
│   ├── @authenticated
│   ├── @login
│   ├── _components
│   ├── _hooks
│   ├── _lib
│   ├── api
│   ├── layout.tsx
│   ├── loading.tsx
│   ├── page.tsx
│   └── providers.tsx
└── middleware.ts

Then my root layout conditionally renders login/auth. 
Everything works as expected except the root dir "/" which is an empty route. 
NextAuth should redirect to '/login' and next itelsf to 'dashboard'. 
It works locally without setting next.config.ts. 
I tried adding the same login in the middleware but it doesnt seem like the middleware is even running (not seeing any logs form there). 
It gets a successful response from '/api/auth/session' but the redirect doesnt happen :/ Any ideas?

conico974:
Fix #263 
This PR fix the uses of named regex in next.config rewrites.
This one https://github.com/sst/open-next/issues/263#issuecomment-1832240508 works just fine now

### DIFF
--- a/packages/open-next/src/core/routing/matcher.ts
+++ b/packages/open-next/src/core/routing/matcher.ts
@@ -92,8 +92,7 @@ const computeParamHas =
     query: Record<string, string | string[]>,
   ) =>
   (has: RouteHas): object => {
-    let params = {};
-    if (!has.value) return params;
+    if (!has.value) return {};
     const matcher = new RegExp(`^${has.value}$`);
     const fromSource = (value: string) => {
       const matches = value.match(matcher);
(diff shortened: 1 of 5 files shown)


packages/open-next/src/core/routing/matcher.ts
```diff
@@ -71,6 +78,45 @@ function checkHas(
...
+    let params = {};
+    if (!has.value) return params;
+    const matcher = new RegExp(`^${has.value}$`);
+    const fromSource = (value: string) => {
+      const matches = value.match(matcher);
+      return matches?.groups ?? {};
+    };
+    switch (has.type) {
+      case "header":
+        params = fromSource(headers[has.key.toLowerCase()] ?? "");
```
khuezy, line 104: Just a minor nit to save a few lines of code: just `return fromSource(...`

khuezy (approved):
LGTM, there were a couple questions in your comment but I'm a bit out of the loop and preoccupied at the moment - sorry.

PR #418, opened 2024-05-20
Fix polyfill for crypto in the middleware 

conico974:
Some libs depends on crypto function or class that are available globally in the edge runtime, but are not present by default in node, this should fix it


chungweileong94:
LGTM

PR #419, opened 2024-05-22
Add support for 'deno' server functions

littledivy:
This patch adds support for running Open-next server functions with the [Deno lambda](https://github.com/denoland/deno-lambda) runtime layer.

Discussed briefly with @conico974 on Discord, adding `runtime: "deno"` made most sense.

## Changes

- Compile a ES module version of `cache.ts` when runtime is set to "deno"
- Add a `deno.json` and rename function entrypoint as `index.ts` (this will be redundant with Deno 2)
- Import node globals in the entrypoint

## Example

Here's how a deployment config would look like:
```typescript
// open-next.config.ts
export default {
  default: {
    runtime: "deno"
  }
}
```

```typescript
// sst.config.ts
export default $config({
  app(input) {
    return {
      name: "my-app",
      removal: input?.stage === "production" ? "retain" : "remove",
      home: "aws",
    };
  },
  async run() {
    new sst.aws.Nextjs("MyWeb", {
      transform: {
        server: {
          /* Amazon Linux 2023 */
          runtime: "provided.al2023",
          /* Deno lambda layer */
          layers: ["arn:aws:lambda:ap-south-1:381491846142:layer:deno:2"]
        }
      }
    });
  }
});
```

```
sst deploy
```

Next.js page: https://dvwb0nol21fnx.cloudfront.net 
API route: https://dvwb0nol21fnx.cloudfront.net/api/hello

### DIFF
--- a/packages/open-next/src/build.ts
+++ b/packages/open-next/src/build.ts
@@ -677,15 +677,16 @@ async function createCacheAssets(monorepoRoot: string) {
 /* Server Helper Functions */
 /***************************/
 
-function compileCache() {
-  const outfile = path.join(options.outputDir, ".build", "cache.cjs");
+export function compileCache(format: "cjs" | "esm" = "cjs") {
+  const ext = format === "cjs" ? "cjs" : "mjs";
+  const outfile = path.join(options.outputDir, ".build", `cache.${ext}`);
   esbuildSync(
     {
       external: ["next", "styled-jsx", "react", "@aws-sdk/*"],
       entryPoints: [path.join(__dirname, "adapters", "cache.js")],
       outfile,
       target: ["node18"],
-      format: "cjs",
+      format,
       banner: {
         js: [
           `globalThis.disableIncrementalCache = ${
--- a/packages/open-next/src/build/createServerBundle.ts
+++ b/packages/open-next/src/build/createServerBundle.ts
@@ -10,6 +10,7 @@ import {
 } from "types/open-next";
 import url from "url";
 
+import { compileCache } from "../build.js";
 import logger from "../logger.js";
 import { minifyAll } from "../minimize-js.js";
 import { openNextReplacementPlugin } from "../plugins/replacement.js";
@@ -36,6 +37,14 @@ export async function createServerBundle(
   // Get all functions to build
   const defaultFn = config.default;
   const functions = Object.entries(config.functions ?? {});
+
+  // Recompile cache.ts as ESM if any function is using Deno runtime
+  if (
+    defaultFn.runtime === "deno" ||
+    functions.some(([, fn]) => fn.runtime === "deno")
+  ) {
+    compileCache("esm");
+  }
 
   const promises = functions.map(async ([name, fnOptions]) => {
     const routes = fnOptions.routes;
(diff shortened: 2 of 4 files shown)


packages/open-next/src/build/createServerBundle.ts
```diff
@@ -37,6 +38,13 @@ export async function createServerBundle(
   const defaultFn = config.default;
   const functions = Object.entries(config.functions ?? {});
 
+  if (
```
conico974, line 41: You should add a comment to explain why it's necessary to call `compileCache` again

packages/open-next/src/build/createServerBundle.ts
```diff
@@ -227,14 +240,18 @@ async function generateBundle(
...
   await esbuildAsync(
     {
       entryPoints: [path.join(__dirname, "../adapters", "server-adapter.js")],
       external: ["next", "./middleware.mjs", "./next-server.runtime.prod.js"],
-      outfile: path.join(outputPath, packagePath, "index.mjs"),
+      outfile: path.join(outputPath, packagePath, `index.${outfileExt}`),
       banner: {
         js: [
           `globalThis.monorepoPackagePath = "${packagePath}";`,
+          "import process from 'node:process';",
```
conico974, line 253: Do we want to add these 2 lines outside of deno ? Not sure if this could cause some trouble

packages/open-next/src/build/createServerBundle.ts
```diff
@@ -280,6 +297,17 @@ function shouldGenerateDockerfile(options: FunctionOptions) {
   return options.override?.generateDockerfile ?? false;
 }
 
+function addDenoJson(outputPath: string, packagePath: string) {
```
conico974, line 300: You said in your PR that it might not be necessary with deno 2, we should add a comment about this in the code as well

packages/open-next/src/build.ts
```diff
@@ -677,15 +677,16 @@ async function createCacheAssets(monorepoRoot: string) {
 /* Server Helper Functions */
 /***************************/
 
-function compileCache() {
-  const outfile = path.join(options.outputDir, ".build", "cache.cjs");
+export function compileCache(format: "cjs" | "esm" = "cjs") {
+  const ext = format === "cjs" ? "cjs" : "mjs";
```
conico974, line 681: Is it really necessary to create 2 variables here ?

conico974 (commented):
Do we want to allow the deno runtime for the other functions ? (except from middleware as this one requires `edge`)
Not sure if it's possible to make it work with the `imageOptimization` function since it requires sharp

packages/open-next/src/build.ts
littledivy: esbuild accepts `format: "esm"` which isn't a recognizable extension so we need this format -> ext mapping here.

packages/open-next/src/build/createServerBundle.ts
littledivy: It's fine, should be exactly same as the globals in Node.

littledivy:
> Do we want to allow the deno runtime for the other functions ? (except from middleware as this one requires edge)
Not sure if it's possible to make it work with the imageOptimization function since it requires sharp

It should work with all of them (including `imageOptimization`) since Deno supports sharp via Node-API. `imageOptimization` options does not have the runtime option yet, it takes `DefaultFunctionOptions`. I can test it out and probably make a follow-up PR adding `runtime` if its runnable via Deno. Does that sound good?

conico974:
> I can test it out and probably make a follow-up PR adding runtime if its runnable via Deno. Does that sound good?

 Yeah that's fine, i'll merge this and a bunch of other PR tomorrow in 3.0.2

conico974 (approved):

conico974:
@littledivy I assume you did not test with the `node` wrapper, only with the default `aws-lambda`.
It should work out of the box, but the default generated dockerfile would not work

littledivy:
Ah yes that makes sense. It should actually be trivial to add generated dockerfile support for Deno too:

```dockerfile
FROM deno-lambda:latest

COPY ./.open-next/server-functions/default/. .

# Round down timestamps to seconds (workaround for https://github.com/denoland/deno/issues/23904)
RUN find node_modules -type f -exec touch -t $(date +%Y%m%d%H%M.00) {} \;

# Run index.ts to warmup the V8 code cache
RUN timeout 15s deno run -Ldebug -A index.ts \
  && timeout 15s deno run -Ldebug -A cache.mjs

CMD ["index.handler"]
```

Preliminary startup time results show its on-par with the `node` lambda.

sommeeeer:
@littledivy whats the status now on running deno as a lambda layer with deno2? we probably need to update our code to address the new changes. could you point me in the right direction?

PR #424, opened 2024-05-24
fix: look for required-server-files.json in outputPath 

mkupko:
If build output path is different from the path to the root of the Next.js app's source code required-server-files.json is not found which throws a build error.

### DIFF
--- a/.changeset/early-olives-kick.md
+++ b/.changeset/early-olives-kick.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+fix: look for required-server-files.json in outputPath 


conico974 (changes requested):
Nice catch.
Since the appPath is not used anymore here, could you refactor the `generateOutput` function ?

MaksymKupko:
> Nice catch. Since the appPath is not used anymore here, could you refactor the `generateOutput` function ?

@conico974 done

packages/open-next/src/build.ts
```diff
@@ -95,7 +95,7 @@ export async function build(
   await createRevalidationBundle(config);
   await createImageOptimizationBundle(config);
   await createWarmerBundle(config);
-  await generateOutput(options.appPath, options.appBuildOutputPath, config);
+  await generateOutput(options.appBuildOutputPath, config);
```
khuezy, line 98: Thanks, LGTM.
As a side note for a new Issue ticket for potential build performance:
Can we wrap all the `create...Bundle` functions on line 94-97 in a `Promise.all` to run the bundling in parallel?
I believe they should all be independent and don't depend on each other.
I'm not sure if the time save would be noticeable though.

conico974: It would probably not make a very big difference, the only 2 that takes a bit of time are the imageOptimization (because of sharp install) and the serverFunction themselves, but yeah that something that we could add.

conico974 (approved):
Thanks LGTM!

PR #426, opened 2024-05-30
fix 404 when no route match at all

conico974:
When a request does not match any route at all, instead of returning the 404 page it returns a 500.
I'm pretty sure it does not happen on every version of next as well, something new that they added at one point.
It's an error in the next request handler itself (it tries to pipe to undefined), they must have moved this logic in the server router.

I need to add some e2e test for this as well.

### DIFF
--- a/.changeset/tame-pets-juggle.md
+++ b/.changeset/tame-pets-juggle.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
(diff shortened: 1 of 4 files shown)


khuezy (approved):

eschaefer:
This might have introduced a regression. See https://github.com/sst/open-next/issues/476#issuecomment-2260827299

PR #432, opened 2024-06-08
Fix env file not being copied in V3

### LINKED ISSUE #428: Undefined environment variables after upgrade to V3
Hi,

I have just noticed that after the upgrade to V3, environment variables without the `NEXT_PUBLIC` prefix are showing as `undefined`. The environment variables are used in server components, and therefore do not need the prefix. Downgrading to V2 resolved the issue.

Thank you.

conico974:
Fix #428

### DIFF
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -254,3 +254,20 @@ export function copyOpenNextConfig(
     path.join(outputPath, "open-next.config.mjs"),
   );
 }
+
+export function copyEnvFile(
+  appPath: string,
+  packagePath: string,
+  outputPath: string,
+) {
+  const baseAppPath = path.join(appPath, ".next/standalone", packagePath);
+  const baseOutputPath = path.join(outputPath, packagePath);
+  const envPath = path.join(baseAppPath, ".env");
+  if (fs.existsSync(envPath)) {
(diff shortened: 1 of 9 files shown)


packages/open-next/src/build/helper.ts
```diff
@@ -254,3 +254,20 @@ export function copyOpenNextConfig(
...
+  packagePath: string,
+  outputPath: string,
+) {
+  const baseAppPath = path.join(appPath, ".next/standalone", packagePath);
+  const baseOutputPath = path.join(outputPath, packagePath);
+  const envPath = path.join(baseAppPath, ".env");
+  if (fs.existsSync(envPath)) {
+    fs.copyFileSync(envPath, path.join(baseOutputPath, ".env"));
+  }
+  const envProdPath = path.join(baseAppPath, ".env.production");
```
khuezy, line 269: would it makes sense to support `.env.staging` or `.env.<arbitary-staging-env>`?

khuezy (approved):
LGTM, just 1 question.

packages/open-next/src/build/helper.ts
conico974: I don't think we should, next don't copy `.env.local` , `.env.development` or `.env.staging` into the standalone output.
Maybe there is some other, but i think we cover much cases here

PR #438, opened 2024-06-09
add new common issue: access denied errors on routes during page refresh and direct url access

sommeeeer:


conico974 (approved):
Thanks LGTM

PR #441, opened 2024-06-11
Fix for incorrect filter logic when copying traced files

vladiulianbogdan:
This fixes a bug that I encountered while deploying [this](https://github.com/andreia-oca/chatbot-ollama) application.

The problem is that the code uses a file from the `highlight.js` module called `csharp.js`. When I deploy the application built by open-next, I encounter the following error at runtime: `Cannot find module 'highlight.js/lib/languages/csharp'`.

I found that the reason for this error is an early return from `copyTracedFiles.ts`, which is falsely skipping the file `.next/standalone/node_modules/highlight.js/lib/languages/csharp.js`.

The solution that I propose is to skip if the path contains "node_modules/sharp" and "node_modules/canius-lite".

### DIFF
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -156,11 +156,10 @@ See the docs for more information on how to bundle edge runtime functions.
   //Actually copy the files
   filesToCopy.forEach((to, from) => {
     if (
-      from.includes("node_modules") &&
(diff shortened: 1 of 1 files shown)


conico974 (approved):
Thanks.
I think this is fine for now and will fix most issues, but we need to find a better way to do this. If someone needs `caniuse-lite` for example, we should have a way to include it in this case

PR #442, opened 2024-06-12
docs: fix links in index.mdx and refactor simple example page

sommeeeer:


docs/pages/config/simple_example.mdx
```diff
@@ -48,45 +83,45 @@ This will generate 2 server functions, the default one and the edge one. The edg
 Edge runtime function have less cold start time, but you can only deploy one route per function. They also does not have the middleware bundled in the function, so you need to use external middleware if you need it in front of the edge function.
```
khuezy, line 83: While you're here, can you also update this:

```suggestion
Edge runtime function have less cold start time, but you can only deploy one route per function. They also do not have the middleware bundled in the function, so you need to use external middleware if you need it in front of the edge function.
```

sommeeeer: for sure. adding another commit

sommeeeer:
@khuezy do we want double or single quotes? :) as it is live now theres a mix

khuezy:
Thanks for the PR, I don't know if we have a lint rule for the quotes. But personally, I use single quote for ts and double quotes for HTML.

sommeeeer:
> Thanks for the PR, I don't know if we have a lint rule for the quotes. But personally, I use single quote for ts and double quotes for HTML.

same here :)

docs/pages/config/simple_example.mdx
```diff
@@ -30,13 +49,59 @@ const config = {
...
+  },
+} satisfies OpenNextConfig;
+
+export default config;
+```
+
+## Running in Lambda@Edge
+
+<Callout type="info" emoji="ℹ️">
+  External middleware is not supported in Lambda@Edge.
```
conico974, line 92: That's not correct actually, in sst you cannot use external middleware with another function deployed in lambda@edge, because it would mean 2 function deployed at the same place.
External middleware is deployed in lambda@edge by default.
We can just remove this, it's more a sst stuff than an OpenNext one

conico974 (changes requested):
Apart from the comment i added, it's good.
Once updated i'll merge

conico974 (approved):
Thanks LGTM!

PR #444, opened 2024-06-15
Fix some cache issue

conico974:
Latest releases introduced multiple *-lite that replaced aws-sdk with aws4fetch. This introduced some bug and uncover some underlying issues with the incremental cache itself.

This PR fixes : 
- S3 cache NoSuchKey error will now always be in debug
- For some reasons, fetch (even with `next:internal` and an authorization header) sometimes will still uses the fetch cache for aws4fetch call. There is now a `globalThis.internalFetch` that we can use that is set at start to normal fetch.
- A weird `SocketError: other side closed` can happen when not reading the body ( which was the case for some aws call )

And the BIG one: 

Cache set to S3 and sending to SQS were not properly awaited. Next itself doesn't seem to wait for cache set to be done before returning the response. 
This has been undetected because those call can achieve before the lambda returns (but no guarantee of that) and because subsequent call to the same lambda instance resumed the execution of those call. The sdk was able to handle restarting the request from scratch, or maybe failing silently in some cases, but aws4fetch wasn't able to handle this.
This could also make some ISR call fails for up to 5min, for example if a revalidation request succeed, but the S3 PutObject failed the queue would not accept anymore request for this failed one for the 5 min deduplication window.

To fix this we used some detached promise along with AsyncLocalStorage to await everything we need before returning the response. 
This does not cause trouble for streaming return time ( at least in my account ) because it does not touch the stream, and it seems that they wait for the end of the handler function to end the lambda. Of course given aws records with streaming this could change, but there is a way around by awaiting those detached promises inside the `_flush` method of the `OpenNextNodeResponse`.
This also means that we might have a way to handle `next/after` natively in lambda

### DIFF
--- a/.changeset/small-rivers-taste.md
+++ b/.changeset/small-rivers-taste.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+Fix some cache issue
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -1,3 +1,5 @@
+import { DetachedPromise } from "utils/promise.js";
+
 import { IncrementalCache } from "../cache/incremental/types.js";
 import { TagCache } from "../cache/tag/types.js";
 import { isBinaryContentType } from "./binary.js";
@@ -167,7 +169,7 @@ export default class S3Cache {
         // If some tags are stale we need to force revalidation
         return null;
       }
-      const requestId = globalThis.__als.getStore() ?? "";
+      const requestId = globalThis.__als.getStore()?.requestId ?? "";
       globalThis.lastModified[requestId] = _lastModified;
       if (cacheData?.type === "route") {
         return {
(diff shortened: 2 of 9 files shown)


sommeeeer:
good job

khuezy (approved):
LGTM! Nice work 👍 

The `next/after` feature would be very useful for a lot of people.

khuezy:
Question: Since it now properly awaits everything before returning the response, is that a significant amount that affects the duration enough for people to care/notice?

conico974:
> The `next/after` feature would be very useful for a lot of people.

The only thing with that is that it can only properly work with streaming.  Hopefully aws can figure out all the streaming issues before Next 15 is out.

> Since it now properly awaits everything before returning the response, is that a significant amount that affects the duration enough for people to care/notice?

Not really most people won't even notice a difference, the only things we await for now are the cache set (only for ISR/SSG) and the message sending to the queue. And those mostly happens not visible to the user with RefreshHit from cloudfront or during ISR calls. 
The only cases where this could be noticeable ( but still very minimal, like 10s of ms ) would be for page not on the CDN (first request or new page).

I'll run a few more test, but if everything is fine, i'll probably make a release for this early next week

PR #446, opened 2024-06-18
Fix: await cache set

conico974:
Don't know how this one could have gone undetected, but set is not properly awaited.
I'll test this and merge this tomorrow, probably do a release as well. 
Maybe the latest PR was not even entirely necessary

### DIFF
--- a/.changeset/eight-parrots-peel.md
+++ b/.changeset/eight-parrots-peel.md
@@ -1,0 +1,11 @@
+---
+"open-next-docs": patch
+"open-next-benchmark": patch
+"app-pages-router": patch
+"app-router": patch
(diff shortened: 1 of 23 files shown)


khuezy (approved):
Good catch. I wonder if there's a lint that trigger Github to warn like it's warning the missing else clause below. At minimum, it would bring our attention to subtle things like this.

conico974:
> Good catch. I wonder if there's a lint that trigger Github to warn like it's warning the missing else clause below. At minimum, it would bring our attention to subtle things like this.

Yeah there is a bunch of them, i thought they was already included to be honest. I'll add them to the eslint config and fix any error that we found here

PR #448, opened 2024-06-21
fix: normalize the rawPath without locale before attempting to asserting the 404 page

JanStevens:


khuezy (commented):
Thanks! We might want to add i18n to the e2e app to better test these scenarios.

JanStevens:
> Thanks! We might want to add i18n to the e2e app to better test these scenarios.

I've updated the examples/pages-routes project to include some i18n config. I looked at the current e2e and they would already cover this bug since they assert the data json calls are resolving.

Would that be sufficient enough?

khuezy:
Thanks for adding the configs. I think it'll be useful to test the actual internationalization of the e2e apps to, that is, to add `[lang]` folder, eg: https://github.com/vercel/next.js/tree/canary/examples/app-dir-i18n-routing/app/%5Blang%5D

JanStevens:
> Thanks for adding the configs. I think it'll be useful to test the actual internationalization of the e2e apps to, that is, to add `[lang]` folder, eg: https://github.com/vercel/next.js/tree/canary/examples/app-dir-i18n-routing/app/%5Blang%5D

Hmm, looking at the example I'm not sure it would work seeing that the middleware takes care of all the i18n stuff with the app router. This means that I wouldn't know the i18n configuration to start with so my new "code" would never be executed 🤔 

TBH not using the app router at all atm, we stick with good old Pages 🤷

conico974:
It shouldn't be necessary to change anything for app dir, the middleware rewrites should already be applied at this point

packages/open-next/src/core/routingHandler.ts
```diff
@@ -22,12 +22,17 @@ export interface MiddlewareOutputEvent {
   origin: Origin | false;
 }
 
+// Add the locale prefix to the regex so we correctly match the rawPath
+const optionalLocalePrefixRegex = !!RoutesManifest.locales.length
+  ? `^/(?:${RoutesManifest.locales.map((locale) => `${locale}/?`).join("|")})?`
+  : null;
```
conico974, line 28: This make `pnpm run build` fail for me, in case people don't have i18n it should be the same so `^/`

packages/open-next/src/core/routingHandler.ts
```diff
@@ -22,12 +22,17 @@ export interface MiddlewareOutputEvent {
   origin: Origin | false;
 }
 
+// Add the locale prefix to the regex so we correctly match the rawPath
+const optionalLocalePrefixRegex = !!RoutesManifest.locales.length
+  ? `^/(?:${RoutesManifest.locales.map((locale) => `${locale}/?`).join("|")})?`
```
conico974, line 27: This cause an issue with 3 e2e test because of this, it should also include the default case (i.e. with no locales prefix)
image
For the next data test it is because of this https://github.com/sst/open-next/blob/9d632d1bd5c3a8d7e7bbd13ae9daaca68da9284e/packages/tests-e2e/tests/pagesRouter/data.test.ts#L13 It ends now with `en.json`

conico974 (changes requested):
There is a few issues with it. 
Ideally we should also handle the `locale:true|false` options in Next config rewrite/redirect

packages/open-next/src/core/routingHandler.ts
JanStevens: What would be the best approach here? Create a new pages app with i18n + additional e2e tests? 🤔 . It's expected that in this case it's `en.json` instead of `index.json` (but only if you have i18n settings).

It's a bit of a mess (on Next.js side)

conico974: In a perfect world, yeah we should do something like that, but honestly just testing with `en.json` is probably enough

conico974 (approved):

PR #449, opened 2024-06-21
Better support for cloudflare external middleware

conico974:
This PR fixes a few issues that happened when trying to use the external middleware with cloudflare.

#### TODO
- [x] Add a new resolver for the image loader compatible with cloudflare ( Probably accessing via url directly )

### DIFF
--- a/packages/open-next/src/converters/edge.ts
+++ b/packages/open-next/src/converters/edge.ts
@@ -68,7 +68,13 @@ const converter: Converter<
         },
       });
 
-      return fetch(req);
+      const cfCache = result.isISR ? { cacheEverything: true } : {};
+
+      return fetch(req, {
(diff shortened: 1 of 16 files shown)


packages/open-next/src/converters/edge.ts
```diff
@@ -68,7 +72,13 @@ const converter: Converter<
         },
       });
 
-      return fetch(req);
+      const cfCache = result.isISR ? { cacheEverything: true } : {};
+
+      return fetch(req, {
+        // This is a hack to make sure that the response is cached by Cloudflare
+        // @ts-expect-error - This is a Cloudflare specific option
```
khuezy, line 79: suggestion: reference the CF docs: `https://developers.cloudflare.com/workers/examples/cache-using-fetch/#caching-html-resources`

PR #451, opened 2024-06-24
Fix/redirect protocol

JanStevens:
Fixes https://github.com/sst/open-next/issues/450

### DIFF
--- a/examples/app-router/next.config.js
+++ b/examples/app-router/next.config.js
@@ -52,7 +52,7 @@ const nextConfig = {
         has: [{ type: "cookie", key: "from", value: "wrongvalue" }],
       },
       {
-        source: "/redirect",
+        source: "/next-config-redirect-without-locale-support",
         destination: "https://open-next.js.org/",
         permanent: false,
         basePath: false,
--- a/examples/app-router/next.config.js
+++ b/examples/app-router/next.config.js
@@ -51,6 +51,13 @@ const nextConfig = {
         permanent: true,
         has: [{ type: "cookie", key: "from", value: "wrongvalue" }],
       },
+      {
+        source: "/redirect",
+        destination: "https://open-next.js.org/",
+        permanent: false,
+        basePath: false,
+        locale: false,
+      },
     ];
   },
   headers() {
--- a/examples/pages-router/next.config.js
+++ b/examples/pages-router/next.config.js
@@ -22,9 +22,9 @@ const nextConfig = {
       ],
     },
   ],
-  redirects: [
+  redirects: () => [
     {
-      source: "/next-config-redirect-without-locale-support",
+      source: "/next-config-redirect-without-locale-support/",
       destination: "https://open-next.js.org/",
       permanent: false,
       basePath: false,
--- a/examples/pages-router/next.config.js
+++ b/examples/pages-router/next.config.js
@@ -24,7 +24,7 @@ const nextConfig = {
   ],
   redirects: [
     {
-      source: "/redirect",
+      source: "/next-config-redirect-without-locale-support",
       destination: "https://open-next.js.org/",
       permanent: false,
       basePath: false,
--- a/examples/pages-router/next.config.js
+++ b/examples/pages-router/next.config.js
@@ -22,6 +22,15 @@ const nextConfig = {
       ],
     },
   ],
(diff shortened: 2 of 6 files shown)


packages/open-next/src/core/routing/matcher.ts
```diff
@@ -216,7 +216,7 @@ export function handleRewrites<T extends RewriteDefinition>(
       rewrittenQuery = unescapeRegex(toDestinationQuery(params));
     }
     rewrittenUrl = isExternalRewrite
-      ? `${protocol}//${rewrittenHost}${rewrittenPath}`
```
khuezy, line 219: Thanks! Just to be overly cautious due to nextjs sometimes being inconsistent... are there scenarios where `protocol` doesn't actually include the `//`?

JanStevens: The `protocol` is returned by the utils function `getUrlParts` which always ensures that `https://` is returned

https://github.com/sst/open-next/blob/9d632d1bd5c3a8d7e7bbd13ae9daaca68da9284e/packages/open-next/src/core/routing/util.ts#L43-L66

khuezy (approved):
LGTM

packages/open-next/src/core/routing/matcher.ts
khuezy: Usually the protocol doesn't include the `//`, so this function should probably be updated.
https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol

JanStevens: Ha valid point indeed! I'll adapt

conico974:
We should add an e2e test for this as well, we could do an external redirect to the docs website for example

packages/open-next/src/core/routing/util.ts
```diff
@@ -52,13 +52,13 @@ export function getUrlParts(url: string, isExternal: boolean) {
     };
   }
 
-  const regex = /^(https?:\/\/)?([^\/\s]+)(\/[^?]*)?(\?.*)?/;
+  const regex = /^(https?:)\/\/?([^\/\s]+)(\/[^?]*)?(\?.*)?/;
```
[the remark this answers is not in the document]
JanStevens: I had the exact same question, why not use new URL with a try catch 🤔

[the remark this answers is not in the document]
khuezy: Ah, great point. @JanStevens would you mind adding a commenting nothing this quirk if there isn't one already? It's not obvious w/o knowledge on how nextjs paths works.

examples/app-router/next.config.js
```diff
@@ -51,6 +51,13 @@ const nextConfig = {
         permanent: true,
         has: [{ type: "cookie", key: "from", value: "wrongvalue" }],
       },
+      {
+        source: "/redirect",
```
conico974, line 55: `/redirect` is already used in the middleware for a redirect. This makes the middleware redirect test fail

examples/pages-router/next.config.js
```diff
@@ -22,6 +22,15 @@ const nextConfig = {
       ],
     },
   ],
+  redirects: [
+    {
+      source: "/redirect",
```
conico974, line 27: This one also fails, page router use `trailingSlash: true`. It first get redirected to `/redirect/`, then it 404

conico974 (changes requested):
This one also fail 2 test.
image

examples/pages-router/next.config.js
JanStevens: I have no idea why my change has this effect, I rather think this is another bug that now surfaces because of the e2e test 🤔. I see in `handleTrailingSlashRedirect` that this is not supported ATM? 

```
// Someone is trying to redirect to a different origin, let's not do that
    url.host !== "localhost" ||
```

Not sure how this should be tackled

conico974: Nope not another bug, just noticed this `redirects` should be a function.
In the e2e test you also need to wait for the redirect and there is multiple text that match `Open source Next.js adapter` (in the right sidebar).
I'm pushing a fix

conico974 (approved):

PR #453, opened 2024-06-26
Check if value header is undefined to avoid crash.

vladiulianbogdan:


packages/open-next/src/converters/aws-apigw-v2.ts
```diff
@@ -97,11 +97,11 @@ function convertToApiGatewayProxyResultV2(
         ),
     )
     .forEach(([key, value]) => {
-      if (value === null) {
+      if (value === null || value === undefined) {
```
khuezy, line 100: Hmm what caused the header to be set as undefined?

khuezy (approved):
LGTM

conico974 (approved):
LGTM

PR #454, opened 2024-06-26
Implement missing methods from response object to fix the middleware issue in next 14.4.12 or lower.

vladiulianbogdan:
This pull request fixes an issue where the middlewares were not functioning properly with the combination of versions specified below. Specifically, the "Middleware — redirect" functionality was failing to work as expected.

SST v2.43.2
Next.js version : 13.4.12
OpenNext v3.0.2

The error was the following:

```
TypeError: res.send is not a function
    at Object.fn (/var/task/node_modules/next/dist/server/next-server.js:1933:33)
    at async Router.execute (/var/task/node_modules/next/dist/server/router.js:315:32)
    at async NextNodeServer.runImpl (/var/task/node_modules/next/dist/server/base-server.js:624:29)
    at async NextNodeServer.handleRequestImpl (/var/task/node_modules/next/dist/server/base-server.js:549:20)
```

The implementation adds the missing methods used by next.

### DIFF
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -306,4 +306,23 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
     }
     callback();
   }
+
+  get sent() {
+    return this.finished || this.headersSent;
+  }
+
+  getHeaderValues(name: string): string[] | undefined {
+    const values = this.getHeader(name);
+
+    if (values === undefined) return undefined;
+
+    return (Array.isArray(values) ? values : [values]).map((value) =>
+      value.toString(),
+    );
+  }
+
+  send() {
+    const body = this.body;
+    this.end(body);
+  }
 }


packages/open-next/src/http/openNextResponse.ts
```diff
@@ -306,4 +306,23 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
   }
+
+  get sent() {
+    return this.finished || this.headersSent;
+  }
+
+  getHeaderValues(name: string): string[] | undefined {
+    const values = this.getHeader(name);
+
+    if (values === undefined) return undefined;
```
khuezy, line 317: Would it be possible if the header was somehow set to `null`?

conico974: It's mostly copied from next code directly, i assume it must be fine https://github.com/vercel/next.js/blob/6474cf0fe3046c148aceb7f6879c959d778457b0/packages/next/src/server/base-http/node.ts#L124-L132

conico974 (approved):
LGTM Thanks!

PR #455, opened 2024-06-26
Pass __nextDataReq in the middleware query params if already present

vladiulianbogdan:
I've noticed that the response returned by the _next/data/.../middleware-geolocation.json endpoint is an entire HTML page instead of a JSON containing the props. I've also deployed the same app on Vercel and it responds with a JSON.

After a discussion with @conico974 on Discord, we came up with this fix.


khuezy (commented):
LGTM, but deferring to @conico974 since you two had the discussion.

conico974 (approved):
LGTM

PR #461, opened 2024-07-02
[windows] Add Windows compatibility for the `resolve` ESBuild plugin

costinsin:
Previously, the separator for the filter regex was hardcoded to `/` (Unix separator).

Now, we're also taking account for the `\` (Windows separator) in the filter regex.

### DIFF
--- a/packages/open-next/src/plugins/resolve.ts
+++ b/packages/open-next/src/plugins/resolve.ts
@@ -44,7 +44,7 @@ export function openNextResolvePlugin({
     name: "opennext-resolve",
     setup(build) {
       logger.debug(`OpenNext Resolve plugin for ${fnName}`);
-      build.onLoad({ filter: /core\/resolve.js/g }, async (args) => {
+      build.onLoad({ filter: /core(\/|\\)resolve\.js/g }, async (args) => {
         let contents = readFileSync(args.path, "utf-8");
         //TODO: refactor this. Every override should be at the same place so we can generate this dynamically
         if (overrides?.wrapper) {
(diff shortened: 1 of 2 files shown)


packages/open-next/package.json
```diff
@@ -46,7 +46,8 @@
     "chalk": "^5.3.0",
     "esbuild": "0.19.2",
     "path-to-regexp": "^6.2.1",
-    "promise.series": "^0.2.0"
+    "promise.series": "^0.2.0",
+    "escape-string-regexp": "^5.0.0"
```
conico974, line 50: Do we really need to install a dependencies just for this ?

packages/open-next/src/plugins/resolve.ts
```diff
@@ -44,7 +46,11 @@ export function openNextResolvePlugin({
     name: "opennext-resolve",
     setup(build) {
       logger.debug(`OpenNext Resolve plugin for ${fnName}`);
-      build.onLoad({ filter: /core\/resolve.js/g }, async (args) => {
+      const filter = new RegExp(
+        escapeStringRegexp(path.join("core", "resolve.js")),
```
conico974, line 50: If the issue is just the `/`, can't we just do a simple regex like that instead
```ts
/core(\/|\\)resolve.js/g
```

costinsin: I have removed the need for an extra dependency in favor of including a `\\` in the regex. You are right, it might not be worth adding a new project dependency just for this.

conico974 (approved):
LGTM! Thanks

PR #465, opened 2024-07-08
[windows] Add Windows compatibility for the `replacement` ESBuild plugin

costinsin:
Previously, the separator for the filter regex was hardcoded to `/` (Unix separator).

Now, we're also taking account for the `\` (Windows separator) in the filter regex.

### DIFF
--- a/.changeset/shy-cups-notice.md
+++ b/.changeset/shy-cups-notice.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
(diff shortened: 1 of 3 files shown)


khuezy (commented):
Can we create a utils function to handle this.
eg:
`toTarget('/core/util.js`)` would generate the regex w/ forward and backslash.

While you're here improving windows support, we should add `windows` to the e2e test matrix.

khuezy:
Sounds good to me.

khuezy (approved):

PR #470, opened 2024-07-12
Fix issues with revalidateTag/revalidatePath

conico974:
- Fix `revalidateTag` for next 15+
- Fix `revalidatePath` not clearing the fetch cache ( When doing a `revalidatePath('/whatever')`, all the `fetch` or `unstable_cache` inside this route would return the cached value instead of refetching )

There is a few things that we need to think about:
Right now the fetch cache in OpenNext works in a different way than in `next start` to take full advantage of DynamoDB.
In `next start` for every get to the incremental cache, they check for every tag used in this get ( and this include layout, page and fetch cache tags ), we only check given the path which entry in DDB is newer, and if so consider the cache entry stale.  This mean we can make one single optimized query to DDB instead of either multiple get or an expensive scan/query.
On the other hand the current DDB solution requires a lot more write than the default cache included in next ( They only write on `revalidateTag` )

I think in the future we should probably support the 2 type of cache ( The next cache could be easily implemented in SQL for example, and it's way better to do it that way in SQL than the DDB way ). Even in DDB, someone barely using on demand revalidation would benefit from a next type of cache.

The other thing is, we may want to allow the `revalidatePath` to not clear the fetch cache, this gives a lot more flexibility and would allow to invalidate only what's really needed ( But that's not how next works right now ). Not sure if that's something that people may want. We could introduce an env variable that if set would make it work like right now  ( it's still possible to clear the fetch cache individually using `revalidateTag` and some custom tag )

### DIFF
--- a/packages/open-next/src/cache/tag/dynamodb-lite.ts
+++ b/packages/open-next/src/cache/tag/dynamodb-lite.ts
@@ -70,6 +70,7 @@ const tagCache: TagCache = {
 
       const tags = Items?.map((item: any) => item.tag.S ?? "") ?? [];
       debug("tags for path", path, tags);
+      // We need to remove the buildId from the path
       return tags.map((tag: string) => tag.replace(`${NEXT_BUILD_ID}/`, ""));
     } catch (e) {
       error("Failed to get tags by path", e);
--- a/packages/open-next/src/cache/tag/dynamodb-lite.ts
+++ b/packages/open-next/src/cache/tag/dynamodb-lite.ts
@@ -70,7 +70,7 @@ const tagCache: TagCache = {
 
       const tags = Items?.map((item: any) => item.tag.S ?? "") ?? [];
       debug("tags for path", path, tags);
-      return tags;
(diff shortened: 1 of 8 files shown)


sommeeeer:
wow @conico974 did you really just fix this today aswell :clap: 

you are such an inspiration!

packages/open-next/src/cache/tag/dynamodb-lite.ts
```diff
@@ -70,7 +70,7 @@ const tagCache: TagCache = {
 
       const tags = Items?.map((item: any) => item.tag.S ?? "") ?? [];
       debug("tags for path", path, tags);
-      return tags;
+      return tags.map((tag: string) => tag.replace(`${NEXT_BUILD_ID}/`, ""));
```
khuezy, line 73: Add the same comment in `dynamodb.ts` too?
`// We need to remove the buildId from the path`

khuezy (approved):
LGTM!

PR #480, opened 2024-07-24
Feat external cache

conico974:
This PR allow the ISR/SSG cache to resolve before reaching `NextServer`.
This has 2 big benefits: 
- It reduces cold start because we don't need to load the js for the route for ISR/SSG ( It could have a big impact for path that match multiple route )
- It allows the cache hit to happen directly inside the external middleware ( This will allow us to make PPR works as intended inside the CDN and not in the server as in standalone mode)
This one probably need to be a Minor release as it will require some changes to the IAC ( external middleware will need extra permissions )

It is still WIP : 
- [x] Make it work with external middleware
- [ ] ~~Maybe implement PPR here ??? (It might be better in a subsequent PR as this might require a lot of changes)~~

Some notes around PPR : 
We probably want next to settle on an implementation, because as of right now there is no way to do it outside of minimal mode.
There is a dirty solution for this, we could have 2 version of NextServer running inside the handler, one in minimalMode for PPR and the normal one for the rest.
Maybe they'll decide to allow this for anyone and not only on vercel. ( we just need to have acces to the `_next/resume` endpoints )

### DIFF
--- a/packages/open-next/src/core/routing/cacheInterceptor.ts
+++ b/packages/open-next/src/core/routing/cacheInterceptor.ts
@@ -2,6 +2,7 @@ import { createHash } from "node:crypto";
 
 import { NextConfig, PrerenderManifest } from "config/index";
 import { InternalEvent, InternalResult } from "types/open-next";
+import { emptyReadableStream, toReadableStream } from "utils/stream";
 
 import { debug } from "../../adapters/logger";
 import { CacheValue } from "../../cache/incremental/types";
@@ -109,7 +110,7 @@ async function generateResult(
   return {
     type: "core",
     statusCode: 200,
-    body,
+    body: toReadableStream(body, false),
     isBase64Encoded: false,
     headers: {
       ...cacheControl,
@@ -180,7 +181,7 @@ export async function cacheInterceptor(
           return {
             type: "core",
             statusCode: cachedData.value.meta?.status ?? 307,
-            body: "",
+            body: emptyReadableStream(),
             headers: {
               ...((cachedData.value.meta?.headers as Record<string, string>) ??
                 {}),
--- a/packages/open-next/src/core/routing/cacheInterceptor.ts
+++ b/packages/open-next/src/core/routing/cacheInterceptor.ts
@@ -1,0 +1,172 @@
+import { createHash } from "node:crypto";
+
+import { PrerenderManifest } from "config/index";
+import { InternalEvent, InternalResult } from "types/open-next";
+
+import { debug } from "../../adapters/logger";
+import { CacheValue } from "../../cache/incremental/types";
+import { generateMessageGroupId } from "./util";
+
+const CACHE_ONE_YEAR = 60 * 60 * 24 * 365;
+const CACHE_ONE_MONTH = 60 * 60 * 24 * 30;
+
+async function computeCacheControl(
+  path: string,
+  body: string,
+  host: string,
+  revalidate?: number | false,
+  lastModified?: number,
+) {
+  let finalRevalidate = CACHE_ONE_YEAR;
+
+  const existingRoute = Object.entries(PrerenderManifest.routes).find(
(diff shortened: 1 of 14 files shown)


conico974:
I think this one is done and should work in most cases.
As for a proper PPR implementation, almost everything is in place but i think we really need to wait to see what they end up doing with the upstream implementation.

In previous implementation they provided the prerendered html at build time (it might still be the case in some occasion as the code is still there) and the only thing left to do to recreate the complete html was to create a stream, push the initial html into the stream, and then push the result of the `_next/resume` endpoint into it and it was done.

At the moment, it looks like they do not provide the prerendered html anymore and we have to reconstruct it. A dirty way that seemed to work was to just create a shell html like that and use it (It's very likely broken in a lot of cases)
```html
<!DOCTYPE html><html>
  <head>
    <meta charSet="utf-8"/>
    <meta name="viewport" content="width=device-width, initial-scale=1"/>
   <!-- Here we insert every these 2 necessary script for the page  -->
    <script src="http://localhost:3000/_next/static/chunks/webpack-6540e76769ad5f68.js" async=""></script>
    <script src="http://localhost:3000/_next/static/chunks/main-app-a079acdba6bedcac.js" async=""></script>
  </head>
  <body>
```
The proper way to do it would be to use the `prerender` function from `react-dom/static.edge` with all the correct next provider as children, but it means a lot of reverse engineering. But hopefully they'll just prerender the html once it is ready

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -0,0 +1,202 @@
...
+    const sMaxAge = Math.max(finalRevalidate - age, 1);
+    debug("sMaxAge", {
+      finalRevalidate,
+      age,
+      lastModified,
+      revalidate,
+    });
+    const isStale = sMaxAge === 1;
+    if (isStale) {
+      const url = NextConfig.trailingSlash ? `${path}/` : path;
```
khuezy, line 58: Would this handle the case of `basePath`, or is that irrelevant here?

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -0,0 +1,202 @@
...
+export async function cacheInterceptor(
+  event: InternalEvent,
+): Promise<InternalEvent | InternalResult> {
+  if (
+    Boolean(event.headers["next-action"]) ||
+    Boolean(event.headers["x-prerender-revalidate"])
+  )
+    return event;
+  // We localize the path in case i18n is enabled
+  let localizedPath = localizePath(event);
```
khuezy, line 132: Same question as above, `basePath` affecting this?

khuezy (approved):
A couple question regarding `basePath`. Other than that, my review is only for the code. I haven't verified the deployment and runtime.
LGTM

packages/open-next/src/core/routing/cacheInterceptor.ts
conico974: Good call, this probably won't work with `basePath`. 
If this options is set i'm pretty sure we need to remove the basePath before doing anything here. ( I need to verify this, i might be wrong )
To be fair it will be pretty much invisible for the user, it will just fail to find anything here, and it will forward the request to NextServer which should find it as expected (even using the cache)

packages/open-next/src/core/routing/cacheInterceptor.ts
conico974: Same as above, actually we probably need to do this here only

packages/open-next/src/core/routing/cacheInterceptor.ts
conico974: I'll finish this tomorrow morning with the fix for the `basePath`. Hopefully the fix is not too hard and everything works in e2e :crossed_fingers:

PR #486, opened 2024-08-06
Feat readable body

conico974:
This PR change the body of `InternalResult` from `string` to `ReadableStream`.
This allow for edge runtime function to use streaming, and it is necessary to prepare for PPR.

This should also remove the need to add a `fakeStream` in non streaming wrapper as the stream is now always consumed if you return a body.

### DIFF
--- a/packages/open-next/src/utils/stream.ts
+++ b/packages/open-next/src/utils/stream.ts
@@ -36,5 +36,8 @@ export function toReadableStream(
 }
 
 export function emptyReadableStream(): ReadableStream {
+  if (process.env.OPEN_NEXT_FORCE_NON_EMPTY_RESPONSE === "true") {
+    return Readable.toWeb(Readable.from(["SOMETHING"]));
+  }
   return Readable.toWeb(Readable.from([]));
 }
--- a/packages/open-next/src/utils/stream.ts
+++ b/packages/open-next/src/utils/stream.ts
@@ -1,0 +1,40 @@
+import { Readable } from "node:stream";
+import type { ReadableStream } from "node:stream/web";
+
+export function fromReadableStream(
+  stream: ReadableStream<Uint8Array>,
+  base64?: boolean,
+): Promise<string> {
+  const reader = stream.getReader();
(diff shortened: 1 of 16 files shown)


packages/open-next/src/utils/stream.ts
```diff
@@ -0,0 +1,43 @@
...
+  isBase64?: boolean,
+): ReadableStream {
+  return Readable.toWeb(
+    Readable.from(Buffer.from(value, isBase64 ? "base64" : "utf8")),
+  );
+}
+
+export function emptyReadableStream(): ReadableStream {
+  if (process.env.OPEN_NEXT_FORCE_NON_EMPTY_RESPONSE === "true") {
+    return Readable.toWeb(Readable.from(["SOMETHING"]));
```
khuezy, line 40: Can this be an empty string or does it have to be a non-empty string?

khuezy (approved):
Just 1 comment, otherwise my tiny grain of salt LGTM

packages/open-next/src/utils/stream.ts
conico974: Empty string didn't work, we could have used `new Uint8Array(8)` but since it is used as a delimiter for the header part of the string i fear that this could introduce other issue later ( being misinterpreted by the streaming server )

packages/open-next/src/wrappers/aws-lambda-streaming.ts
```diff
@@ -95,7 +98,18 @@ const handler: WrapperHandler = async (handler, converter) =>
 
       const response = await handler(internalEvent, streamCreator);
 
-      return converter.convertTo(response);
+      // If at this point the headers have not been written, it means that we are using the edge runtime
+      if (!headersWritten) {
```
conico974, line 102: Actually this one is causing trouble. When using ISR/SSG `headersWritten` is false at this point (probably some kind of race condition as this is done on a stream )

`writeHeaders` below will throw an error, but somehow it still works for the user.

PR #487, opened 2024-08-08
fix lambda streaming hanging after return

conico974:
It looks like aws has made an update again to how the runtime works when using streaming.
Now even after `end()` has been called on the stream and the handler returned, the lambda keeps running in the background doing nothing until it timeout (and you'll be billed for the full duration)

This should fix it.

### DIFF
--- a/packages/open-next/src/wrappers/aws-lambda-streaming.ts
+++ b/packages/open-next/src/wrappers/aws-lambda-streaming.ts
@@ -23,7 +23,12 @@ function formatWarmerResponse(event: WarmerEvent) {
 
 const handler: WrapperHandler = async (handler, converter) =>
   awslambda.streamifyResponse(
-    async (event: AwsLambdaEvent, responseStream): Promise<AwsLambdaReturn> => {
+    async (
+      event: AwsLambdaEvent,
+      responseStream,
+      context,
+    ): Promise<AwsLambdaReturn> => {
+      context.callbackWaitsForEmptyEventLoop = false;
       if ("type" in event) {
         const result = await formatWarmerResponse(event);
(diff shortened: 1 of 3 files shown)


khuezy (approved):
AWS!!!

packages/open-next/src/wrappers/aws-lambda-streaming.ts
```diff
@@ -23,7 +23,12 @@ function formatWarmerResponse(event: WarmerEvent) {
 
 const handler: WrapperHandler = async (handler, converter) =>
   awslambda.streamifyResponse(
-    async (event: AwsLambdaEvent, responseStream): Promise<AwsLambdaReturn> => {
+    async (
+      event: AwsLambdaEvent,
+      responseStream,
+      context,
+    ): Promise<AwsLambdaReturn> => {
+      context.callbackWaitsForEmptyEventLoop = false;
```
khuezy, line 31: Didn't Dax add this to the `sst` layer at one point, which broken `open-next`?

conico974: Did he ? He did it in the remix server in ion, and at the time i tried it in OpenNext as well ( to fix the 30s hanging issue ), but it ended up causing issue with ISR because of the pending promise ( at the time i didn't know that ).

khuezy: I may have misremembered, but iirc he added that to the base SSR construct in `sst`. Things may be changed now, but it was many months ago.

conico974: You're probably right actually, i thought you were talking about the remix stuff. 
Anyway it works now

PR #497, opened 2024-08-26
fix issue with fetch cache for new page

### LINKED ISSUE #493: OpenNext Ignores force-cache Option on fetch Requests (Data Cache Behavior Differs from Vercel)
I recently migrated my project to OpenNext, deploying it with `sst@v3`. It's a news website built using a custom CMS developed with Next.js. Previously, when hosted on Vercel, nearly all fetch requests to the configuration endpoint were served from cache (with a 99.99% cache hit rate, according to Vercel and our server logs).

However, after the migration, the server is being hit hundreds of times per hour, suggesting that the cache is being bypassed—likely whenever new pages are rendered. This behavior differs significantly from our previous experience on Vercel, where caching worked as expected.

### Problem

- **Expected Behavior**: Fetch requests should be cached, with minimal server hits (as seen on Vercel).
- **Current Behavior**: The server is hit far more frequently, implying the cache is not functioning as intended.

### Questions

1. What could be causing this cache issue after migrating to OpenNext?
2. How can I debug or resolve this problem?

I'd greatly appreciate any help, insights, or advice on this issue.

Thank you in advance!

### Useful Links

- [Next.js Fetch API Documentation](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching#fetch-api)
- [Next.js Fetch Options (cache)](https://nextjs.org/docs/app/api-reference/functions/fetch#optionscache)

conico974:
When creating new ISR page at runtime, inserted tags would make the fetch cache stale.
This will also fix cache miss when deploying using the same prebuilt data (i.e. not running `next build` again)
Fix #493

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -147,27 +147,24 @@ export default class S3Cache {
 
       if (value === undefined) return null;
 
-      // For cases where we don't have tags, we need to ensure that we insert at least an entry
-      // for this specific paths, otherwise we might not be able to invalidate it
+      // For cases where we don't have tags, we need to ensure that the soft tags are not being revalidated
+      // We only need to check for the path as it should already contains all the tags
       if ((tags ?? []).length === 0) {
-        // First we check if we have any tags for the given key
-        const storedTags = await globalThis.tagCache.getByPath(key);
-        if (storedTags.length === 0) {
-          // Then we need to find the path for the given key
-          const path = softTags?.find(
-            (tag) =>
-              tag.startsWith("_N_T_/") &&
-              !tag.endsWith("layout") &&
-              !tag.endsWith("page"),
+        // Then we need to find the path for the given key
+        const path = softTags?.find(
+          (tag) =>
+            tag.startsWith("_N_T_/") &&
+            !tag.endsWith("layout") &&
+            !tag.endsWith("page"),
+        );
+        if (path) {
+          const pathLastModified = await globalThis.tagCache.getLastModified(
+            path.replace("_N_T_/", ""),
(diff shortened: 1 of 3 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -338,6 +335,9 @@ export default class S3Cache {
           tagsToWrite.map((tag) => ({
             path: key,
             tag: tag,
+            // In case the tags are not there we just need to create them
+            // but we don't want them to return frrom `getLastModified` as they are not stale
```
khuezy, line 339:
```suggestion
            // but we don't want them to return from `getLastModified` as they are not stale
```

khuezy (approved):
LGTM, just a typo in the comment.

packages/open-next/src/adapters/cache.ts
```diff
@@ -147,27 +147,24 @@ export default class S3Cache {
 
       if (value === undefined) return null;
 
-      // For cases where we don't have tags, we need to ensure that we insert at least an entry
-      // for this specific paths, otherwise we might not be able to invalidate it
+      // For cases where we don't have tags, we need to ensure that the soft tags are not being revalidated
+      // We only need to check for the path as it should already contains all the tags
```
sommeeeer, line 151: Typo here too. Should be contain

sommeeeer (changes requested):
Another typo, should be contain

packages/open-next/src/adapters/cache.ts
sommeeeer: Typo here too.
```suggestion
Replace "contains" with "contain"
```

PR #502, opened 2024-09-01
fix wrong locale in middleware

### LINKED ISSUE #500: nextUrl.locale is not updated with pathname locale in middleware
When using NextJS subpath routing and i18n configuration and attempting to do some simple redirect to handle the locale notice that the `req.nextUrl.locale` is not being updated with the pathname locale.

Essentially NextJS Internatlonalization allows you handle localization with path routing and when changing locales via the path route this should be reflected in the `req.nextUrl.locale`. So if you visit `/pt/home` then the `req.nextUrl.locale` would have `pt` however it seems when releasing this via SST this is not reflected and it is always using the `defaultLocale`. 

I attempted to do some digging in the code and noticed the `detectLocale` function is slightly different to that of NextJS: https://github.com/vercel/next.js/blob/f4c1a406e420e5ee5c01040e642a2e2fcc97f72f/packages/next/src/shared/lib/i18n/get-locale-redirect.ts#L48 where they set the `pathname.locale` and in open-next this is not included. I haven't tested to see if that would solve my problem but currently I am running into a Too many redirects error because the variable is not properly reflecting the language value and so keeps trying to redirect to the language.

Reproduction here: https://github.com/hect1c/sst-standard-nextjs-repro

- Once released to AWS if you visit the `site.url` provided and just append `/pt` to the url you will notice in the server function logs that the `req.nextUrl.locale` is `en` when it should be `pt`.
- Also if you add the cookie `test.locale=pt` and then visit the `site.url` again it will attempt to redirect you yo `/pt` which it does but since the`req.nextUrl.locale` on the redirect is still `en` it will keep trying to redirect causing the error

conico974:
Fix #500

### DIFF
--- a/.changeset/silly-hounds-admire.md
+++ b/.changeset/silly-hounds-admire.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+fix wrong locale in middleware
--- a/packages/open-next/src/core/routing/middleware.ts
+++ b/packages/open-next/src/core/routing/middleware.ts
@@ -59,6 +59,7 @@ export async function handleMiddleware(
 ): Promise<MiddlewareOutputEvent | InternalResult> {
   const { rawPath, query } = internalEvent;
   const normalizedPath = normalizeLocalePath(rawPath);
+  // We only need the normalizedPath to check if the middleware should run
   const hasMatch = middleMatch.some((r) => r.test(normalizedPath));
   if (!hasMatch) return internalEvent;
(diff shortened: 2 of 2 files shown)


khuezy (approved):

hect1c:
Any idea by when we may be able to release this? In the meantime will try and do a patch to get this through our pipeline

conico974:
@hect1c I'll try to make a release tomorrow

PR #503, opened 2024-09-01
patch asyncStorage for ISR and fetch

### LINKED ISSUE #493: OpenNext Ignores force-cache Option on fetch Requests (Data Cache Behavior Differs from Vercel)

conico974:
ISR request are treated as On demand revalidation by Next, and it makes the fetch data cache not work as expected (the fetch cache is bypassed for every ISR revalidation request)
Should fix #493 

I'm not a very big fan of this solution but we either do that or patch some files in Next code itself and it's probably even worse.

### DIFF
--- a/.changeset/mighty-elephants-design.md
+++ b/.changeset/mighty-elephants-design.md
@@ -1,0 +1,5 @@
+---
+"open-next": patch
+---
+
+patch asyncStorage for ISR and fetch
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -1,3 +1,5 @@
+import url from "node:url";
+
 import {
   copyFileSync,
   existsSync,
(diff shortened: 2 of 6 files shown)


khuezy (approved):
looks crazy to me

PR #506, opened 2024-09-07
fix cookies improperly set on node

conico974:
When setting multiple cookies using the node wrapper it will fail. This should fix it


conico974:
Closing in favor of #514 Fix are already applied there

PR #513, opened 2024-09-12
Improve config validation

conico974:
This PR improve `open-next.config.ts` file validation so that it's easier to understand what's wrong.
It also provides a better error when a route cannot be found for function splitting.

### DIFF
--- a/packages/open-next/src/build/validateConfig.ts
+++ b/packages/open-next/src/build/validateConfig.ts
@@ -1,10 +1,24 @@
 import {
   FunctionOptions,
+  IncludedConverter,
+  IncludedWrapper,
   OpenNextConfig,
   SplittedFunctionOptions,
 } from "types/open-next";
 
 import logger from "../logger.js";
+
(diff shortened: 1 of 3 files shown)


packages/open-next/src/build/validateConfig.ts
```diff
@@ -1,26 +1,68 @@
...
+  IncludedConverter,
+  IncludedWrapper,
   OpenNextConfig,
   SplittedFunctionOptions,
 } from "types/open-next";
 
 import logger from "../logger.js";
 
+const compatibilityMatrix: Record<IncludedWrapper, IncludedConverter[]> = {
+  "aws-lambda": [
```
khuezy, line 12: Would be nice to add this matrix test in a unit test, but perhaps we can do that in the new org

khuezy (approved):
LGTM

packages/open-next/src/build/validateConfig.ts
conico974: Yeah i was thinking we could do it in the new org as well.

PR #514, opened 2024-09-12
Fix cloudflare env

conico974:
This PR fix an issue with cloudflare env variable being undefined if the middleware does not run.
This was breaking `enableCacheInterception` with the middleware in cloudflare.

This PR also fix some issue with cookies being not properly set when set both in the routing layer (i.e. middleware or next.config) and the route itself.
It also fix an issue with cookies and the `node` wrapper

### DIFF
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -159,22 +159,26 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
     this.headersSent = true;
     // Initial headers should be merged with the new headers
     // These initial headers are the one created either in the middleware or in next.config.js
-    // We choose to override response headers with middleware headers
-    // This is different than the default behavior in next.js, but it allows more customization
-    // TODO: We probably want to change this behavior in the future to follow next
-    // We could add a prefix header that would allow to force the middleware headers
-    // Something like open-next-force-cache-control would override the cache-control header
+    const mergeHeadersPriority =
+      globalThis.__als?.getStore()?.mergeHeadersPriority ?? "middleware";
     if (this.initialHeaders) {
-      this.headers = {
-        ...this.headers,
-        ...this.initialHeaders,
-      };
+      this.headers =
+        mergeHeadersPriority === "middleware"
+          ? {
+              ...this.headers,
+              ...this.initialHeaders,
+            }
+          : {
+              ...this.initialHeaders,
+              ...this.headers,
+            };
       const initialCookies = parseCookies(
         (this.initialHeaders[SET_COOKIE_HEADER] as string | string[]) ?? [],
       ) as string[];
-      //Do we want to filter out the cookies that are already set?
-      // At the moment cookies from the middlewware will override the ones set in the route
-      this._cookies = [...this._cookies, ...initialCookies];
+      this._cookies =
+        mergeHeadersPriority === "middleware"
+          ? [...this._cookies, ...initialCookies]
+          : [...initialCookies, ...this._cookies];
     }
     this.fixHeaders(this.headers);
 
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -176,10 +176,6 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
       };
     }
     this.fixHeaders(this.headers);
-    if (this._cookies.length > 0) {
-      // For cookies we cannot do the same as for other headers
-      this.headers[SET_COOKIE_HEADER] = this._cookies;
-    }
 
     if (this.streamCreator) {
       this.responseStream = this.streamCreator?.writeHeaders({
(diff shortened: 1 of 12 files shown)


packages/open-next/src/wrappers/cloudflare.ts
```diff
@@ -12,8 +12,16 @@ const handler: WrapperHandler<
 > =
   async (handler, converter) =>
   async (event: Request, env: Record<string, string>): Promise<Response> => {
-    //@ts-expect-error - process is not defined in cloudflare workers
-    globalThis.process = { env };
+    globalThis.process = process;
```
khuezy, line 15: Where is the `process` variable coming from? Is that specific to cloudflare?

conico974: I added this as a js banner in the esbuild that i did for the cloudflare worker.
Now that i think about it we should probably do this esbuild part inside of OpenNext as well. We'd get a single file out of it which is easier for cloudflare worker.

I'll publish a sample repo later with SST v3 with different way to deploy OpenNext on different target.
Right now i have EC2, ECS and lambda with either cloudfront or cloudflare in front of it ( and with cloudflare the middleware runs in the worker )
Maybe that's something we could put in the new org as well

packages/open-next/src/wrappers/node.ts
```diff
@@ -10,6 +10,7 @@ const wrapper: WrapperHandler = async (handler, converter) => {
     const internalEvent = await converter.convertFrom(req);
     const _res: StreamCreator = {
       writeHeaders: (prelude) => {
+        res.setHeader("Set-Cookie", prelude.cookies);
```
khuezy, line 13: I don't think `Set-Cookie` allows for setting multiple cookies at once (could be wrong). Need to split the cookies and `res.addHeader` if so.

conico974: Actually it does ( I thought the same thing as well at first ).

khuezy: That's really weird, according to the spec: https://httpwg.org/specs/rfc6265.html#overview it shouldn't. Since the comma character may also be part of the cookie value.

conico974: `prelude.cookies` is an array, it's from here 
https://github.com/sst/open-next/blob/faefb33a2a485c571c2cc43680f1b702a03b58c7/packages/open-next/src/http/openNextResponse.ts#L187

sommeeeer: It will work with an array.

khuezy (approved):

packages/open-next/src/wrappers/node.ts
conico974: Actually there was an issue here ( i forgot to remove the custom wrapper on my test :roll_eyes: ).
We needed to remove the `set-cookie` header from the headers, otherwise the broken one would take precedence.
There was also another issue with the set-cookie that affected everything. If cookies were set both in the middleware and in a route, only the middleware ones would be applied

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -174,18 +169,29 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
         ...this.headers,
         ...this.initialHeaders,
       };
+      const initialCookies = parseCookies(
+        (this.initialHeaders[SET_COOKIE_HEADER] as string | string[]) ?? [],
+      ) as string[];
+      //Do we want to filter out the cookies that are already set?
```
khuezy, line 175:
```suggestion
      // Do we want to filter out the cookies that are already set?
```

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -174,18 +169,29 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
+      // At the moment cookies from the middlewware will override the ones set in the route
```
khuezy, line 176:
```suggestion
      // At the moment cookies from the middleware will override the ones set in the route
```

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -174,18 +169,29 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
+      this._cookies = [...this._cookies, ...initialCookies];
```
khuezy, line 177: Good question, we should ask the community? In terms of flow, middleware comes before the route so I'd assume the route should have the final say in the cookies. Just my pov, but you could argue the other way depending on your use case.

conico974: Yeah we could ask the community.
Right now for headers, middleware will override route, it makes sense that it's the same for cookies. The thing with the route having the final say is that in some case ( app router for example ) there is no way to modify the headers others than by having the middleware take precedence. 

There is also another option here, we could use an env variable (or an option in the config) to decide which should take priority.

Issue #522, opened 2024-09-23
basePath does not works as expected with API routes


socsieng:
## Issue

NextJS app deployed using SST and OpenNext does not respect `basePath` property defined in NextJS config for API routes.

I'd love to work on this but unfortunately I don't know where to start.

Potentially related to: https://github.com/sst/open-next/issues/508

## Environment

- Platform: macOS Sequoia 15.0 (24A335)
- NodeJS: 22.3.0
- Package manager: pnpm 9.7.0
- NextJS: 14.2.11
- OpenNext: 3.1.3
- SST: 3.1.11

## Reproduction steps

Clone repository at https://github.com/socsieng/next-sst which includes detailed reproduction steps. Reproduction steps also included inline.

### 1. `create-next-app` and initialize SST

```bash
npx create-next-app@latest next-sst

cd next-sst
sst init
```

### 2. Update the `next.config.mjs` file to include the `basePath`

```javascript
/** @type {import('next').NextConfig} */
const nextConfig = {
  basePath: '/base-path',
};

export default nextConfig;
```

### 3. Add an API route at `app/api/hello/route.ts`

```typescript
import { NextRequest, NextResponse } from "next/server";

// eslint-disable-next-line @typescript-eslint/no-unused-vars
export async function GET(request: NextRequest) {
  return NextResponse.json({ message: "Hello world!" });
}
```

### 4. Use latest version of `ope-next` in `sst.config.ts`

`3.1.3` at the time of writing.

```typescript
/// <reference path="./.sst/platform/config.d.ts" />

export default $config({
  app(input) {
    return {
      name: "next-sst",
      removal: input?.stage === "production" ? "retain" : "remove",
      home: "aws",
    };
  },
  async run() {
    new sst.aws.Nextjs("MyWeb", {
      openNextVersion: "3.1.3"
    });
  },
});
```

### 5. Deploy the app

```bash
sst deploy --stage api-test
```

### 6. Open the following URLs in the browser

1. https://assignedhost.cloudfront.net/base-path
2. https://assignedhost.cloudfront.net/base-path/api/hello

Replace `assignedhost` with the actual CloudFront distribution URL.

## Expected results

Both 6.1 and 6.2 return successful responses.

## Actual results

6.1 returns a successful response, but 6.2 returns a 404 error. Note that changing the URL for 6.2 to exclude the `basePath` (https://assignedhost.cloudfront.net/api/hello) works.

Note that this works as expected with `sst dev`:

[...]

PR #523, opened 2024-09-23
fix: fix basePath support for API routes

socsieng:
Fixes: #522

I've verified changes locally but I couldn't find a way to easily include unit tests for this change. Please let me know if there are any additional changes you'd like me to make.

### DIFF
--- a/packages/open-next/src/core/routingHandler.ts
+++ b/packages/open-next/src/core/routingHandler.ts
@@ -34,6 +34,11 @@ const optionalBasepathPrefixRegex = !!RoutesManifest.basePath
   ? `^${RoutesManifest.basePath}/?`
   : "^/";
 
+// Add the basePath prefix to the api routes
+const apiPrefix = !!RoutesManifest.basePath
+  ? `${RoutesManifest.basePath}/api`
+  : "/api";
+
 const staticRegexp = RoutesManifest.routes.static.map(
   (route) =>
     new RegExp(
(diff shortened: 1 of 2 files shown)


packages/open-next/src/core/routingHandler.ts
```diff
@@ -144,9 +145,12 @@ export default async function routingHandler(
   // Api routes are not present in the routes manifest except if they're not behind /api
   // /api even if it's a page route doesn't get generated in the manifest
   // Ideally we would need to properly check api routes here
+  const apiPrefix = RoutesManifest.basePath
+    ? path.join("/", RoutesManifest.basePath, "api")
```
conico974, line 149: Do we really need `path/posix` here ? A template literal is probably enough here unless i'm missing something
This part could run in a different runtime than node and `path/posix` might not be available there.

The `apiPrefix` could also be computed outside the function, no need to recompute this for every route. Not a big deal though

.changeset/purple-clouds-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"open-next": patch
```
conico974, line 2: Could you remove the changeset for now, i'll add this again before release
The repo will change name very soon and this will cause trouble with changeset

conico974 (changes requested):
Thanks for the PR.
Just a few minor things to change though

packages/open-next/src/core/routingHandler.ts
socsieng: > Do we really need `path/posix` here ? A template literal is probably enough here unless i'm missing something
> This part could run in a different runtime than node and `path/posix` might not be available there.

I was trying to be clever about supporting `basePath` values with different combinations of leading/trailing `/`. Given the [NextJS docs](https://nextjs.org/docs/app/api-reference/next-config-js/basePath) uses `/docs` as the example and that existing code already makes similar assumptions about `basePath`, a template literal works.

conico974 (approved):
LGTM. Thanks

PR #524, opened 2024-09-26
lazily initialize lite client

conico974:
In some cases the env variable are not populated at import time, this make the lite client being able to handle this.

This also allow for some other use cases such as being able to use different region or different table/bucket/queue, by defining these env variable at runtime.

### DIFF
--- a/packages/open-next/src/cache/incremental/s3-lite.ts
+++ b/packages/open-next/src/cache/incremental/s3-lite.ts
@@ -30,7 +30,7 @@ const awsFetch = async (key: string, options: RequestInit) => {
   const { CACHE_BUCKET_REGION, CACHE_BUCKET_NAME } = process.env;
   const client = getAwsClient();
   const url = `https://${CACHE_BUCKET_NAME}.s3.${CACHE_BUCKET_REGION}.amazonaws.com/${key}`;
-  return customFetchClient(client)(url, options).finally(console.log);
+  return customFetchClient(client)(url, options);
 };
 
 function buildS3Key(key: string, extension: Extension) {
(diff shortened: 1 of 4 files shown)


packages/open-next/src/cache/incremental/s3-lite.ts
```diff
@@ -7,24 +8,33 @@ import { parseNumberFromEnv } from "../../adapters/util";
...
+    return awsClient;
+  }
+};
 
-const awsFetch = customFetchClient(awsClient);
+const awsFetch = async (key: string, options: RequestInit) => {
+  const { CACHE_BUCKET_REGION, CACHE_BUCKET_NAME } = process.env;
+  const client = getAwsClient();
+  const url = `https://${CACHE_BUCKET_NAME}.s3.${CACHE_BUCKET_REGION}.amazonaws.com/${key}`;
+  return customFetchClient(client)(url, options).finally(console.log);
```
khuezy, line 33: Should we be using `debug` here instead of `console.log`?

packages/open-next/src/cache/incremental/s3-lite.ts
```diff
@@ -61,29 +68,21 @@ const incrementalCache: IncrementalCache = {
...
-        isFetch ? "fetch" : "cache",
-      )}`,
+      buildS3Key(key, isFetch ? "fetch" : "cache"),
       {
         method: "PUT",
         body: JSON.stringify(value),
       },
     );
     if (response.status !== 200) {
+      console.log(response);
```
khuezy, line 78: `debug` instead of `console.log`

packages/open-next/src/cache/incremental/s3-lite.ts
conico974: Yeah, it shouldn't have stayed there, let me remove this quickly

khuezy (approved):
LGTM!

PR #525, opened 2024-09-26
prepare for moving the repo

conico974:
This should be enough changes to move the repo to `@opennextjs/opennextjs-aws` and everything should work


packages/open-next/package.json
```diff
@@ -2,7 +2,7 @@
   "publishConfig": {
     "access": "public"
   },
-  "name": "open-next",
+  "name": "@opennextjs/aws",
   "version": "3.1.3",
   "bin": {
     "open-next": "./dist/index.js"
```
conico974, line 8: Do we want to keep it as `open-next` or do we want to change that as well ?

PR #529, opened 2024-10-02
fix: change reference from open-next to @opennextjs/aws

sommeeeer:
change reference from `open-next` to `@opennextjs/aws` in every 
`package.json` in the example repos. let me know if anything else needs to change.


conico974 (commented):
https://github.com/opennextjs/opennextjs-aws/actions/runs/11139867869/job/30957412300?pr=529
You'll need to update the lockfile as well, otherwise it will fail the linting, and probably a bunch of other stuff

sommeeeer:
The build failed on `example/` after updating the `pnpm` version to 9. For some reason it was not creating its own `node_modules` folder. Might be a bug with pnpm itself, as I was not able to build the `example/` locally either. Trying to revert the pnpm version now

khuezy:
If there are different package versions across multiple apps, `pnpm` may get confused and cause problems.

khuezy:
There's also: https://github.com/opennextjs/opennextjs-aws/blob/main/examples/app-router/package.json#L17C5-L17C32

conico974:
> If there are different package versions across multiple apps, `pnpm` may get confused and cause problems.

Yeah but it should fail everywhere, not just on the example package. And keeping it on the old lockfile seems to do the trick

conico974:
> There's also: https://github.com/opennextjs/opennextjs-aws/blob/main/examples/app-router/package.json#L17C5-L17C32 Sorry, wrong link. Why doesn't `pages-router` have the `open-next` dependency?

I'm not even sure why there is a dep on `open-next`. It seems that it's not even used

khuezy:
Yea, I'm not sure why it's a dep either. A mistake a long time ago? The `examples/` app builds open next internally via `node ../../packages/open-next/dist/index.js build`

conico974:
We can keep it for now, this would allow us to type the config file properly and maybe later add some custom adapter.
That would be great to add this in the e2e test as well

conico974:
Btw i think we should probably add a line here https://github.com/opennextjs/opennextjs-aws/blob/dd9c10b2569367c4a0ebb03ee6ae42b354ea4f6a/packages/open-next/package.json#L8
I think right now it won't work with npx, we'll need to add `"@opennextjs/aws": "./dist/index.js" `
We should keep the `open-next` one for compatibility for people who install open-next locally

conico974 (approved):
LGTM Thanks!!

khuezy:
```
> node ../../packages/open-next/dist/index.js build --streaming --build-command "npx turbo build"

node:internal/modules/cjs/loader:1080
  throw err;
  ^

Error: Cannot find module '/home/runner/work/opennextjs-aws/opennextjs-aws/packages/open-next/dist/index.js'
```
Hmm this is strange, did the folder structure change at all when this was migrated to the new org?

conico974:
@khuezy Found it it's the `--filter` just before in the github actions

khuezy:
Great catch! Easy to miss that step,

PR #532, opened 2024-10-03
fix content-length incorrectly set in proxyRequest

mjwbenton:
As discussed on Discord here: https://discord.com/channels/1283128968140161065/1286094576788177059/1291327268450467893

Using rewrites with an external server currently has an issue when the server responds with a compressed response. `proxyRequest` decompresses the response, and suitably filters out the `content-encoding` header from the response to indicate that it is not compressed. However currently it still passes through the `content-length` header untouched. This means that downstream the response gets truncated to the length of the compressed response even though a decompressed response is being sent.

This PR filters out the optional `content-length` header such that the content does not get truncated.

Thanks!

### DIFF
--- a/.changeset/mjwbenton.md
+++ b/.changeset/mjwbenton.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix content-length incorrectly set in proxyRequest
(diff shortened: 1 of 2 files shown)


conico974 (approved):
LGTM Thanks

PR #535, opened 2024-10-08
Fix release

conico974:
This should fix release.

I also changed the changelog so that it use `@changesets/changelog-github`. This should automatically add a thanks  with the username and a link to their github. 
Hopefully this doesn't break anything ( It should have a `GITHUB-TOKEN` so it should be fine )


khuezy (approved):
LGTM

PR #536, opened 2024-10-09
use `pkg.pr.new` to publish PR prereleases

dario-piotrowicz:


dario-piotrowicz:
Hey @conico974 👋 

@vicb mentioned that you were interested in exploring the use of `pkg.pr.new`? 🙂 

I figured I could start the effort and get the ball rolling ⚽ 

Please have a look, also please let me know, is the validation and build scripts supposed not to work? (the also don't seem to work for me locally on `main`)

.github/workflows/pre-release.yml
```diff
@@ -1,6 +1,9 @@
 name: Pre-release
 
-on: workflow_dispatch
+on:
+  pull_request:
+    branches:
+      - main
```
conico974, line 6: We probably want to filter to run this only on change on the `packages/open-next` folder itself.
Really not a big deal but change made to e2e would trigger this which is not really useful

.github/workflows/pre-release.yml
```diff
@@ -21,10 +24,8 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
-      - name: Check codestyle
-        run: pnpm lint
+      - name: Build package
+        run: pnpm -F @opennextjs/aws build
```
dario-piotrowicz, line 28: this could be `pnpm build` if preferred, I set it to only build the aws package as the build script was failing in ci for some reason 😕

conico974 (commented):
Did you run `pnpm install` with pnpm 9 ? Someone had done this when we moved the repo and it caused a ton of issues (including the build failing).
Not sure why it cause an issue (and i haven't took the time to look at it), but reverting to pnpm 8 fixed the issue. 
We'll need to figure this out but it might have something to do with a big change in the lockfile version between pnpm 8 and 9

.github/workflows/pre-release.yml
conico974: Probably related to the comment i left below about pnpm

.github/workflows/pre-release.yml
dario-piotrowicz: good point! 👍

dario-piotrowicz:
> Did you run `pnpm install` with pnpm 9 ? Someone had done this when we moved the repo and it caused a ton of issues (including the build failing). Not sure why it cause an issue (and i haven't took the time to look at it), but reverting to pnpm 8 fixed the issue. We'll need to figure this out but it might have something to do with a big change in the lockfile version between pnpm 8 and 9

Awesome! thanks yeah I tried with pnpm 9, let's give pnpm 8 a go 🙂

dario-piotrowicz:
@conico974 unfortunately using pnpm 8 didn't seem to help 🥲

conico974:
Yeah not sure what's going on here. 
This https://github.com/opennextjs/opennextjs-aws/actions/runs/11234182017 which run yesterday had absolutely no problem on those file. 
But somehow it fails today on those same file (and that's not file you touched in your PR)

My guess is that there is some kind of issue with the update in the lockfile. 
It seems to have updated a ton of dependencies and one of them is probably causing some kind of issue.

I'll try to take another look at this tomorrow

conico974:
Same issue as what happened on this PR https://github.com/opennextjs/opennextjs-aws/pull/529

dario-piotrowicz:
@conico974 all better now 🫡 

I basically restarted the whole thing from scratch, but starting using pnpm 8 from the start now and as you can see the lock file is much much much smaller... I guess pnpm was getting confused by the dependencies or something 😕  (even if I did try deleting the lock file and clearing the node_modules 🤷)

conico974 (approved):
Great, that's perfect.
We'll need to figure out what's going on with this lockfile, there is something very wrong here.
Anyway thanks for the PR

PR #537, opened 2024-10-10
update opennextjs-aws repository base

dario-piotrowicz:
I've noticed that in a few places the repo base is outdated

this seems generally fine as github just redirects to this repository, but it's not really that polished and can break tools such as `pkg.pr.new` that can rely on the package.json's repository to be correct (see https://github.com/opennextjs/opennextjs-aws/pull/536#issuecomment-2403509138, the PR package there doesn't work properly because of the incorrect repo in the package.json)

### DIFF
--- a/docs/pages/index.mdx
+++ b/docs/pages/index.mdx
@@ -41,7 +41,7 @@ We need your help keeping it up to date and feature complete. Make sure to [**jo
 
 ## Features
 
-OpenNext aims to support all Next.js 14 features. Some features are work in progress. Please open a [new issue](https://github.com/sst/open-next/issues/new) to let us know!
+OpenNext aims to support all Next.js 14 features. Some features are work in progress. Please open a [new issue](https://github.com/opennextjs/opennextjs-aws/issues/new) to let us know!
 
 - [x] App & Pages Router
 - [x] API routes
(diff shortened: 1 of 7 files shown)


conico974 (approved):
LGTM Thanks.
There is a bunch of work that needs to be done in this repo to make it cleaner.
I'll create a bunch of issue and list everything that should be done

conico974:
@dario-piotrowicz Looks like it is still broken.
The url set by the pkg-pr-new bot is not the correct one here, the one that works is this one : https://pkg.pr.new/opennextjs/opennextjs-aws/@opennextjs/aws@3b0bdd3

Weird thing is that it use the correct one in stackblitz

dario-piotrowicz:
> @dario-piotrowicz Looks like it is still broken. The url set by the pkg-pr-new bot is not the correct one here, the one that works is this one : https://pkg.pr.new/opennextjs/opennextjs-aws/@opennextjs/aws@3b0bdd3
> 
> Weird thing is that it use the correct one in stackblitz

ok, let me see....

dario-piotrowicz:
@conico974 oops.... it looks like I missed a spot! 😓  https://github.com/conico974/open-next/blob/185a4739607b0418b650ab077aaea567988e53bf/packages/open-next/package.json#L64

let me open a quick PR for that! sorry! 😓

PR #542, opened 2024-10-10
Basic PPR

conico974:
This PR add basic support for PPR.

This is the PPR built into next standalone, this mean that nothing can be served from the CDN (except the prefetched rsc).

The html are served from the server every time

### DIFF
--- a/packages/open-next/src/build.ts
+++ b/packages/open-next/src/build.ts
@@ -523,8 +523,12 @@ async function createCacheAssets(monorepoRoot: string) {
     () => true,
     (filepath) => {
       const ext = path.extname(filepath);
-      const newFilePath =
+      let newFilePath =
         ext !== "" ? filepath.replace(ext, ".cache") : `${filepath}.cache`;
+      // Handle prefetch cache files for partial prerendering
+      if (newFilePath.endsWith(".prefetch.cache")) {
+        newFilePath = newFilePath.replace(".prefetch.cache", ".cache");
+      }
       switch (ext) {
         case ".meta":
         case ".html":
--- a/.changeset/angry-meals-suffer.md
+++ b/.changeset/angry-meals-suffer.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Basic support for PPR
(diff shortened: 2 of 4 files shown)


packages/open-next/src/build.ts
```diff
@@ -523,8 +523,12 @@ async function createCacheAssets(monorepoRoot: string) {
     () => true,
     (filepath) => {
       const ext = path.extname(filepath);
-      const newFilePath =
+      let newFilePath =
         ext !== "" ? filepath.replace(ext, ".cache") : `${filepath}.cache`;
+      // Handle prefetch cache files for partial prerendering
+      if (newFilePath.endsWith(".prefetch.cache")) {
+        newFilePath = newFilePath.replace(".prefetch.cache", ".cache");
```
khuezy, line 530: This is very unlikely, but is it possible someone can have a `/apps/.prefetch/page.tsx` route. This would cause the actual route path to get replaced. Probably doesn't need to be handled but just thought I'd bring up the possibility 😄

khuezy (approved):
LGTM. Is `ppr` that easy to enable? Even if it's just a basic version.

packages/open-next/src/build.ts
conico974: I had this exact same thought and came to the same conclusion as well. No need to worry about that i think.

And if they use PPR on that route, it would work

conico974:
Yeah we had already everything in place, so it was very easy (I thought it would be a little harder though, but not by much)

The thing is, it's probably not worth it in a lot of cases. The S3 latency could make it totally worthless depending on what's prerendered. And you could end up paying a lot more, 1 S3 GET for every request instead of a few ms of a running lambda

khuezy:
Hmm shouldn't those static assets be generated w/ `cache-control` headers, something like `s-maxage=1yearish, immutable`. I would assume those static assets hit S3 only once and then stored at the CDN.

khuezy:
I must not fully understand the PPR architecture. I assumed the static shell parts are determined at build time. Then when the client makes a request, it streams back the shell that points to the static asset, then continues w/ the suspense boundary stuff.

I haven't dived into the source code to figure this out yet though.

conico974:
That's basically that. 
The only thing is that it does that in a single stream, and they don't provide us with only the `resume` endpoint (unless using minimal mode). 
And to add to that, the first html shell is not always fully generated at build time, sometimes it is generated at runtime from some RSC.

PR #543, opened 2024-10-10
update repositry details in `@opennextjs/aws` package.json

dario-piotrowicz:


dario-piotrowicz:
@conico974 sorry my bad! now it should be all good 😓

dario-piotrowicz:
> I think we need to wait for the next release for it to work well

ahhhh! ok good find! 👍 

yeah I guess it's probably that

conico974 (approved):
That's fine, we'll wait for the next release.
In the meantime we just have to look into stackblitz to get the correct url.

Thanks for that

packages/open-next/package.json
```diff
@@ -61,7 +61,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "git@github.com:opennextjs/opennextjs-aws.git",
+    "url": "git+https://github.com:opennextjs/opennextjs-aws.git",
```
conico974, line 64: @dario-piotrowicz Wait is this correct actually ? 
The one in cloudflare is just this `"url": "https://github.com/opennextjs/opennextjs-cloudflare.git"`
Maybe that's why it doesn't work

dario-piotrowicz: mh... `git+` should be correct as that what's included in the npm docs: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#repository
🤔 

I'm happy to give the http version a go too, I think that they're likely both valid

dario-piotrowicz: @conico974 yeah the longer version is not helping, I do think we just need to wait for the next release 😕

conico974: It was worth a try.
Ok i'll merge this then, we can revisit this if it still don't work on next release

PR #544, opened 2024-10-10
fix all the various examples eslint warnings

dario-piotrowicz:
I've noticed that every PR is getting eslint warnings coming from the `examples` directory, so I figured I could open a quick PR to address them 🙂 

For example: https://github.com/opennextjs/opennextjs-aws/pull/543/files

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -32,6 +32,14 @@ module.exports = {
 
     "@typescript-eslint/no-non-null-assertion": "warn",
   },
+  overrides: [
+    {
+      files: ["examples/**/*"],
+      rules: {
+        "unused-imports/no-unused-vars": "off",
+      },
+    },
+  ],
   parserOptions: {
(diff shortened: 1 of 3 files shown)


.eslintrc.cjs
```diff
@@ -32,6 +32,14 @@ module.exports = {
 
     "@typescript-eslint/no-non-null-assertion": "warn",
   },
+  overrides: [
+    {
+      files: ["examples/**/*"],
+      rules: {
+        "unused-imports/no-unused-vars": "off",
```
dario-piotrowicz, line 39: I am guessing that in the `examples` directory you might want to keep unused vars? I'm happy to just fix the actual issues instead of disabling the rule if that's preferred 🙂

conico974: Yeah i guess that's fine. 
Everything in `examples` is there for e2e and as example, it make sense that we'll have unused variable sometimes.

PR #545, opened 2024-10-11
fix all the remaining eslint warnings

dario-piotrowicz:
continuation from #544 🙂 

Some changes might be controversial I am happy to adjust them any way you prefer, I just applied what I thought more sensible and/or my personal preference 🙂

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -15,7 +15,7 @@ module.exports = {
     ],
     "sonarjs/elseif-without-else": "warn",
     "sonarjs/no-duplicate-string": "warn",
-    "sonarjs/cognitive-complexity": "warn",
+    "sonarjs/cognitive-complexity": ["warn", 35],
 
     // We add some typescript rules - The recommended rules breaks too much stuff
     // TODO: We should add more rules, especially around typescript types
@@ -29,8 +29,6 @@ module.exports = {
     ],
 
     "@typescript-eslint/unbound-method": "error",
-
-    "@typescript-eslint/no-non-null-assertion": "warn",
   },
   overrides: [
     {
--- a/packages/open-next/src/build/generateOutput.ts
+++ b/packages/open-next/src/build/generateOutput.ts
@@ -92,6 +92,8 @@ interface OpenNextOutput {
   };
 }
 
+const indexHandler = "index.hander";
+
 async function canStream(opts: FunctionOptions) {
   if (!opts.override?.wrapper) {
     return false;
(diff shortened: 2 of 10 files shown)


.eslintrc.cjs
```diff
@@ -29,8 +29,6 @@ module.exports = {
     ],
 
     "@typescript-eslint/unbound-method": "error",
-
-    "@typescript-eslint/no-non-null-assertion": "warn",
```
dario-piotrowicz, line 33: I am personally not a fan of this one 😅 

if someone uses a non-null assertion I think that's generally for a good reason so having to add an eslint disabling comment for it feels very unnecessary/annoying to me 🤷

.eslintrc.cjs
```diff
@@ -15,7 +15,7 @@ module.exports = {
     ],
     "sonarjs/elseif-without-else": "warn",
     "sonarjs/no-duplicate-string": "warn",
-    "sonarjs/cognitive-complexity": "warn",
+    "sonarjs/cognitive-complexity": ["warn", 35],
```
dario-piotrowicz, line 18: `35` allows most functions in the codebase, I could just add disabling comment to each one instead if preferred (there were around 10-ish or something I think)

I did not attempt on splitting the functions as I am not really familiar with the codebase here (but I am happy to give it a go if you want 🙂)

.eslintrc.cjs
conico974: Yeah I'm fine with this
We just need to verify on every PR to make sure it's not used unwisely

.eslintrc.cjs
conico974: I agree.
I need to take a closer look at the PR later today, I remember that there was a reason that it was this order. Might be old code

conico974 (approved):
Sorry it was old code, there was a reason for the order back when we used multiple file for cache, but now that we have a single file it doesn't matter

conico974:
https://github.com/opennextjs/opennextjs-aws/actions/runs/11312902537/job/31460954988

packages/open-next/src/build/generateOutput.ts
```diff
@@ -92,6 +92,8 @@ interface OpenNextOutput {
   };
 }
 
+const indexHandler = "index.hander";
```
conico974, line 95: There was a typo here

dario-piotrowicz: sorry 😓

conico974: No problem

sommeeeer: > sorry 😓

Typos happens to everyone. Your PR made the code more DRY and easier to read. Great work ⭐️

Issue #538, opened 2024-10-10
[Chores] Cleanup the repo


conico974:
We should take advantage of the new repo to cleanup a bunch of things in this repo to make it cleaner.

- [x] We can remove everything related to docs here. The new doc is on another repo anyway
- [x] We should either totally redo the `example` or just remove it and rely entirely on the different `examples` repo
- [x] Globals are defined a bit everywhere right now. We should at least define all the type in a single place with some comment to know where and when they are applied.
- [x] We could move to biome instead of eslint. It's a lot faster and it is a single dependency
- [x] Figure out the issue with the lockfile. Using pnpm 9 (with the new lockfile version) breaks almost everything in the repo
- [x] We  could also add [syncpack](https://github.com/JamieMason/syncpack/) in this repo, it would help a lot with different dependencies and would allow us to lint dep, so that we mostly use fixed version in `package.json`
- [x] Update the README

PR #546, opened 2024-10-12
chore: remove the docs

sommeeeer:
#538

this PR removes the `/docs` entirely as it is now in a seperate [repo](https://github.com/opennextjs/docs)

ran `npx pnpm@8 i` after that to update the lockfile.


conico974:
Before doing that, i think ideally we'd want to do it in 2 steps.
First we update the actual doc site (it's still out there, it just does a redirect immediately) to remove everything besides the index page. Maybe add some kind of message that the docs site has moved in case redirection fails for some reason.
And then we can remove the repo (Or maybe even keep it with a README that explains that the doc moved)

Issue #539, opened 2024-10-10
[Test] Improve test


conico974:
Right now only e2e test work and not everything is properly covered (It's almost impossible to cover everything anyway with e2e given the number of possibilities)

- [ ] Fix the unit test package and integrate it in GH actions
- [ ] We could run a lot more e2e test by running them locally on node
- [ ] Try to cover as much as possible the routing part with unit test

If someone has other ideas on how to improve testing, i'm all ears

socsieng:
I'm keen to contribute

conico974:
@socsieng The easiest one to start with would be the unit test https://github.com/opennextjs/opennextjs-aws/tree/main/packages/tests-unit/tests
All the tests here are broken and needs to be ported for OpenNext v3. It's probably easier to just start over.
What we should cover in unit test are all the converters (maybe not the edge one for now), and most of the things in `core/routing` (not the middleware) that would already be a great start.

You need to use pnpm 8 though (pnpm 9 cause some trouble)

socsieng:
Thoughts about colocating the tests with the code instead of in a separate package?

conico974:
Not a big fan at all personally. Separate package makes things a bit cleaner

socsieng:
I've had a go at just fixing the existing tests: https://github.com/opennextjs/opennextjs-aws/pull/551

Wanted to confirm that this is what you wanted before proceeding.

PR #551, opened 2024-10-14
Fix broken tests

socsieng:
Initial PR to try and fix broken tests.

For: https://github.com/opennextjs/opennextjs-aws/issues/539


packages/tests-unit/tests/adapter.utils.test.ts
conico974, line 1: I'd rather have one test file per file. It makes it easier to know what's covered and what's not.
So it would be split in 2 `core/routing/util.test.ts` and `http/util.test.ts`

conico974 (commented):
As for the test failing it's probably some kind of config in vitest itself.
Does it work locally ?

socsieng:
> As for the test failing it's probably some kind of config in vitest itself. Does it work locally ?

It was working locally. It is failing now after `git clean`. Looking into it.

socsieng:
Tests fixed. Let me know if you're happy to proceed. I can continue to flesh out additional tests in this PR or separate pull requests if you prefer.

conico974 (changes requested):
Separate PR will be easier to review.
If you could just move the `cache.test.ts` following the same logic as the other files.
And delete `example.test.ts`

conico974 (approved):
LGTM
Thanks !!!

PR #552, opened 2024-10-15
fix: remove duplicated code

vicb:


packages/open-next/src/wrappers/aws-lambda-streaming.ts
```diff
@@ -76,8 +76,6 @@ const handler: WrapperHandler = async (handler, converter) =>
 
       const streamCreator: StreamCreator = {
         writeHeaders: (_prelude) => {
-          _prelude.headers["content-encoding"] = contentEncoding;
```
conico974, line 79: Is it duplicated ?
I think we need to set it here, the compression is done by the wrapper itself

vicb: image

Check l84

conico974: :facepalm: My bad

conico974 (approved):
Thanks

PR #553, opened 2024-10-15
Make function dependencies explicit

vicb:
We (opennext-cloudflare) plan to reuse code from opennext-aws.

For that we would like to extract code out of `build.ts` so that we can call those functions from our own `build.ts`.

As a first step, this PR explicit function dependencies.

The PR contains multiple commits to ease the review.

### DIFF
--- a/packages/open-next/src/build.ts
+++ b/packages/open-next/src/build.ts
@@ -77,27 +77,27 @@ export async function build(
 
   // Build Next.js app
   printHeader("Building Next.js app");
-  setStandaloneBuildMode(config, options);
-  buildNextjsApp(config, options);
+  setStandaloneBuildMode(options);
+  buildNextjsApp(options);
 
   // Generate deployable bundle
   printHeader("Generating bundle");
   initOutputDir(tempDir, options);
 
   // Compile cache.ts
-  compileCache(config, options);
+  compileCache(options);
 
   // Compile middleware
-  await createMiddleware(config, options);
+  await createMiddleware(options);
 
   createStaticAssets(options);
-  await createCacheAssets(config, options);
+  await createCacheAssets(options);
 
-  await createServerBundle(config, options);
-  await createRevalidationBundle(config, options);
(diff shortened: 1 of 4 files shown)


packages/open-next/src/build/helper.ts
```diff
@@ -17,37 +17,69 @@ const __dirname = url.fileURLToPath(new URL(".", import.meta.url));
...
+        logger.info("Monorepo detected at", currentPath);
+      }
+      return { root: currentPath, packager: found.packager };
+    }
+    currentPath = path.dirname(currentPath);
+  }
+
+  // note: a lock file (package-lock.json, yarn.lock, or pnpm-lock.yaml) is
+  //       not found in the app's directory or any of its parent directories.
+  //       We are going to assume that the app is not part of a monorepo.
```
conico974, line 79: We may want to add at least a warning here. I don't think this is supposed to happen

packages/open-next/src/build.ts
```diff
@@ -198,47 +167,48 @@ function printHeader(header: string) {
...
 
-function printOpenNextVersion() {
-  const { openNextVersion } = options;
-  logger.info(`OpenNext v${openNextVersion}`);
+function printOpenNextVersion(options: BuildOptions) {
+  logger.info(`OpenNext v${options.openNextVersion}`);
 }
 
-function initOutputDir(tempDir: string) {
+function initOutputDir(srcTempDir: string, options: BuildOptions) {
```
conico974, line 178: I guess we could move `srcTempDir` inside options as well

conico974 (approved):
Just 2 little comment, but other than that, it looks good to me

packages/open-next/src/build/helper.ts
vicb: I'll address in a follow up PR

packages/open-next/src/build.ts
vicb: I'll address in a follow up PR

PR #554, opened 2024-10-15
Minor refactoring

vicb:


packages/open-next/src/build.ts
```diff
@@ -175,33 +167,26 @@ function printOpenNextVersion(options: BuildOptions) {
...
   fs.rmSync(options.outputDir, { recursive: true, force: true });
-  const destTempDir = options.tempDir;
-  fs.mkdirSync(destTempDir, { recursive: true });
-  fs.writeFileSync(
-    path.join(destTempDir, "open-next.config.mjs"),
-    openNextConfig,
+  const { buildDir } = options;
+  fs.mkdirSync(buildDir, { recursive: true });
+
+  fs.copyFileSync(
```
vicb, line 178: We can actually simplify by copying the whole temp folder. I'll amend the PR later today

vicb:
PR updated, ready to review

packages/open-next/src/build.ts
```diff
@@ -47,9 +47,11 @@ export async function build(
   showWindowsWarning();
 
   // Load open-next.config.ts
-  const tempDir = initTempDir();
+  const tempConfigDir = fs.mkdtempSync(
```
vicb, line 50: maybe we can rename this to `tempBuildDir` because it will become the `buildDir` (move)

conico974: Yeah it make sense

packages/open-next/src/build.ts
```diff
@@ -175,34 +169,15 @@ function printOpenNextVersion(options: BuildOptions) {
...
-  if (openNextConfigEdge) {
-    fs.writeFileSync(
-      path.join(destTempDir, "open-next.config.edge.mjs"),
-      openNextConfigEdge,
-    );
-  }
+  const { buildDir } = options;
+  fs.mkdirSync(buildDir, { recursive: true });
+  fs.renameSync(options.tempConfigDir, buildDir);
+  process.exit(0);
```
conico974, line 180: Why the `process.exit` here ?

vicb: Oops, because I wanted to stop here while debugging.
I'll remove!

packages/open-next/src/build.ts
```diff
@@ -175,34 +169,14 @@ function printOpenNextVersion(options: BuildOptions) {
...
-  );
-  if (openNextConfigEdge) {
-    fs.writeFileSync(
-      path.join(destTempDir, "open-next.config.edge.mjs"),
-      openNextConfigEdge,
-    );
-  }
+  const { buildDir } = options;
+  fs.mkdirSync(buildDir, { recursive: true });
+  fs.renameSync(options.tempBuildDir, buildDir);
```
conico974, line 179: We cannot use renameSync here.
Just tested on my computer and this crashed with this error : 
`EXDEV: cross-device link not permitted, rename '/tmp/.open-next-tempaOh2Rc' -> '/mnt/ssd2/projects/next-app-dir/.open-next/.build'`
We probably need to copy

conico974 (changes requested):

packages/open-next/src/build.ts
vicb: Oh different devices...
We could create the temp build folder in the app folder instead of the OS temp folder and that should solve the problem.
Thanks for catching this. 
I'll update tomorrow.

vicb: I have updated to use use `cpSync`.
I picked `cpSync` over creating a folder in the current directory not to leave a folder in the current directory when it's not a next app.

conico974 (approved):
LGTM, we have to fix e2e test though before merging. We need to update e2e example apps to next 15

conico974 (approved):
Perfect it should fix e2e. Thanks

PR #555, opened 2024-10-15
fix: proxy resolves once the full stream has been completed.

JackParn33:
Currently if a request with the Accept-Encoding head is proxied e.g. with NextJs rewrites the response resolves before the full response has been processed.


khuezy (commented):
Hey thanks for looking into this!
Will this have any side effect with requests that don't have the `Accept-Encoding` headers?

conico974 (approved):
LGTM Thanks

PR #556, opened 2024-10-15
Add tests for converters

socsieng:


packages/tests-unit/tests/converters/aws-apigw-v1.test.ts
```diff
@@ -0,0 +1,264 @@
...
+      multiValueHeaders: {},
+      httpMethod: "POST",
+      isBase64Encoded: false,
+      path: "/",
+      pathParameters: null,
+      queryStringParameters: {
+        test: "test",
+      },
+      multiValueQueryStringParameters: {
+        test: ["test"],
```
khuezy, line 153: What happens when the same `key (test)` has different values? Do we want to merge the two? Is that even possible?
eg:
```
      queryStringParameters: {
        test: "test",
      },
      multiValueQueryStringParameters: {
        test: ["test-multi"],
      }
```

khuezy (approved):
Thanks so much for doing this! When ready, we should add this to the github workflow to run the unit test on PRs (`synchronize`)

conico974:
> When ready, we should add this to the github workflow to run the unit test on PRs (synchronize)
It already runs on validate right now

packages/tests-unit/tests/converters/aws-apigw-v1.test.ts
conico974: I guess it shouldn't be possible, hopefully aws has some kind of validation on this to avoid this kind of case.

khuezy:
Ah nice, I didn't realize.

conico974 (approved):
That's very nice thank you.
LGTM

PR #562, opened 2024-10-16
Split build.ts

vicb:
This will make it easier to reuse code (i.e. for opennext-cloudflare).

This PR is mainly moving code around.

I'll add inline comments for other differences to make the review easier.


packages/open-next/src/build/compileCache.ts
```diff
@@ -0,0 +1,45 @@
...
+import * as buildHelper from "./helper.js";
+
+/**
+ * Compiles the cache adapter.
+ *
+ * @param options Build options.
+ * @param format Output format.
+ * @returns The path to the compiled file.
+ */
+export function compileCache(
```
vicb, line 12: This is an added function.

It calls the two existing function for Node and edge if needed.

The edge test has been moved to this function.

packages/open-next/src/build/compileConfig.ts
```diff
@@ -1,20 +1,74 @@
...
+ * Compiles the OpenNext configuration.
+ *
+ * The configuration is always compiled for Node.js and for the edge only if needed.
+ *
+ * @param baseDir Directory where to look for the configuration.
+ * @param openNextConfigPath Override the default configuration when provided. Relative to baseDir.
+ * @param nodeExternals Externals for the Node.js compilation.
+ * @return The configuration and the build directory.
+ */
+export async function compileOpenNextConfig(
```
vicb, line 21: This is an added function.

It calls the two existing function for Node and edge if needed.

The edge test has been moved to this function.

packages/open-next/src/build.ts
```diff
@@ -44,36 +25,25 @@ export async function build(
   openNextConfigPath?: string,
   nodeExternals?: string,
 ) {
-  showWindowsWarning();
+  showWarningOnWindows();
+
+  const baseDir = process.cwd();
+  const distDir = url.fileURLToPath(new URL(".", import.meta.url));
```
conico974, line 31: Maybe call it `openNextDistDir` as in options. distDir is a bit confusing

packages/open-next/src/build/createAssets.ts
```diff
@@ -0,0 +1,261 @@
...
+                ),
+              },
+              revalidatedAt: { N: "1" },
+            });
+          });
+        },
+      );
+    }
+
+    if (metaFiles.length > 0) {
```
conico974, line 219: We could probably split everything below as well. We could create a `compileTagCacheProvider` function. 
It would get rid of the eslint warning as well

packages/open-next/src/build/compileConfig.ts
```diff
@@ -1,20 +1,74 @@
...
+  const usesEdgeRuntime =
+    config.middleware?.external ||
+    Object.values(config.functions || {}).some((fn) => fn.runtime === "edge");
+  if (!usesEdgeRuntime) {
+    logger.debug(
+      "No edge runtime found in the open-next.config.ts. Using default config.",
+    );
+    //Nothing to do here
+  } else {
+    compileOpenNextConfigEdge(baseDir, buildDir, config.edgeExternals ?? []);
```
conico974, line 61: Should be `sourcePath` here, not `baseDir`

conico974 (changes requested):
Just a few comment

packages/open-next/src/build/compileConfig.ts
vicb: Good catch!

packages/open-next/src/build/createAssets.ts
vicb: I've added a TODO.
I prefer this PR to be mostly code move and we can iterate in follow-up PRs.

conico974 (approved):
LGTM
Thanks

vicb:
Thanks for your help @conico974

PR #564, opened 2024-10-17
Add unit tests for route handler

socsieng:
Still a WIP progress while I try to familiarise myself with expected behaviour, but wanted some early feedback on the approach for now.


packages/tests-unit/tests/core/routingHandler.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  const event = createEvent({
+    url: "/_next/data/lnRb3bjsx2nspyrLXPfpa/test",
+  });
+
+  const response = (await routingHandler(event)) as InternalResult;
+
+  expect(response.statusCode).not.toEqual(404);
+});
+
+it("should redirect trailing slashes", async () => {
```
socsieng, line 85: I plan on being able to provide mock implementations for things like `redirects` to test different redirect behaviours.

conico974 (commented):
Before you go any further, i'll give my advice on this.
We shouldn't unit test big component like this routingHandler file, we should rather test the little function used inside like `addNextConfigHeaders` or `fixDataPage`

For me we should keep as little mock as possible, because otherwise at one point you risk end up testing against your mock instead. 
And we also risk having to provide thousands of mock to cover all the test we need.

We could maybe reconsider this later and add integration test, but the priority right now is these small function

socsieng:
Closing this PR and replacing with https://github.com/opennextjs/opennextjs-aws/pull/568

PR #566, opened 2024-10-17
fix: create the directory before copying the config

vicb:
Sorry for breaking e2e :(


conico974 (approved):
Don't worry, that's why we have e2e

PR #568, opened 2024-10-17
Add unit tests for route matching

socsieng:
I haven't included tests for `handleFallbackFalse` because I don't feel like I understand the expected behaviour well enough.


conico974 (approved):
LGTM 
Thanks a lot for doing that

PR #569, opened 2024-10-18
update readme

sommeeeer:
This PR is gonna update the README as stated as requested in #538.

Let me know if there is more updates needed. What about Next 15? Should we say something about it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ To run `OpenNext` locally:
 
 OpenNext provide you with a warmer function that can be used to reduce cold start.
 
-On Lambda, there is multiple scenarios where a lambda will trigger a cold start even if you have some warmed instance. For example if you have more request than warm instance you'll get a cold start. Also NextJs lazy load the routes, so even if you hit a warm instance, this specific route might not have been loaded yet.
+On Lambda, there are multiple scenarios where a lambda will trigger a cold start even if you have some warmed instance. For example if you have more request than warm instance you'll get a cold start. Also NextJs lazy load the routes, so even if you hit a warm instance, this specific route might not have been loaded yet.
 
 ## Acknowledgements
 
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ OpenNext aims to support all Next.js 14 features. Some features are work in prog
 - [x] Image optimization
 - [x] [NextAuth.js](https://next-auth.js.org)
 - [x] Running at edge
-- [x] Almost no coldstart
+- [*] [Almost no coldstart](#coldstart)
 
 ## Who is using OpenNext?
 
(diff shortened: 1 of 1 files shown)


conico974 (changes requested):
We should also update the docs link at the top. It should be way bigger.
Far too many people don't go to the docs.
We could add a message here, before going on discord or creating an issue check the docs ( there is even a search bar ) or something similar

README.md
```diff
@@ -31,14 +31,15 @@ OpenNext aims to support all Next.js 14 features. Some features are work in prog
...
+- [x] Server actions
 - [x] Image optimization
 - [x] [NextAuth.js](https://next-auth.js.org)
 - [x] Running at edge
 - [x] No cold start
 
 ## Who is using OpenNext?
 
-[Gymshark UK](https://uk.gymshark.com), [Udacity](https://engineering.udacity.com/deploying-next-js-on-the-edge-with-sst-is-sst-the-game-changer-its-claimed-to-be-1f05a0abc27c), [TUDN](https://www.tudn.com), [NHS England](https://github.com/nhs-england-tools/terraform-aws-opennext)
+[Gymshark UK](https://uk.gymshark.com), [Udacity](https://engineering.udacity.com/deploying-next-js-on-the-edge-with-sst-is-sst-the-game-changer-its-claimed-to-be-1f05a0abc27c), [TUDN](https://www.tudn.com), [NHS England](https://github.com/nhs-england-tools/terraform-aws-opennext) & [Nike](https://about.nike.com)
```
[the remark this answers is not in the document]
sommeeeer: > Did they give their explicit permission ?

nope

README.md
```diff
@@ -31,17 +32,16 @@ OpenNext aims to support all Next.js 14 features. Some features are work in prog
 - [x] Server-side rendering (SSR)
 - [x] Incremental static regeneration (ISR)
 - [x] Middleware
+- [x] Server actions
 - [x] Image optimization
 - [x] [NextAuth.js](https://next-auth.js.org)
 - [x] Running at edge
-- [x] No cold start
+- [x] Almost no coldstart
```
khuezy, line 39: We should add an asterisk to provide more details.
eg, on Lambda, there will always be a scenario where a Lambda is created that is > warmed instance, so that request will be cold. For Nextjs, not all routes are warmed from the warm event.

README.md
```diff
@@ -78,6 +85,12 @@ To run `OpenNext` locally:
    path/to/open-next/packages/open-next/dist/index.js build
    ```
 
+### Coldstart
+
+OpenNext provide you with a warmer function that can be used to reduce cold start.
+
+On Lambda, there is multiple scenario where a lambda will trigger a cold start even if you have some warmed instance. For example if you have more request than warm instance you'll get a cold start. Also NextJs lazy load the routes, so even if you hit a warm instance, this specific route might not have been loaded yet.
```
khuezy: Thanks for the details!

Just a minor correction on the sentence: `there is multiple scenario` should be `there are multiple scenarios`

sommeeeer (commented):
Should this maybe be: `For example if you have more requests than warm instances you'll get a cold start.`

khuezy (approved):

conico974 (approved):
LGTM Thanks

PR #574, opened 2024-10-18
New option to install packages

### LINKED ISSUE #571: [Feature] Add an install option in config
Every function (default, image optimization, middleware ... ) should have an install option to install native dependencies for the correct arch. This would also allow to install sharp on the default/splitted functions if needed.

Something like that : 
```ts
install : {
 packages: ['sharp', ...],
 arch: 'arm'|'x64',
}
```

conico974:
This PR fix #571 

#### How to use

In any of the function in `open-next.config.ts` you can now add a new option that will install dependencies that may not be included (For example sharp): 
```ts
install : {
  packages: ["sharp@0.33"],
  arch: "x64", // or arm64
  nodeVersion: "18", // or 20 or 22
  libc: "glibc" // or musl
}
```
Install itself is optional, and everything in it is optional besides packages

### DIFF
--- a/packages/open-next/src/build/installDeps.ts
+++ b/packages/open-next/src/build/installDeps.ts
@@ -1,7 +1,7 @@
 import { execSync } from "child_process";
-import * as fs from "fs";
-import * as os from "os";
-import * as path from "path";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
 import { InstallOptions } from "types/open-next";
 
 import logger from "../logger.js";
--- a/packages/open-next/src/build/installDeps.ts
+++ b/packages/open-next/src/build/installDeps.ts
@@ -1,7 +1,8 @@
-import { execSync } from "child_process";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
+
+import { execSync } from "child_process";
 import { InstallOptions } from "types/open-next";
 
 import logger from "../logger.js";
--- a/packages/open-next/src/build/installDeps.ts
+++ b/packages/open-next/src/build/installDeps.ts
@@ -1,0 +1,50 @@
+import { execSync } from "child_process";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import { InstallOptions } from "types/open-next";
+
+import logger from "../logger.js";
+
+export function installDependencies(
+  outputDir: string,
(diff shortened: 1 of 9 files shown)


packages/open-next/src/build/installDeps.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    const name = outputDir.split("/").pop();
+    // First we create a tempDir
+    const tempInstallDir = fs.mkdtempSync(
+      path.join(os.tmpdir(), `open-next-install-${name}`),
+    );
+    logger.info(`Installing dependencies for ${name}...`);
+    // We then need to run install in the tempDir
+    // We don't install in the output dir directly because it could contain a package.json, and npm would then try to reinstall not complete deps from tracing the files
+    const installCommand = `npm install --arch=${installOptions.arch ?? "arm64"} --platform=linux --target=${installOptions.nodeVersion ?? "18"} --libc=${installOptions.libc ?? "glibc"} ${installOptions.packages.join(" ")}`;
```
vicb, line 25: Do we want to use whatever packager was used for Next as it has been retrieved anyway?

conico974: Maybe, but we'll need to figure out how to pass all these options. 
I don't think they support all of them, or there is no docs. If some other package manager support it we could use them and fallback to npm when not possible

vicb: We could pass the `buildOptions` and the function name to `installDeps`.
It would mean adding the `outputPath` and the packager there.
But that's probably a nice to have for later.

conico974: I thought about using `buildOptions` and `name`, but i think it would just overcomplicate things.
It could be a splitted functions (which are under functions in the config) and server functions are inside a different folder than the rest.
We could still pass `buildOptions` to get the packager

packages/open-next/src/build/installDeps.ts
```diff
@@ -0,0 +1,54 @@
+import { execSync } from "child_process";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
```
vicb, line 4:
```suggestion
import fs from "node:fs";
import os from "node:os";
import path from "node:path";
```

vicb (approved):

PR #575, opened 2024-10-18
Add protobuf to common binary formats

zdenham:
Came across an issue where `open-next` seems to be encoding `application/x-protobuf` in utf8 corrupting the binary.

### DIFF
--- a/.changeset/honest-keys-protect.md
+++ b/.changeset/honest-keys-protect.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Add protobuf to common binary formats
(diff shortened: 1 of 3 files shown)


khuezy (commented):
Thanks for the PR! Should we consolidate both objects? I think they were duplicated.

sommeeeer:
> Thanks for the PR! Should we consolidate both objects? I think they were duplicated.

We can safely remove the one in `/utils` (its only used in a unit test), however the one in `/adapters` doesn't have `audio/wavaudio/webm`. we should probably add that, and also change the import in the unit test. then we can safely remove the `/utils` one. 

However after thinking about this, we should probably do the opposite? As this is a util function, and a better fit for it is the `/utils` folder.

conico974 (approved):
LGTM Thanks
We'll figure the duplicate file later

PR #576, opened 2024-10-19
Add unit tests for handling middleware

socsieng:


packages/open-next/src/core/routing/middleware.ts
```diff
@@ -53,8 +59,7 @@ export async function handleMiddleware(
   const url = initialUrl.toString();
   // console.log("url", url, normalizedPath);
 
-  // @ts-expect-error - This is bundled
-  const middleware = await import("./middleware.mjs");
+  const middleware = await middlewareLoader("./middleware.mjs");
```
conico974, line 62: You cannot do that, when the middleware is external we use an esbuild plugin to change the resolve.
This will break it

packages/open-next/src/core/routing/middleware.ts
```diff
@@ -28,6 +28,11 @@ type MiddlewareOutputEvent = InternalEvent & {
   externalRewrite?: boolean;
 };
 
+type MiddlewareLoader = (path: string) => Promise<any>;
+function defaultMiddlewareLoader(path: string) {
+  return import(path);
```
conico974, line 33: It needs to be hardcoded here.

packages/open-next/src/core/routing/middleware.ts
```diff
@@ -28,6 +28,11 @@ type MiddlewareOutputEvent = InternalEvent & {
   externalRewrite?: boolean;
 };
 
+type MiddlewareLoader = (path: string) => Promise<any>;
```
conico974, line 31: Properly typing the return would be great

conico974 (changes requested):
I have mixed feelings about this one (kind of the same issue than with the `routingHandler` )
It's fine for now i guess, we can change it later if needs be
We could add a note at the top to remind us about that

socsieng:
> It's fine for now i guess, we can change it later if needs be We could add a note at the top to remind us about that

What kind of note, and where did you have in mind?

packages/open-next/src/core/routing/middleware.ts
socsieng: Middleware returns a [`NextResponse`](https://nextjs.org/docs/app/building-your-application/routing/middleware#nextresponse). What are your thoughts on tests taking a dependency on next?

conico974: That's for end user, the middleware here just returns a `Response`.

That's one of the reason i don't like mocking this kind of stuff. It's user code, compiled a first time by Next, recompiled by us to be usable here. Properly mocking this is virtually impossible.

conico974:
> > It's fine for now i guess, we can change it later if needs be We could add a note at the top to remind us about that
> 
> What kind of note, and where did you have in mind?

Something at the top of the test file.
Something like : Ideally we'd want to not test this function as a whole, but rather break it into smaller component and test these component independently

packages/open-next/src/core/routing/middleware.ts
socsieng: > Properly mocking this is virtually impossible.

I don't know about this... I think it is reasonable to test how the middleware response should be handled.

conico974 (approved):
LGTM Thanks

PR #577, opened 2024-10-19
Add unit tests for cache interceptor

socsieng:


packages/tests-unit/tests/core/routing/cacheInterceptor.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    const body = await fromReadableStream(result.body);
+    expect(body).toEqual("Hello, world!");
+    expect(result).toEqual(
+      expect.objectContaining({
+        type: "core",
+        statusCode: 200,
+        isBase64Encoded: false,
+      }),
+    );
+    expect(result.headers).toHaveProperty("cache-control");
```
conico974, line 139: We want to test the value of these headers as well (except etag)

packages/tests-unit/tests/core/routing/cacheInterceptor.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    routes: {
+      "/albums": {
+        initialRevalidateSeconds: false,
+        srcRoute: "/albums",
+        dataRoute: "/albums.rsc",
+      },
+      "/revalidate": {
+        initialRevalidateSeconds: 60,
+        srcRoute: "/revalidate",
+        dataRoute: "/revalidate.rsc",
```
conico974, line 21: You set it as if it were an app router page here, but then you use it as a page router page in test below.
It doesn't matter for the test but for consistency it should be correct here

packages/tests-unit/tests/core/routing/cacheInterceptor.test.ts
```diff
@@ -0,0 +1,238 @@
...
+
+  it("should retrieve page router content from cache", async () => {
+    const event = createEvent({
+      url: "/revalidate",
+    });
+    incrementalCache.get.mockResolvedValueOnce({
+      value: {
+        type: "page",
+        html: "Hello, world!",
+      },
```
conico974, line 172: Since this route is ISR, we also want to test other case like when the page is a HIT.
You'll need to add `lastModified` here for it to work, and change the expected result as well.
I'm fine with merging this right now and we'll add this later in a subsequent PR if you want, in this case just add a TODO here

conico974 (approved):
LGTM Thanks.
Next time if you could avoid force pushing into the same commit it would be great. It makes review harder because i have to go through the whole PR everytime

PR #578, opened 2024-10-19
Add unit test coverage reporting

socsieng:
Example output: https://github.com/opennextjs/opennextjs-aws/actions/runs/11415198269#summary-31764988744

image

### DIFF
--- a/.github/actions/test/action.yml
+++ b/.github/actions/test/action.yml
@@ -7,3 +7,9 @@ runs:
         - name: Run tests
           run: pnpm test
           shell: bash
+
+        - name: Report test coverage
+          if: always() 
+          uses:  davelosert/vitest-coverage-report-action@v2
+          with:
+              vite-config-path: ./packages/tests-unit/vitest.config.ts
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -8,6 +8,9 @@ on:
 jobs:
     validate:
         runs-on: ubuntu-latest
+        permissions:
+            contents: read
+            pull-requests: write
         steps:
             - name: Checkout
               uses: actions/checkout@v3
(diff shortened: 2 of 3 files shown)


conico974:
Do we want to add this ?
That's great when you can achieve good coverage but given the code we have, i don't think it make that much sense.
cc @khuezy

conico974:
@socsieng Wait is it only on the action that this appear or on the PR as well ?
If it's on the action only forget my comment

socsieng:
> Do we want to add this ? That's great when you can achieve good coverage but given the code we have, i don't think it make that much sense. cc @khuezy

I thought it could serve as a reminder to include tests for future contributions

conico974:
> > Do we want to add this ? That's great when you can achieve good coverage but given the code we have, i don't think it make that much sense. cc @khuezy
> 
> I thought it could serve as a reminder to include tests for future contributions

That's a good point, but given how some stuff cannot be unit tested i wonder if it won't make people try to write test and us having to ask to remove them and having to explain this everytime.

Or maybe it won't have any effect, the changeset bot as been there for a while and almost no PR contains a changeset :joy: 

Btw i've checked it doesn't work right now because the GH action don't have the permission to write PR

socsieng:
> Btw i've checked it doesn't work right now because the GH action don't have the permission to write PR

I noticed that too. I set the permissions in the workflow file but I'm guessing it's related to my permissions on this repository.

I tested it out on my fork and it works as expected: https://github.com/socsieng/open-next/pull/1#issuecomment-2423790167

It might work for subsequent pull requests after merge... 🤞

khuezy (commented):
Hey sorry I missed the ping here. LGTM Thanks!

The coverage report is nice but we shouldn't get too discouraged by the coverage numbers. 100% coverage is a trap!

PR #579, opened 2024-10-19
Refactor overrides

### LINKED ISSUE #570: Refactor overrides
- [x] Move every possible overrides into the `overrides` folder
- [x] Always provide a dummy override (Very useful for custom overrides)
- [x] Always resolve overrides from here https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/core/resolve.ts
- [x] Update the resolve plugin https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/plugins/resolve.ts

conico974:
Fix #570 
For review you probably want to look commit by commit


packages/open-next/src/overrides/imageLoader/dummy.ts
```diff
@@ -0,0 +1,11 @@
+import { ImageLoader } from "types/open-next";
```
vicb, line 1: Use `import type` - it applies to all the dummy overrides.
I think there is an autofix rule for eslint.

vicb: I think it's https://typescript-eslint.io/rules/consistent-type-imports/

conico974: Nice catch, i'll add it later today in this PR as well as fixing all the missing one

packages/open-next/src/overrides/wrappers/dummy.ts
```diff
@@ -0,0 +1,13 @@
+import type { WrapperHandler } from "types/open-next";
+
+const dummyWrapper: WrapperHandler = async () => {
+  return async () => {
+    return;
+  };
+};
```
vicb, line 7:
```suggestion
const dummyWrapper: WrapperHandler = async () => async () => undefined;
```

vicb (approved):
Looks good.
Maybe move "[added and fixed eslint rules for type import](https://github.com/opennextjs/opennextjs-aws/pull/579/commits/b472368d97e94b5283920aadc1e6b3fb48046fb6)" in a separate PR to remove the noise from here?

conico974:
> Looks good. Maybe move "[added and fixed eslint rules for type import](https://github.com/opennextjs/opennextjs-aws/pull/579/commits/b472368d97e94b5283920aadc1e6b3fb48046fb6)" in a separate PR to remove the noise from here?

@vicb That's probably a good idea yeah
I think this one ( along with #574 ) should be considered minor because they could break existing code. This one for custom overrides and the other one because it change the way to modify sharp install

conico974:
I've merged this one, but i'll stop the e2e test as they are broken at the moment.
I've run the e2e test on my account and it works

PR #580, opened 2024-10-19
fix: readme features list 

sommeeeer:
This PR is to fix the features list. The `Almost no coldstart` element.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ OpenNext aims to support all Next.js 14 features. Some features are work in prog
 - [x] Image optimization
 - [x] [NextAuth.js](https://next-auth.js.org)
 - [x] Running at edge
-- [*] [Almost no coldstart](#coldstart)
+- [x] [Almost no coldstart (*)](#coldstart)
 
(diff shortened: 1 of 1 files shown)


sommeeeer:
Had to use `<h1>` instead of inline `style=` on the HTML element cause [GitHub markdown filters out](https://stackoverflow.com/questions/44831505/which-inline-html-styles-does-github-markdown-accept) out inline `style` attributes.

conico974 (approved):
LGTM Thanks

README.md
```diff
@@ -13,9 +13,9 @@
 
 ---
 
-<p align="center">
-  <a style="font-size: 2.5rem;font-weight:bold;" href="https://open-next.js.org/aws">Docs</a>
-</p>
+<h1 align="center">
+  <a href="https://open-next.js.org/aws">Docs</a>
```
conico974, line 17: Wait the docs is not correct here. It should be `opennext` not `open-next`

conico974 (approved):

PR #581, opened 2024-10-19
Fix github actions warnings for node20

socsieng:
> The following actions use a deprecated Node.js version and will be forced to run on node20: actions/checkout@v3, actions/setup-node@v3, pnpm/action-setup@v2, actions/cache@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/


conico974 (approved):
LGTM Thanks

PR #585, opened 2024-10-22
Support i18n localeDetection with value false

alacroix:
Hey there,

We found that the locale detection is forced and doesn't support the option `localeDetection` to `false` as linked below:

https://nextjs.org/docs/pages/building-your-application/routing/internationalization#disabling-automatic-locale-detection

Since it's an opt-out option, I forced the exact check with `false`

Thanks!

### DIFF
--- a/.changeset/giant-horses-sin.md
+++ b/.changeset/giant-horses-sin.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Support i18n localeDetection with value false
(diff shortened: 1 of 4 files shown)


khuezy:
Hey thanks! Would you be open to adding some unit tests around the i18 stuff? eg: https://github.com/opennextjs/opennextjs-aws/tree/main/packages/tests-unit/tests/core/routing

I want to add the i18 to our `e2e` apps too since support for it is still iffy, but that's for another day.

alacroix:
Done @khuezy 👍 

I didn't test the `accept-header` file since it's copied from NextJS

```js
// Copied from Next.js source code
// https://github.com/vercel/next.js/blob/canary/packages/next/src/server/accept-header.ts
```

conico974 (approved):
LGTM Thanks for that

PR #590, opened 2024-10-22
fix 500 error page being cached

conico974:
It seems that sometimes the 500 error page gets cached for some reason.
We don't want that to happen.

Update: 

This can happen for page router if your 500 (or 404) page use `getStaticProps`. 
For every route in page router that throws an exception, the cache-control gets overwritten and set as if it was an SSG route. The problem is that it doesn't remove the previous header that was set (including set-cookie). This could lead to session leakage.

Haven't tested in Next 15, this is probably fixed there since they don't override your cache-control anymore ( It might still happen if you don't set any cache-control for the route )

I added an env variable `OPEN_NEXT_DANGEROUSLY_SET_ERROR_HEADERS` that if set to true will disable the cache-control override that we do

### DIFF
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -182,6 +182,7 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
           : [...initialCookies, ...this._cookies];
     }
     this.fixHeaders(this.headers);
+    this.fixHeadersForError();
 
     // We need to fix the set-cookie header here
     this.headers[SET_COOKIE_HEADER] = this._cookies;
@@ -275,6 +276,7 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
   getFixedHeaders(): OutgoingHttpHeaders {
     // Do we want to apply this on writeHead?
     this.fixHeaders(this.headers);
+    this.fixHeadersForError();
     // This way we ensure that the cookies are correct
     this.headers[SET_COOKIE_HEADER] = this._cookies;
     return this.headers;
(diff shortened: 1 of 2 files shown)


conico974:
> Would this be related at all w/ the one user on discord having a random session leak?

It's probably it's error yeah. I need to investigate this more closely but based on what he describes it is likely the issue.
If the `set-cookie` headers has been set and an error occurs after that then the 500 could be cached.
Still not sure why next would set cache-control on a 500 though, seems very weird

khuezy:
Yea I'm not sure what the use case would be for errors to be cached. The HTTP spec says not to cache 500's.

conico974:
> Yea I'm not sure what the use case would be for errors to be cached. The HTTP spec says not to cache 500's.

There is a use case that i can think of. Just to reduce the load on the server when you know that something is broken, you could cache the error for a small amount of time. 
Still i don't think it's a good idea, and it might just be that since they load the 500 page from the incremental cache, they assume it is ISR/SSG and apply the cache-control headers.

sommeeeer:
this is interesting, probably what that guy was experiencing. i like the idea of having an environment variable to opt out of it. would be nice to know why `this.setHeader()` is not working tho.

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -384,4 +386,20 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
+
+  // For some reason, next returns the 500 error page with some cache-control headers
+  // We need to fix that
+  private fixHeadersForError() {
+    if (process.env.OPEN_NEXT_DANGEROUSLY_SET_ERROR_HEADERS === "true") {
+      return;
+    }
+    // We only check for 404 and 500 errors
+    // The rest should be errors that are handled by the user and they should set the cache headers themselves
+    if (this.statusCode === 404 || this.statusCode === 500) {
```
khuezy, line 398: Do we want to make this fix on ALL status except `2xx` and `301` (permanent redirect)?

khuezy (approved):
LGTM, just 1 question on the statuCodes

packages/open-next/src/http/openNextResponse.ts
conico974: I don't think so, this is a very special case of Next overwriting the cache-control and it could only happen for 404 and 500.
There is the case of 307/308 by redirecting in ISR but those are ISR so we expect the status code.

The rest of the status code are user adding those themselves. And they should be able to do so if they wish to but that's their responsability

PR #595, opened 2024-10-25
chore: upgrade pnpm to 9

sommeeeer:
this PR bumps pnpm to 9.

btw, pnpm is giving a warning. should we remove `workspaces` from `package.json`? or is it used by something else?

```bash
 WARN  The "workspaces" field in package.json is not supported by pnpm. Create a "pnpm-workspace.yaml" file instead.
```


khuezy:
Yea I think we can remove the `workspaces` in package.json. We're using the `pnpm-workspace.yaml`

khuezy (approved):
LGTM thanks!

conico974 (approved):
LGTM as well. Thanks
I'll merge this one after the release though, just in case

PR #597, opened 2024-10-26
Don't force args for install

conico974:
For some libs it seems that the default args that we used doesn't work, we can make all of them optional

### DIFF
--- a/packages/open-next/src/build/installDeps.ts
+++ b/packages/open-next/src/build/installDeps.ts
@@ -23,11 +23,18 @@ export function installDependencies(
     logger.info(`Installing dependencies for ${name}...`);
     // We then need to run install in the tempDir
     // We don't install in the output dir directly because it could contain a package.json, and npm would then try to reinstall not complete deps from tracing the files
-    const installCommand = `npm install --arch=${
-      installOptions.arch ?? "arm64"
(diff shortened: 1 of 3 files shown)


packages/open-next/src/build/installDeps.ts
```diff
@@ -23,11 +23,18 @@ export function installDependencies(
...
+      : "";
+    const targetOption = installOptions.nodeVersion
+      ? `--target=${installOptions.nodeVersion}`
+      : "";
+    const libcOption = installOptions.libc
+      ? `--libc=${installOptions.libc}`
+      : "";
+
+    const additionalArgs = installOptions.additionalArgs ?? "";
+    const installCommand = `npm install --platform=linux ${archOption} ${targetOption} ${libcOption} ${additionalArgs} ${installOptions.packages.join(" ")}`;
```
khuezy, line 37: what if the platform is `Windows`?

conico974: People still use windows as a server ? :joy: 
That's only for the server. Honestly there is probably dozens of other issue trying to run OpenNext on windows.

khuezy: I read someone was insterested in running ON on Azure 🤷 
We'll cross that bridge when we get there.

khuezy (approved):
🚢 it

PR #598, opened 2024-10-26
Fix external redirect trailing

### LINKED ISSUE #591: External redirects issues without trailing slash
What settings do I need to use in NextJS to redirect ALL 'source/path' from a host/domain to forward to the index/homepage of another domain.

this **WORKS**:
```
      {
        source: '/(.*)',
        destination: 'https://primary.domain.com',
        permanent: false,
        locale: false,
        has: [
          {
            type: 'header',
            key: 'x-forwarded-host',
            value: 'other.com',
          },
        ],
      },
```

this does **NOT** work:
```
      {
        source: '/:path*',
        destination: 'https://primary.domain.com',
        permanent: false,
        locale: false,
        has: [
          {
            type: 'header',
            key: 'x-forwarded-host',
            value: 'other.com',
          },
        ],
      },
```
it outputs 
`location: https://primary.domain.com/undefined` as response headers

BUT if you just put a trailing slash at the end ... then that **WORKS**....
```
      {
        source: '/:path*',
        destination: 'https://primary.domain.com/',
        permanent: false,
        locale: false,
        has: [
          {
            type: 'header',
            key: 'x-forwarded-host',
            value: 'other.com',
          },
        ],
      },
```

https://discord.com/channels/983865673656705025/1298382310831554560/1298662265037262930

conico974:
Fix #591

### DIFF
--- a/packages/open-next/src/core/routing/matcher.ts
+++ b/packages/open-next/src/core/routing/matcher.ts
@@ -229,7 +229,7 @@ export function handleRewrites<T extends RewriteDefinition>(
     }
     rewrittenUrl = isExternalRewrite
       ? `${protocol}//${rewrittenHost}${rewrittenPath}`
-      : `/${rewrittenPath}`;
+      : `${rewrittenPath}`;
     // Should we merge the query params or use only the ones from the rewrite?
     finalQuery = {
       ...query,
--- a/.changeset/fuzzy-camels-check.md
+++ b/.changeset/fuzzy-camels-check.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Fix external redirect trailing
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -47,7 +47,7 @@ export function getUrlParts(url: string, isExternal: boolean) {
     const match = url.match(regex);
     return {
       hostname: "",
-      pathname: match?.[1] ?? url,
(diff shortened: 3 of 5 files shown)


packages/open-next/src/core/routing/matcher.ts
```diff
@@ -229,7 +229,7 @@ export function handleRewrites<T extends RewriteDefinition>(
     }
     rewrittenUrl = isExternalRewrite
       ? `${protocol}//${rewrittenHost}${rewrittenPath}`
-      : `/${rewrittenPath}`;
+      : `${rewrittenPath}`;
```
khuezy, line 232: Are we sure we can trust `rewrittenPath` will never have a leading slash? If no, maybe we need a util function that strips the final `url` from redundant slashes.

I just wanted to make sure we don't play whack-a-mole with these kind of things.

khuezy (approved):
LGTM, just small comment.

packages/open-next/src/core/routing/matcher.ts
conico974: It does always have one now actually. This line has changed https://github.com/opennextjs/opennextjs-aws/blob/2c2b3438a21f15981deeb3e3cfc82a0691879c1c/packages/open-next/src/core/routing/util.ts#L50

The only things where i could see an error here is if the url itself is broken, so it would be a user error.

I made the change so that the pathname from `getUrlPart` always start with a leading slash (or is empty).
Before this change pathname would have a leading slash when external and none when internal

PR #599, opened 2024-10-26
fix workflow

conico974:
For ref : https://github.com/opennextjs/opennextjs-aws/actions/runs/11522037287


khuezy (approved):
Good catch LGTM

PR #600, opened 2024-10-26
fix: add more reliable sources to the revalidate e2e test

sommeeeer:
This PR adds more reliable sources to the fetches in `revalidateTag.test.ts` e2e

### DIFF
--- a/examples/app-router/app/revalidate-path/page.tsx
+++ b/examples/app-router/app/revalidate-path/page.tsx
@@ -1,26 +1,19 @@
 export default async function Page() {
-  const timeInParis = await fetch(
-    "https://www.timeapi.io/api/time/current/zone?timeZone=Europe%2FParis",
-    {
-      next: {
-        tags: ["path"],
-      },
+  const responseSST = await fetch("https://sst.dev", {
+    next: {
+      tags: ["path"],
     },
-  );
+  });
   // This one doesn't have a tag
-  const timeInLondon = await fetch(
-    "https://www.timeapi.io/api/time/current/zone?timeZone=Europe%2FLondon",
-  );
-  const timeInParisJson = await timeInParis.json();
-  const parisTime = timeInParisJson.dateTime;
-  const timeInLondonJson = await timeInLondon.json();
-  const londonTime = timeInLondonJson.dateTime;
+  const responseOpenNext = await fetch("https://opennext.js.org");
+  const reqIdSst = responseSST.headers.get("x-amz-cf-id");
+  const dateInOpenNext = responseOpenNext.headers.get("date");
   return (
     <div>
-      <h1>Time in Paris</h1>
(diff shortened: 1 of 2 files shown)


khuezy (commented):
We're on next 15 now correct? If so, fetch by default doesn't cache, so I don't think this e2e would test for the revalidate since the amz-id would be unique on each request.

sommeeeer:
> We're on next 15 now correct? If so, fetch by default doesn't cache, so I don't think this e2e would test for the revalidate since the amz-id would be unique on each request.

yeah we are on 15, forgot about the new beautiful cache behaviour. however, you can see the behavior yourself too: [`/revalidate-path`](https://dz1qocqqyz45x.cloudfront.net/revalidate-path)
[`/api/revalidate-path`](https://dz1qocqqyz45x.cloudfront.net/api/revalidate-path)

for some reason its working as it was next 14. maybe its something im missing thats completely obvious here. need to dig more into it

sommeeeer:
yeah so the fetch caches is not cached by default, but our page is. we are modifying the cache headers on `/revalidate-path` in the middleware, so CloudFront doesn't cache it. The page is SSG still, so if I add `export const dynamic = `force-dynamic` to the top, you can see that the fetches is still not cached. I'll put this into draft, so I can work on fixing this.

sommeeeer:
alright, so this one was a bit confusing at first, but it must be related to the fact [#L46-L54](https://github.com/opennextjs/opennextjs-aws/blob/1c271709497708d393e35f1fb5cfe78ca0f955f2/examples/app-router/middleware.ts#L46-L54) made me think this route is not SSG. 

since `/revalidate-path` is actually SSG, you can see that by looking at the `x-nextjs-cache` header being `HIT` and in the build log.

so each visit to `/revalidate-path` would still show the fetches as they were cached, but thats just because the page is static. triggering `revalidatePath("/revalidate-path")` would revalidate the entire page and thus also the fetches. 

therefor we need to add `export const dynamic = "force-dynamic"` on top of this page and also `force-cache` on the fetches to get the caching-behavior we are aiming for. 

this made me realize tho, should we maybe rework this a bit?

conico974:
> therefor we need to add export const dynamic = "force-dynamic" on top of this page and also force-cache on the fetches to get the caching-behavior we are aiming for.

That's not the point of this test. We test that it revalidate the page not the fetch cache.

We may want to add a new e2e test to test the fetch cache behavior, but it's likely broken again in Next 15.
We could just cache those fetch here and add other test. (I'm surprised that the codemod did not change that to use caching, seems weird)

sommeeeer:
> That's not the point of this test. We test that it revalidate the page not the fetch cache.

alright, so i guess i can just keep the code as it is for now, and in a future PR add another one for the fetch cache.

conico974 (approved):
LGTM Thanks

PR #601, opened 2024-10-27
fix fetch cache for next 15

conico974:
They changed the options again, fetchCache were never hit, it would be set but never read

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -150,11 +150,10 @@ export default class S3Cache {
     if (globalThis.disableIncrementalCache) {
       return null;
     }
-    const _isFetchCache = isFetchCache(options);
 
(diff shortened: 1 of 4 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -122,21 +144,17 @@ export default class S3Cache {
...
-        ? options.kindHint
-          ? options.kindHint === "fetch"
-          : options.fetchCache
-        : options;
+    const _isFetchCache = isFetchCache(options);
 
     const softTags = typeof options === "object" ? options.softTags : [];
     const tags = typeof options === "object" ? options.tags : [];
-    return isFetchCache
+    return _isFetchCache
```
vicb, line 157:
```suggestion
    return isFetchCache(options)
```

and remove `_isFetchCache`

vicb (approved):

PR #603, opened 2024-10-28
Fix redirect when containing "+" and decode values for URLSearchParams

john-trieu-nguyen:
Ran into a bug when using next.config.js redirects.

An example configuration causing a failed redirect (404 error):
```
{
  source: '/foo',
  destination: '/search?bar=hello+world&baz=new%2C+earth',
  permanent: true
}
```

Without encoding the "+", the redirect will fail.
Also updated convertToQueryString to decode the values when appending to URLSearchParams as per spec.

### DIFF
--- a/packages/tests-unit/tests/core/routing/matcher.test.ts
+++ b/packages/tests-unit/tests/core/routing/matcher.test.ts
@@ -266,6 +266,24 @@ describe("handleRedirects", () => {
 
     expect(result).toBeUndefined();
   });
+
+  it("should redirect with + character and query string", () => {
+    const event = createEvent({
+      url: "/foo",
+    });
+
+    const result = handleRedirects(event, [
+      {
+        source: "/foo",
+        destination: "/search?bar=hello+world&baz=new%2C+earth",
+        internal: true,
+        statusCode: 308
+      },
+    ]);
+
+    expect(result.statusCode).toEqual(308);
+    expect(result.headers.Location).toEqual("/search?bar=hello+world&baz=new%2C+earth");
+  });
 });
 
 describe("handleRewrites", () => {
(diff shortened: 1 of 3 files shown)


khuezy (approved):
LGTM thanks!

packages/tests-unit/tests/core/routing/matcher.test.ts
```diff
@@ -266,6 +266,26 @@ describe("handleRedirects", () => {
...
     expect(result).toBeUndefined();
   });
+
+  it("should redirect with + character and query string", () => {
+    const event = createEvent({
+      url: "/foo",
+    });
+
+    const result = handleRedirects(event, [
+      {
```
conico974, line 276: You're not using the correct things here. You should be using the format used in the output from `.next/routes-manifest.json`

conico974 (changes requested):
Test cannot pass with this

packages/tests-unit/tests/core/routing/matcher.test.ts
khuezy: Nice catch I missed this. Does the unit test run before or after approval? I must have missed the failing test.

john-trieu-nguyen: I've rewrote the test to pass.
`expect(result).toBeUndefined();`

This does feel a little untuitive... The "should redirect matching path" also expects result to be undefined though, so maybe I'm missing something. Is it because it's an internal redirect?

conico974: It should not be undefined, the problem is what args you're passing to the `handleRedirects` fn.
And the `"should redirect matching path"` has a mistake that i fixed in a recent PR, it should also not return undefined.

The params should contain a regex, and the easiest way to get a correct one is to define your redirect in next config and run build, then get the correct form from `.next/routes-manifest.json`

conico974: > Does the unit test run before or after approval?
It runs after, that's probably why you missed it

john-trieu-nguyen: Done, thanks!

conico974 (approved):
LGTM Thanks

PR #604, opened 2024-10-28
fix(middleware): always compiles the middleware

vicb:
Prior to this PR the middleware would only be compiled when a middleware.ts exists.

@conico974 Please double check the `!external` path - I'm less familiar with that one. Thanks!


packages/open-next/src/adapters/plugins/README.md
```diff
@@ -1,10 +1,10 @@
-### Known issues
+# Known issues
 
-Do not include `types` in #override and #imports, as esbuild will remove preceeding comments (ie it removes //#override id)when it builds.
+Do not include `types` in #override and #imports, as esbuild will remove preceding comments (ie it removes //#override id)when it builds.
 
 Instead, put the `import type` outside like:
 
-```
+```js
```
conico974, line 7: Should be ts here

conico974 (approved):
LGTM Thanks
I've tested deploying without an external middleware and it works just fine

PR #611, opened 2024-10-30
fix(middleware): respect host protocol

vicb:
The middleware handler default to https and it's causing troubles with the cloudflare dev server (wrangler dev) which uses http by default.

It is possible to force wrangler dev to use https (--local-protocol https) but the browser then shows a warning because the dev certificate is not secure.

### DIFF
--- a/packages/open-next/src/core/routing/middleware.ts
+++ b/packages/open-next/src/core/routing/middleware.ts
@@ -54,9 +54,11 @@ export async function handleMiddleware(
   // We bypass the middleware if the request is internal
   if (internalEvent.headers["x-isr"]) return internalEvent;
 
+  const protocol = new URL(internalEvent.url).protocol;
   const host = internalEvent.headers.host
-    ? `https://${internalEvent.headers.host}`
+    ? `${protocol}//${internalEvent.headers.host}`
     : "http://localhost:3000";
+
   const initialUrl = new URL(normalizedPath, host);
   initialUrl.search = convertToQueryString(query);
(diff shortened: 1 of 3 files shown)


vicb:
@conico974 I have only update the middleware as I'm not sure if we will use the edge-adapter.

packages/open-next/src/core/routing/middleware.ts
```diff
@@ -54,8 +54,10 @@ export async function handleMiddleware(
   // We bypass the middleware if the request is internal
   if (internalEvent.headers["x-isr"]) return internalEvent;
 
+  const protocol = new URL(internalEvent.url).protocol;
```
conico974, line 57: We can't do that here, in lambda `url` only contains the rawPath and the query, no protocol or host
We could use https://github.com/opennextjs/opennextjs-aws/blob/811bdc0008213eed9f859fc505c92d85327a1bac/packages/open-next/src/core/routing/util.ts#L44 along with isExternal https://github.com/opennextjs/opennextjs-aws/blob/811bdc0008213eed9f859fc505c92d85327a1bac/packages/open-next/src/core/routing/util.ts#L19

If not found we default to `https`, would this work for you ?

conico974 (commented):
What's the issue with the edge-adapter ?

vicb:
@conico974 PTAL and let me know if the implementation is ok. Thanks!

conico974 (approved):
LGTM Thanks

PR #612, opened 2024-10-30
feature(edge): add a way for convertTo to return a Request

vicb:
This is a temporary "hack" for the cloudflare adapter.

Why we need this:
- It is not possible to fetch a worker in the same zone
- Our first implementation will bundle the middleware and the server in a single worker

We will work on a better architecture later but this will unblock cloudflare for now and does not add overhead/code to the current edge converter. The alternative would be to fork the edge converter but it would be harder to keep the implemenations in sync.

### DIFF
--- a/.changeset/fluffy-peas-live.md
+++ b/.changeset/fluffy-peas-live.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+feature(edge): add a way for convertTo to return a Request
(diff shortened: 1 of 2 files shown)


conico974 (approved):
LGTM Thanks

Everything "edge" related will likely need a rework at some point, i don't like the idea of the converter actually doing a fetch.
It should probably be done on the wrapper side so that the edge converter is not dependent on cloudflare or any other potential platform.

vicb:
> Everything "edge" related will likely need a rework at some point, i don't like the idea of the converter actually doing a fetch.
> It should probably be done on the wrapper side so that the edge converter is not dependent on cloudflare or any other potential platform.

+1 to that.

I think having multiple possible infra "implementation" in the same code will help us come with a better archi.

PR #613, opened 2024-10-30
feat(middleware): add ability to force single build pass

vicb:
This is the second (and last) hack required for the middleware bundle on cloudflare.

For us the second build pass is hanlded by wrangler. This PR adds an option to disable that pass.

### DIFF
--- a/.changeset/yellow-phones-juggle.md
+++ b/.changeset/yellow-phones-juggle.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+feat(middleware): add ability to force single build pass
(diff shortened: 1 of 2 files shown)


conico974 (approved):
LGTM

FYI you can't merge multiple PR at the same time without crashing the e2e test (the cdk cannot update an app that are already in the `UPDATE_IN_PROGRESS` state )
You need to wait for the end of the previous e2e test. Also something that we should rework in the future.

vicb:
> FYI you can't merge multiple PR at the same time without crashing the e2e test (the cdk cannot update an app that are already in the `UPDATE_IN_PROGRESS` state )
> You need to wait for the end of the previous e2e test. Also something that we should rework in the future.

Ah cool, I was just looking at why tests failed. Thanks!

PR #614, opened 2024-10-31
add main and experimental prereleases

dario-piotrowicz:
> [!note]
> this PR is matching what I am doing here: https://github.com/opennextjs/opennextjs-cloudflare/pull/115
> (I'm not sure if we also want an `experimental` branch here too?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Besides the standard npm releases we also automatically publish prerelease packa
   Updated with every push to the `main` branch, this prerelease contains the most up to date yet (reasonably) stable version of the package.
 - `https://pkg.pr.new/@opennextjs/aws@experimental`
   Updated with every push to the `experimental` branch, this prerelease contains the latest experimental version of the package (containing features that we want to test/experiment on before committing to).
-  Which you can simply install directly with your package manager of choice, for example: `npm i https://pkg.pr.new/@opennextjs/aws@experimental`
+  Which you can simply install directly with your package manager of choice, for example: `npm i https://pkg.pr.new/@opennextjs/aws@main`
 
 ## Contribute
 
--- a/README.md
+++ b/README.md
@@ -65,7 +65,11 @@ Besides the standard npm releases we also automatically publish prerelease packa
 - `https://pkg.pr.new/@opennextjs/aws@experimental`
   Updated with every push to the `experimental` branch, this prerelease contains the latest experimental version of the package (containing features that we want to test/experiment on before committing to).
 
-Which you can simply install directly with your package manager of choice, for example: `npm i https://pkg.pr.new/@opennextjs/aws@main`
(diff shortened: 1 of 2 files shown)


conico974 (approved):
Yeah it make sense to have an experimental branch as well, other than that LGTM

dario-piotrowicz:
> Yeah it make sense to have an experimental branch as well, other than that LGTM

done 🙂 

Given the extra branch I've also re-aligned the text here with that of the Cloudflare adapter 

please let me know if it looks all good

conico974 (approved):
LGTM, Thanks for that

README.md
```diff
@@ -56,6 +56,15 @@ This will output A LOT of additional logs to the console. This also disable mini
...
 You can read more about the configuration in the [docs](https://opennext.js.org/aws/config)
 
+## Preleases
+
+Besides the standard npm releases we also automatically publish prerelease packages on branch pushes (using [`pkg.pr.new`](https://github.com/stackblitz-labs/pkg.pr.new)):
+
+- `https://pkg.pr.new/@opennextjs/aws@main`:
+  Updated with every push to the `main` branch, this prerelease contains the most up to date yet (reasonably) stable version of the package.
+- `https://pkg.pr.new/@opennextjs/aws@experimental`
+  Updated with every push to the `experimental` branch (when present), this prerelease contains the latest experimental version of the package (containing features that we want to test/experiment on before committing to).
```
vicb, line 66:
```suggestion
  Updated with every push to the `experimental` branch, this prerelease contains the latest experimental version of the package (containing features that we want to test/experiment on before committing to).
```

It's present when you push to it :)

README.md
```diff
@@ -56,6 +56,15 @@ This will output A LOT of additional logs to the console. This also disable mini
...
+  Which you can simply install directly with your package manager of choice, for example: `npm i https://pkg.pr.new/@opennextjs/aws@main`
```
vicb, line 67:
```suggestion
  Which you can simply install directly with your package manager of choice, for example: `npm i https://pkg.pr.new/@opennextjs/aws@experimental`
```

vicb (approved):
Thanks Dario!

README.md
dario-piotrowicz: good point 😅

README.md
dario-piotrowicz: Sorry I accepted the suggestion but then realised that this is probably not really what we want, since the `experimental` prerelease might not be the always relevant (like when there is no "experimentation" going on etc... and the `experimental` branch either doesn't exist or is out of date).

using `main` here does seem the right choice to me, I also noticed that I messed up the formatting 🤦

This is how it looks now:
Screenshot 2024-11-01 at 09 50 57

Hopefully this is fine? 🙂

PR #616, opened 2024-11-04
chore: configure biome linter

maxmorozoff:
This PR migrates project’s linting configuration from ESLint to [Biome](https://biome.dev/).
See: https://github.com/opennextjs/opennextjs-aws/issues/538

## Changes
- Migrated configuration
- Updated CI Action

## Performance improvements:
With Biome, linting the project now takes only 45ms on my machine.

## Known Issues
Not all ESLint rules have direct equivalents in Biome as noted in [Biome's documentation](https://biomejs.dev/linter/rules-sources/#rules-from-other-sources).


khuezy:
Thanks for the PR! I'm not familiar w/ biome so I'll defer to @conico974 and @vicb

vicb:
I'll defer to @conico974 as he created the linked cleanup issue.

_I usually agree with "if it ain't broken don't fix it" - especially when it comes to JS "build" tooling._

conico974:
> I usually agree with "if it ain't broken don't fix it" - especially when it comes to JS "build" tooling.

I kind of agree with that, but biome is so fast and it's a single deps that i think it may be worth it.

Unless there is a good reason to not merge this i think we should do it. I'll test that everything is working correctly tomorrow

khuezy:
@maxmorozoff 
`With Biome, linting the project now takes only 45ms on my machine.`
How long did it take `eslint`?

maxmorozoff:
> @maxmorozoff 
> `With Biome, linting the project now takes only 45ms on my machine.`
> How long did it take `eslint`?

Around 40 seconds. 
However, ESLint had more rules and plugins enabled. 

You could consider enabling additional rules in Biome for a closer comparison.

As a temporary solution, we could run ESLint after Biome. This would have minimal impact on performance while allowing for a more gradual adoption, as it may require more fixes across the codebase.

vicb:
> I usually agree with "if it ain't broken don't fix it"

On the other hand, TS update is blocked by the current eslint config: 

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=4.3.5 <5.4.0

YOUR TYPESCRIPT VERSION: 5.6.3

Please only submit bug reports when using the officially supported version.

=============
```

Switching to biome could fix that so that we can update TS - which would allow some code simplification, as newer TS have improved type inference

biome.json
```diff
@@ -0,0 +1,57 @@
+{
+  "$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
+  "vcs": { "enabled": false, "clientKind": "git", "useIgnoreFile": false },
```
conico974, line 3: We want this to be enabled and use the ignore files, otherwise it will try to lint a lot of useless files

conico974 (changes requested):
There is some missing stuff here.

And we probably want to either switch to `recommended: true` (and probably disable some rules that we don't need) or add more rules.

I don't think using eslint for other missing rule is a good idea, we either fully switch to biome or stick with eslint. There is no point in running multiple linters

biome.json
```diff
@@ -0,0 +1,57 @@
...
+  "files": {
+    "ignoreUnknown": true,
+    "ignore": [
+      "node_modules/**",
+      "**/node_modules/**",
+      "**/pnpm-lock.yaml",
+      "**/dist/**",
+      "**/out/**",
+      "packages/open-next/assets/sharp-node-modules",
+      "packages/open-next/src/minimize-js.ts"
```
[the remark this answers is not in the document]
maxmorozoff: The `minimize-js.ts` has an `eslint-disable` comment. In Biome to ignore this file, it should be configured in `biome.json` ([docs](https://biomejs.dev/linter/#ignore-code))

Other paths could be removed.

maxmorozoff:
The current config lints the project with no errors and with 2091 warnings. Some rules come with automated safe fixes that could be applied to the codebase. Some other rules have unsafe fixes. 

## Steps involved
### Enabled all rules
```json
{
  "linter": {
    "enabled": true,
    "rules": {
      "all": true
    }
  },
}
```
### Saved lint report to json and extracted rule names with a script

Encountered some issues with the --reporter=json option, as it's experimental. I used yq as a workaround.
```js
// $ pnpm lint --diagnostic-level=error --reporter=junit | yq -p=xml -o=json > report.json

import report from "./report.json";

const errors = report.testsuites.testsuite.filter(
  (testcase) =>
    testcase["+@errors"] &&
    testcase.testcase["+@name"].startsWith("org.biome.lint"),
);
const rulesSet = new Set(errors.map((testcase) => testcase.testcase["+@name"]));
const config = {};
rulesSet.forEach((rule) => {
  const [, , , pkg, name] = rule.split(".");
  config[pkg] = {
    all: true,
    ...config[pkg],
    [name]: "warn",
  };
});

console.log(JSON.stringify(config, null, 2));
```

biome.json
```diff
@@ -0,0 +1,57 @@
...
+    ]
+  },
+  "formatter": { "enabled": false },
+  "organizeImports": { "enabled": false },
```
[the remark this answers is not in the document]
maxmorozoff: Should we configure formatting in `biome.json` or within a [.editorconfig](https://editorconfig.org/)?

conico974:
@maxmorozoff We don't want to enable all rule, that's for sure.
For the recommended one we can remove: 
- everything a11y related (as these probably only happens for the e2e test)
- I'd remove the `noForEach` as well

The rest is probably fine, and we can apply the safe one in this PR. 
For the unsafe one, we may have to look one by one and fix them in subsequent PR. As of now we can set them as warning and fix them later. We may want to add some other rule as well, but again in future PR

maxmorozoff:
@conico974, the default value for `recommended` is `true`, so I've omitted it. Accessibility errors occurred in the `examples/` directory. Should I disable these rules globally, or should we keep the recommended rules? 
Also, should I update `.vscode/settings.json` and add `.vscode/extensions.json`? 
Should I add the commit with fixes in `.git-blame-ignore-revs`?

conico974:
@maxmorozoff 
> Should I disable these rules globally, or should we keep the recommended rules?

Disable them, examples only contains e2e, we don't care that much about having perfect app there

> Also, should I update .vscode/settings.json and add .vscode/extensions.json?
Should I add the commit with fixes in .git-blame-ignore-revs?

Good idea for both questions, go ahead

maxmorozoff:
@conico974, could you please review the latest commits and let me know if they look good to you? Any additional feedback is also welcome!

Thank you!

khuezy:
What do you think of adding `--diagnostic-level=error`. When I try `biome lint` it includes a bunch of warns

I'm playing around w/ biome and it's really fast!

conico974:
> What do you think of adding --diagnostic-level=error. When I try biome lint it includes a bunch of warns

My advice is that we should fix those warning (or disable them if we feel they're useless) just not in this PR, i just want to avoid a 300 files changed PR where it would be way too easy to miss something that might break something (especially with the eslint plugin we have that can replace code)

> I'm playing around w/ biome and it's really fast!

Yeah i've tested it a bit and it's so fast, it seems almost unreal :laughing:. There is a bunch of new javascript tooling around lately that are insane speed wise (like the [oxc](https://github.com/oxc-project/oxc) project)

.editorconfig
conico974, line 1: Do we need this file if biome is doing the formatting?

biome.jsonc
```diff
@@ -0,0 +1,93 @@
...
+        "noExplicitAny": "info",
+        "noGlobalIsNan": "warn",
+        "noImplicitAnyLet": "warn",
+        "noAssignInExpressions": "warn",
+        "noConfusingVoidType": "warn",
+        "useNamespaceKeyword": "warn" // safe fix
+      },
+      "style": {
+        "noUnusedTemplateLiteral": "warn",
+        "noInferrableTypes": "warn", // safe fix
```
conico974, line 26: Did you apply these safe fixes ? If so no need to set them as warn, we want to avoid them as much as possible so an error is fine

conico974 (commented):
I just have 2 question here, but if everything is fine we should merge this
As for what's need to be done next, we need to either ignore or fix the lint rule, in this order : suspicious, correctness, performance, complexity and lastly style.
When all of this is done we'll need to think if we want to add other rule

biome.jsonc
maxmorozoff: I've only applied the fixes that were causing formatting or import order errors, which I forgot to mention. Now, running `pnpm lint` exits with 0, while `pnpm lint:fix` would apply additional linting fixes. It might be better to address these in the next PR, as they are more nuanced and might not be necessary if some rules are disabled.

.editorconfig
maxmorozoff: `.editorconfig` is optional, as formatting can be managed through the Biome config. However, including this file helps enforce consistent settings in supported editors and can be read by various tools, not just Biome, which could be beneficial in some scenarios.

Unfortunately, it doesn't support all of Biome's formatting settings, so we may still need to add certain rules in `biome.jsonc`. Therefore, it can't be considered a single source of truth for formatting, and you may prefer to keep all settings in the Biome config.

vicb:
great work on this PR!

.editorconfig
conico974: I'd rather have a single source of truth, it's confusing to have 2 formatter doing the same thing. 
We should remove the `.editorconfig` and only keep biome

conico974:
> biome use to be rome and oxc is now void0 

And biome use oxc under the hood and oxc also have a linter :exploding_head:

conico974 (approved):
LGTM Thanks a lot for this

maxmorozoff:
@conico974 Thank you! Would you like me to open the next PR to address the linting warnings, or do you plan to handle it later? Happy to help!

conico974:
@maxmorozoff i'd be happy if you open them yeah

conico974:
Maybe i spoke too soon, looks like @vicb is already taking care of a bunch of them

vicb:
#627 should conflict with #626 and I don't think I'll have time to work on the PR before Tuesday, so up to you.

PR #617, opened 2024-11-06
feat: add support for next15 geolocation

vicb:


packages/open-next/src/overrides/converters/aws-cloudfront.ts
conico974, line 1: I think i misunderstood what you were saying on discord, we don't want to modify this but the cloudfront function in sst.
This one is only used when using lambda@edge (either with external middleware or as a full server fuction).
If we'd do that we'll also have to modify the other lambda converter

If we'd want to provide these headers for every function, we could keep the `x-open-next` one in the cloudfront functions and we apply the `x-vercel` in here : https://github.com/opennextjs/opennextjs-aws/blob/c5aa622ddb8b89e5c73233201f30826e668a843e/packages/open-next/src/core/routingHandler.ts#L73

We could add a function there that would add the `x-vercel` headers only if the `x-open-next` headers are present

packages/open-next/src/overrides/wrappers/cloudflare.ts
```diff
@@ -6,12 +6,20 @@ import type {
 
 import type { MiddlewareOutputEvent } from "../../core/routingHandler";
 
+const cfPropNameToHeaderName = {
+  city: "x-vercel-ip-city",
```
conico974, line 10: If we decide to change the headers in `routingHandler` then it would make sense to rename those as `x-open-next`

conico974 (commented):
I misunderstood what we were discussing on discord, i added a few comment to explain

vicb:
@conico974 PTAL - I think it's better... after 25 force pushes 🤣

packages/open-next/src/core/routing/matcher.ts
```diff
@@ -149,7 +143,7 @@ export function addNextConfigHeaders(
     source,
     locale,
   } of configHeaders) {
-    const path = locale === false ? rawPath : localizedRawPath;
+    const path = locale === false ? event.rawPath : localizePath(event);
```
conico974, line 146: The `localizePath` was done outside in order to avoid recomputing the same value for every value. 
Not that big of a deal but worth mentioning

conico974 (approved):
Just a little note, other than that LGTM !!

packages/open-next/src/core/routing/matcher.ts
vicb: OOps, I missed that, I'll revert

PR #618, opened 2024-11-06
use a literal path to import the config

vicb:
This will help with Cloudflare were paths are resolved at build time.

Tested in both flat and monorepo setup.


khuezy (approved):

vicb:
Thanks @khuezy!

PR #619, opened 2024-11-06
refactor(overrides): move types to top level types folder

vicb:


khuezy (approved):
Thanks!

packages/open-next/src/types/overrides.ts
conico974, line 1: We probably want to add the other type of override here as well. They are all in `types/open-next.ts` at the moment: `WrapperHandler`, `Wrapper`,`OpenNextHandler`, `Warmer`, `OriginResolver`, `ImageLoader` and `Converter`

This way everything that can be overrided is in this file

khuezy:
ooof lots of type errors now.

vicb:
looks like I missed a few files 😅

conico974 (approved):
LGTM Thanks

PR #623, opened 2024-11-08
refactor: make callbacks consistent

vicb:
Minor change but 2 things were annoying me:

- The 2 callbacks were passed a different path (relative vs absolute)
- Some "work" was being done in the predicate function in createServerBundle.


packages/open-next/src/build/helper.ts
```diff
@@ -167,47 +167,56 @@ export async function esbuildAsync(
...
+ * @see `traverseFiles`.
+ *
+ * @param root Root directory to search.
+ * @param conditionFn Predicate used to delete the files.
+ */
 export function removeFiles(
   root: string,
-  conditionFn: (file: string) => boolean,
-  searchingDir: string = "",
+  conditionFn: (paths: {
```
conico974, line 180: Maybe we could extract the paths as it's own type since it is used here and in the 2 traverseFiles

conico974 (approved):
Just made a small comment, not mandatory at all
Other than that LGTM

vicb:
PTAL

conico974 (approved):
Perfect Thanks

PR #625, opened 2024-11-08
fix: cloudflare geolocation

vicb:
Use the code rather than the fullname.
See https://vercel.com/guides/geo-ip-headers-geolocation-vercel-functions#geolocation-headers

_(cloudfront correctly uses `cloudfront-viewer-region` vs `cloudfront-viewer-region-name`)_


conico974 (approved):
LGTM Thanks

khuezy (approved):

PR #626, opened 2024-11-08
next/after support and better AsyncLocalStorage context in general

conico974:
This PR make the AsyncLocalStorage available to the middleware as well as the edge function.

It also provide support for `next/after` everytime including for lambda. On lambda without streaming it will block the response until all after promise are either resolved or rejected. Tested on lambda with streaming, it works as expected.

It can also be used to emulate vercel request context (the waitUntil) for lib that may rely on it on serverless env. It needs this env variable `EMULATE_VERCEL_REQUEST_CONTEXT` to be set to be enabled

And for wrapper that has native support for `waitUntil` they can provide `globalThis.openNextWaitUntil` ( Should work on cloudflare, not tested though)

I haven't found a way to test this with our current e2e setup, if someone has an idea ?


vicb (approved):
Nice!
A couple nits.

khuezy:
oh jeez that's a lot of lint warnings lol

vicb (changes requested):
I think there are still a few instances of `__als` in the tests

conico974:
Ho right i missed them, should be good now

khuezy (approved):
LGTM. Would it make sense to add an e2e test for the next/after stuff?

conico974:
@khuezy Yeah it would make sense, but i haven't figured out a way with our current setup.
The e2e test on the next repo for this rely on local console.log, which we cannot really use here.
Probably something with an endpoint, but haven't figured it out yet

khuezy:
To truly test `after`, the page request has to fully terminate, so the ALS would lose context w/ a new lambda... so I think we have to store some flag in DDB or something.

conico974:
Ho this give me an idea, we create 2 endpoint one that got the after and another one that would be ssg.
On the one with after we call `revalidatePath` in a promise that resolve like 5s in `after`. We then just need to check if the ssg has changed or not and if the initial call didn't take more than 5s.

This is a bit hacky, but that's the only thing i can think of right now.
I'll work on this later tonight

khuezy:
Great idea, I don't think we'd need 2 endpoints for this. It can just be `after/page.tsx` where we call 
```
after(() => {
   revalidatePath('/after') // inside 5 second timeout
})
return <>{time}</>
```

The test would reload the page and check that the time hasn't changed until after 5 seconds...
I think that would work? I could be missing something tho.

examples/app-router/app/api/after/ssg/route.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export const dynamic = "force-static";
+
+export async function GET() {
+  const dateFn = unstable_cache(() => new Date().toISOString(), ["date"], {
+    tags: ["date"],
+  });
+  const date = await dateFn();
+  console.log("date", date);
+  return NextResponse.json({ date: date });
```
sommeeeer, line 12: shouldn't this be a `*.ts` file since its a route?

conico974: Good point, i'll change it

sommeeeer: great work on this PR btw

.changeset/new-dolphins-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+add support for next/after
```
vicb, line 5: Maybe add

> It can also be used to emulate vercel request context (the waitUntil) for lib that may rely on it on serverless env. It needs this env variable EMULATE_VERCEL_REQUEST_CONTEXT to be set to be enabled

packages/open-next/src/utils/promise.ts
```diff
@@ -58,3 +58,64 @@ export class DetachedPromiseRunner {
...
+  /** This should be considered unstable until `unstable_after` is stablized. */
+  const NEXT_REQUEST_CONTEXT_SYMBOL = Symbol.for("@next/request-context");
+
+  // This is needed by some lib that relies on the vercel request context to properly await stuff.
+  // Remove this when vercel builder is updated to provide '@next/request-context'.
+  const VERCEL_REQUEST_CONTEXT_SYMBOL = Symbol.for("@vercel/request-context");
+
+  const openNextStoreContext = globalThis.__openNextAls.getStore();
+
+  const awaiter =
```
vicb, line 83: nit: rename this to `waitUntil` and change below `get: () => ({ waitUntil })` - I feel like introducing a new name is not worth

vicb (approved):
Thanks,
only a few very minor things

examples/app-router/app/api/after/ssg/route.tsx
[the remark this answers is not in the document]
conico974: I think i'll remove the `console.log` entirely, not really useful

PR #627, opened 2024-11-08
refactor: code simplification

vicb:
A few refactors I had accumulated in a branch.

Some of them depends on recent TS which is why it's updated.

Contains the auto fixes from biome

### DIFF
--- a/packages/open-next/src/core/patchAsyncStorage.ts
+++ b/packages/open-next/src/core/patchAsyncStorage.ts
@@ -3,13 +3,13 @@ const mod = require("module");
 const resolveFilename = mod._resolveFilename;
 
 export function patchAsyncStorage() {
-  mod._resolveFilename = function (
+  mod._resolveFilename = ((
     originalResolveFilename: typeof resolveFilename,
     request: string,
     parent: any,
     isMain: boolean,
     options: any,
-  ) {
+  ) => {
     if (
       request.endsWith("static-generation-async-storage.external") ||
       request.endsWith("static-generation-async-storage.external.js")
@@ -35,5 +35,5 @@ export function patchAsyncStorage() {
       );
 
     // We use `bind` here to avoid referencing outside variables to create potential memory leaks.
-  }.bind(null, resolveFilename);
+  }).bind(null, resolveFilename);
 }
--- a/packages/open-next/src/core/require-hooks.ts
+++ b/packages/open-next/src/core/require-hooks.ts
@@ -132,7 +132,7 @@ function isApp() {
 }
 
 export function applyOverride() {
-  mod._resolveFilename = function (
+  mod._resolveFilename = ((
     originalResolveFilename: typeof resolveFilename,
     requestMapApp: Map<string, string>,
     requestMapPage: Map<string, string>,
@@ -140,7 +140,7 @@ export function applyOverride() {
     parent: any,
     isMain: boolean,
     options: any,
-  ) {
+  ) => {
     const hookResolved = isApp()
       ? requestMapApp.get(request)
       : requestMapPage.get(request);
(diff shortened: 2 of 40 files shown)


packages/open-next/src/core/patchAsyncStorage.ts
```diff
@@ -35,5 +35,5 @@ export function patchAsyncStorage() {
       );
 
     // We use `bind` here to avoid referencing outside variables to create potential memory leaks.
-  }.bind(null, resolveFilename);
+  }).bind(null, resolveFilename);
```
conico974, line 38: I know that bind work differently for `this` for arrow function vs normal ones. I guess it would not have any effect here, not entirely sure. If there is no effect then it's fine

packages/open-next/src/core/require-hooks.ts
```diff
@@ -132,21 +132,21 @@ function isApp() {
...
+  ) => {
     const hookResolved = isApp()
       ? requestMapApp.get(request)
       : requestMapPage.get(request);
     if (hookResolved) request = hookResolved;
     return originalResolveFilename.call(mod, request, parent, isMain, options);
 
     // We use `bind` here to avoid referencing outside variables to create potential memory leaks.
-  }.bind(null, resolveFilename, hookPropertyMapApp, hookPropertyMapPage);
+  }).bind(null, resolveFilename, hookPropertyMapApp, hookPropertyMapPage);
```
conico974, line 151: Same question as above

conico974 (commented):
I have a question, and i don't know the answer. I need to run some test to be sure.
If you know the answer then the rest LGTM and you can merge

packages/open-next/src/core/patchAsyncStorage.ts
maxmorozoff: If `bind` is not used to set the value of `this`, it can be used to pre-supply parameters. Here’s an example: https://stackoverflow.com/a/50630093

conico974: Yeah i know that, i have a different concern here, in the stackoverflow answer they said this 
> Partially, because bind returns a new function that calls the original function with a specific this — but arrow functions don't use the this they're called with, they completely ignore it

That's the part i'm concerned about, this **could** mean that if the underlying function use `this` it might have a completely different behaviour than expected. Not sure though

maxmorozoff: If it seems unsafe, we could disable the `useConst` rule globally or only for these cases.

I can make these changes in a separate PR tomorrow, focusing solely on biome-related modifications. This will make it easier to handle `.git-blame-ignore-revs`. After merging, you could then rebase this PR.

conico974: This is not a global things, these 2 function are very specific as their goal is to override existing functions.
I will not be able to test this today, if you'd wish to open a PR, maybe try focusing on stuff not covered in this PR, there is still a lot of warning https://github.com/opennextjs/opennextjs-aws/actions/runs/11746705545/job/32727036460?pr=627, no need to do the same job twice

vicb: But:
- do we want to update the comment / add warning
- report an issue to biome (this is an auto-fix)

conico974: > do we want to update the comment / add warning

I guess we can, but this is such a weird edge case that i hope we don't have to do this anytime soon 

> report an issue to biome (this is an auto-fix)

The behavior in biome is correct actually, if we add a `this` inside the function then biome does not report an error anymore. 
There might be an issue if we used the function directly instead of through `.call`, but since it is used to monkey patch a node built-in, this is probably not something that is covered (It should probably even be it's own rule, like `dontOverrideNodeBuiltIn`

PR #629, opened 2024-11-14
feat(debug): never minify in debug mode

vicb:


packages/open-next/src/build/bundleNextServer.ts
```diff
@@ -42,7 +42,11 @@ const externals = [
   "next/dist/compiled/next-server",
 ];
 
-export async function bundleNextServer(outputDir: string, appPath: string) {
+export async function bundleNextServer(
```
conico974, line 45: FYI this function should not be necessary in the future. Next should provide us directly with an already bundled `NextServer` in the future

conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review!

PR #630, opened 2024-11-14
Feat: Allow overriding the proxying for external rewrite

conico974:
This PR add an option to override the old `proxyRequest` fn that was used for external rewrites.
It also automatically make the proxying when used in the external middleware, no need to do it in a converter or the wrapper.

By default it is the same as before, but it could use `fetch` for the proxying

@vicb PTAL, i haven't tested the fetch part, i'd appreciate some feedback on this

### DIFF
--- a/packages/open-next/src/adapters/middleware.ts
+++ b/packages/open-next/src/adapters/middleware.ts
@@ -6,6 +6,7 @@ import { createGenericHandler } from "../core/createGenericHandler";
 import {
   resolveIncrementalCache,
   resolveOriginResolver,
+  resolveProxyRequest,
   resolveQueue,
   resolveTagCache,
 } from "../core/resolve";
@@ -17,6 +18,10 @@ globalThis.__openNextAls = new AsyncLocalStorage();
 const defaultHandler = async (internalEvent: InternalEvent) => {
   const originResolver = await resolveOriginResolver(
     globalThis.openNextConfig.middleware?.originResolver,
+  );
+
+  const externalRequestProxy = await resolveProxyRequest(
+    globalThis.openNextConfig.middleware?.override?.proxyExternalRequest,
   );
 
   //#override includeCacheInMiddleware
@@ -43,14 +48,15 @@ const defaultHandler = async (internalEvent: InternalEvent) => {
         let origin: Origin | false = false;
         if (!result.isExternalRewrite) {
           origin = await originResolver.resolve(result.internalEvent.rawPath);
+          return {
+            type: "middleware",
+            internalEvent: result.internalEvent,
+            isExternalRewrite: result.isExternalRewrite,
+            origin,
+            isISR: result.isISR,
+          };
         }
-        return {
-          type: "middleware",
-          internalEvent: result.internalEvent,
-          isExternalRewrite: result.isExternalRewrite,
-          origin,
-          isISR: result.isISR,
-        };
+        return externalRequestProxy.proxy(result.internalEvent);
       }
 
       debug("Middleware response", result);
--- a/packages/open-next/src/adapters/middleware.ts
+++ b/packages/open-next/src/adapters/middleware.ts
@@ -45,9 +45,10 @@ const defaultHandler = async (internalEvent: InternalEvent) => {
       const result = await routingHandler(internalEvent);
       if ("internalEvent" in result) {
         debug("Middleware intercepted event", internalEvent);
-        let origin: Origin | false = false;
         if (!result.isExternalRewrite) {
-          origin = await originResolver.resolve(result.internalEvent.rawPath);
+          const origin = await originResolver.resolve(
+            result.internalEvent.rawPath,
+          );
           return {
             type: "middleware",
             internalEvent: result.internalEvent,
--- a/packages/open-next/src/adapters/middleware.ts
+++ b/packages/open-next/src/adapters/middleware.ts
@@ -1,7 +1,7 @@
 import type { InternalEvent, Origin } from "types/open-next";
 import { runWithOpenNextRequestContext } from "utils/promise";
 
-import { debug } from "../adapters/logger";
+import { debug, error } from "../adapters/logger";
 import { createGenericHandler } from "../core/createGenericHandler";
 import {
   resolveIncrementalCache,
@@ -57,7 +57,23 @@ const defaultHandler = async (internalEvent: InternalEvent) => {
             isISR: result.isISR,
           };
         }
-        return externalRequestProxy.proxy(result.internalEvent);
+        try {
+          return externalRequestProxy.proxy(result.internalEvent);
+        } catch (e) {
+          error("External request failed.", e);
+          return {
+            type: "middleware",
+            internalEvent: {
+              ...result.internalEvent,
+              rawPath: "/500",
+              url: "/500",
+              method: "GET",
+            },
+            isExternalRewrite: result.isExternalRewrite,
+            origin: false,
+            isISR: result.isISR,
+          };
+        }
       }
 
       debug("Middleware response", result);
--- a/packages/open-next/src/adapters/middleware.ts
+++ b/packages/open-next/src/adapters/middleware.ts
@@ -69,7 +69,8 @@ const defaultHandler = async (internalEvent: InternalEvent) => {
               url: "/500",
               method: "GET",
             },
-            isExternalRewrite: result.isExternalRewrite,
+            // On error we need to rewrite to the 500 page which is an internal rewrite
+            isExternalRewrite: false,
             origin: false,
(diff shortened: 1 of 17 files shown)


packages/open-next/src/adapters/middleware.ts
```diff
@@ -43,14 +48,15 @@ const defaultHandler = async (internalEvent: InternalEvent) => {
         let origin: Origin | false = false;
         if (!result.isExternalRewrite) {
           origin = await originResolver.resolve(result.internalEvent.rawPath);
```
vicb, line 50: nit: inline origin l 55

packages/open-next/src/core/requestHandler.ts
```diff
@@ -65,6 +66,35 @@ export async function openNextHandler(
         delete headers[rawKey];
       }
 
+      if (
+        "isExternalRewrite" in preprocessResult &&
+        preprocessResult.isExternalRewrite
+      ) {
+        try {
+          const proxyResult = await globalThis.proxyExternalRequest.proxy(
```
vicb, line 74: nit: drop the extra local var? `preprocessResult = ...`

packages/open-next/src/overrides/proxyExternalRequest/fetch.ts
```diff
@@ -0,0 +1,28 @@
+import type { ProxyExternalRequest } from "types/overrides";
+import { emptyReadableStream } from "utils/stream";
+
+const fetchProxy: ProxyExternalRequest = {
+  name: "fetch-proxy",
+  // @ts-ignore
+  proxy: async (internalEvent) => {
+    const { url, headers, method, body } = internalEvent;
+    const response = await fetch(url, {
```
vicb, line 9: do we need to try/catch?

packages/open-next/src/types/open-next.ts
```diff
@@ -102,6 +103,8 @@ export type IncludedOriginResolver = "pattern-env";
 
 export type IncludedWarmer = "aws-lambda";
 
+export type IncludedProxyExternalRequest = "node" | "fetch";
```
vicb, line 106: add "dummy" (it's also missing from other types

packages/open-next/src/build/edge/createEdgeBundle.ts
```diff
@@ -23,7 +26,11 @@ interface BuildEdgeBundleOptions {
   entrypoint: string;
   outfile: string;
   options: BuildOptions;
-  overrides?: OverrideOptions;
+  overrides?: OverrideOptions & {
+    originResolver?:
```
vicb, line 30: Is it used there?

vicb (commented):
Super quick review because I'm busy again today.
Will do a deeper review on Monday

packages/open-next/src/overrides/proxyExternalRequest/fetch.ts
conico974: I don't think we want to do it here. If something goes wrong, we want to return the user 500 route, and it would require changing the type (or fetching the 500 directly) and doing it for every proxy
There is one try/catch in `requestHandler`, and we should add one in `middleware.ts` as well

packages/open-next/src/build/edge/createEdgeBundle.ts
conico974: Yeah, if you define a custom `originResolver` in the config for the external middleware, it will replace the default one with the dummy in the final bundle (since it's not used)

packages/open-next/src/core/resolve.ts
```diff
@@ -124,7 +125,7 @@ export async function resolveOriginResolver(
  * @__PURE__
  */
 export async function resolveWarmerInvoke(
-  warmer?: LazyLoadedOverride<Warmer> | "aws-lambda",
+  warmer?: LazyLoadedOverride<Warmer> | IncludedWarmer,
```
vicb, line 128: you update l 115 in a similar way `string` -> `IncludedOriginResolver`

vicb: l 101 as well.
Maybe use the same as 86

packages/open-next/src/build/edge/createEdgeBundle.ts
vicb: But l68+ should be update too, no?

conico974: Ho right, i missed it

packages/open-next/src/core/requestHandler.ts
```diff
@@ -65,6 +66,34 @@ export async function openNextHandler(
         delete headers[rawKey];
       }
 
+      if (
+        "isExternalRewrite" in preprocessResult &&
+        preprocessResult.isExternalRewrite
```
vicb, line 71:
```suggestion
        preprocessResult.isExternalRewrite === true
```

packages/open-next/src/core/requestHandler.ts
```diff
@@ -65,6 +66,34 @@ export async function openNextHandler(
...
+              type: "core",
+              rawPath: "/500",
+              method: "GET",
+              headers: {},
+              url: "/500",
+              query: {},
+              cookies: {},
+              remoteAddress: "",
+            },
+            isExternalRewrite: false,
```
vicb, line 90:
```suggestion
            isExternalRewrite: true,
```

vicb (approved):
I added a few comments, LGTM.

packages/open-next/src/adapters/middleware.ts
```diff
@@ -40,17 +45,35 @@ const defaultHandler = async (internalEvent: InternalEvent) => {
...
+          error("External request failed.", e);
+          return {
+            type: "middleware",
+            internalEvent: {
+              ...result.internalEvent,
+              rawPath: "/500",
+              url: "/500",
+              method: "GET",
+            },
+            isExternalRewrite: result.isExternalRewrite,
```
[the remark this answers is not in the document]
conico974: The value here doesn't matter that much (in this case it will not be used), but actually it should be false. The `/500` path is an internal one

packages/open-next/src/core/requestHandler.ts
conico974: Same as in `middleware.ts` this rewrite is internal (even though initially it was attempting an external rewrite)

packages/open-next/src/core/requestHandler.ts
conico974: Ho actually, this one cannot work, `preprocessResult` can be either an `InternalResult` or a `MiddlewareOutputEvent`

packages/open-next/src/core/requestHandler.ts
vicb: @conico974 Thanks for the feedback. What about adding a comment in both locations?

packages/open-next/src/plugins/resolve.ts
```diff
@@ -26,6 +32,9 @@ export interface IPluginSettings {
       | LazyLoadedOverride<OriginResolver>
       | IncludedOriginResolver;
     warmer?: LazyLoadedOverride<Warmer> | IncludedWarmer;
+    proxyExternalRequest?:
```
vicb, line 35: nit: 
- use `OverrideOptions["..."]` for `warmer` and `proxyExternalRequest`
- `DefaultOverrideOptions<any, any>` can also be changed for `OverrideOptions`, right?

packages/open-next/src/build/edge/createEdgeBundle.ts
```diff
@@ -84,6 +91,10 @@ export async function buildEdgeBundle({
                       : "sqs-lite",
                 }
               : {}),
+            originResolver:
```
vicb, line 94: add `proxyExternalRequest`

packages/open-next/src/plugins/resolve.ts
conico974: > DefaultOverrideOptions<any, any> can also be changed for OverrideOptions, right?

Yes

> OverrideOptions["..."] for warmer and proxyExternalRequest

We can't for the warmer, it is not defined in OverrideOptions

vicb: Right I mixed with wrapper 🙃

vicb:
@conico974 can we merge this?

conico974:
@vicb Yeah if everything is ok for the fetch/cloudflare part

vicb:
> @vicb Yeah if everything is ok for the fetch/cloudflare part

It looks okay, I'll test this more thoroughly soon and fix any thing that would need to be fixed.

Does that sound ok?

PR #632, opened 2024-11-16
fix 304 incorrectly set as 200

### LINKED ISSUE #631: Conditional GET requests don't appear to be handled correctly
When [etags](https://nextjs.org/docs/app/api-reference/next-config-js/generateEtags) are enabled in Next.js, conditional GET requests to an OpenNext lambda with an etag matching the page generated/cached by the server and considered fresh seem to result in a status 200 response with an empty body rather than the expected 304.

As far as I can tell, the key interaction is with [this behaviour](https://github.com/vercel/next.js/blob/0f9f442b7bcf028f26e6095c68a5f57c31356041/packages/next/src/server/send-payload.ts#L26-L30) in the Next server when sending the 304 response:

```js
function sendEtagResponse(req, res, etag) {
  ...
  if (fresh(req.headers, { etag })) {
    res.statusCode = 304
    res.end()
    return true
  }
  ...
}
```

and [this block](https://github.com/vercel/next.js/blob/0f9f442b7bcf028f26e6095c68a5f57c31356041/packages/next/src/server/base-server.ts#L1702-L1722) which captures the status code before the response is sent, and restores it afterwards. 

```js
    const originalStatus = res.statusCode
    ...
    if (!res.sent) {
      await this.sendRenderResult(req, res, {...}) // calls through to sendEtagResponse 
      res.statusCode = originalStatus
    }
```

The commit that added this (https://github.com/vercel/next.js/commit/74153e1208e0f3cab8fa736e9192fd3ff7ad12cc) is explicitly aiming to make sure that the 304 status code is hidden from the rest of the server so that it doesn't itself get cached.

However the response object supplied by OpenNext is not behaving the same way as a normal node one - when `res.end()` is called on a `ServerResponse`, the current value of `statusCode` is committed immediately, while with the [`OpenNextNodeResponse`](https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/http/openNextResponse.ts#L199) `flushHeaders` is only called once the stream `"finish"` event has been emitted asynchronously, by which time the status code has been reset back to 200 (or whatever other value it was previously). The status then returned to the lambda caller is incorrectly a 200 with an empty body, which may then get cached by CloudFront or similar, exacerbating the issue.

Not entirely sure what the right fix is here - the Next behaviour is a bit suspect, but does ultimately work with their Node server, and presumably it is necessary to allow the status code to be reset from 304 on the response to make sure it doesn't itself get cached. Maybe the `OpenNextNodeResponse` can internally snapshot `statusCode` at the time the response should be committed and use that for the lambda status, but still allow it to be mutated?

### Reproduction

* Deploy a Next.js app router application with `generateEtags: true` config and a SSG route to lambda.
* Make a request to that route to get an etag header value. The body should be returned as expected.
* Make a second request with `if-none-match` set to that etag value. This request will return 200 with no body.

In case its relevant, I've been testing with the `aws-lambda-streaming` wrapper enabled and the `nodejs20.x` lambda runtime.

conico974:
Fix #631

### DIFF
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -312,7 +312,7 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
     //BE CAREFUL: Aws keeps rolling out broken streaming implementations even on accounts that had working ones before
     //This is not dependent on the node runtime used
     if (
-      bodyLength < 1 &&
+      bodyLength === 0 &&
       // We use an env variable here because not all aws account have the same behavior
       // On some aws accounts the response will hang if the body is empty
       // We are modifying the response body here, this is not a good practice
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -78,23 +78,11 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
 
   constructor(
     private fixHeaders: (headers: OutgoingHttpHeaders) => void,
-    onEnd: (headers: OutgoingHttpHeaders) => Promise<void>,
+    private onEnd: (headers: OutgoingHttpHeaders) => Promise<void>,
     private streamCreator?: StreamCreator,
     private initialHeaders?: OutgoingHttpHeaders,
   ) {
     super();
-    this.once("finish", () => {
-      if (!this.headersSent) {
-        this.flushHeaders();
-      }
-      // In some cases we might not have a store i.e. for example in the image optimization function
-      // We may want to reconsider this in the future, it might be intersting to have access to this store everywhere
-      globalThis.__openNextAls
-        ?.getStore()
-        ?.pendingPromiseRunner.add(onEnd(this.headers));
-      const bodyLength = this.getBody().length;
-      this.streamCreator?.onFinish(bodyLength);
-    });
   }
 
   // Necessary for next 12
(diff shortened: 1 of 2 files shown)


packages/open-next/src/http/openNextResponse.ts
```diff
@@ -305,16 +293,26 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
+
+    //This is only here because of aws broken streaming implementation.
+    //Hopefully one day they will be able to give us a working streaming implementation in lambda for everyone
+    //If you're lucky you have a working streaming implementation in your aws account and don't need this
+    //If not you can set the OPEN_NEXT_FORCE_NON_EMPTY_RESPONSE env variable to true
+    //BE CAREFUL: Aws keeps rolling out broken streaming implementations even on accounts that had working ones before
+    //This is not dependent on the node runtime used
     if (
-      this.getBody().length < 1 &&
+      bodyLength < 1 &&
```
vicb, line 315:
```suggestion
      bodyLength === 0 &&
```

vicb (commented):
What about your comment on the issue:

> Yeah unfortunately the way next does thing cause a race condition. There is no good way to do this.
> The fix works well with streaming wrapper, but not with non streaming one because by the time the request is done and ready to be sent, next has already reset it.
And not applying the status code Next sets will cause the initial issue in Next appear again.
>I'll try to open an issue in the Next repo when I got some time, the fix should be done upstream.

Would there be a way to add test?

conico974:
> Would there be a way to add test?

Not with our current setup, as i said in a previous comment in the issue, this does not happen with the default cloudfront policy (which is what we use).
In order to test this properly we would have to run OpenNext locally without cloudfront ( which is probably what we should implement at some point so that we could run more e2e test more easily )

conico974:
@vicb unless there is something you'd want changed, i'm gonna merge this one and make a release

vicb:
Great, thanks!

PR #633, opened 2024-11-18
fix(assets): check if favicon.ico is a file

chanceaclark:
## What/why?

In the case a Next.js application has a route handler for the `favicon.ico` path, we need to check to see if it's a file before copying it over to the output directory.

Example:
```ts
// /app/favicon.ico/route.ts
export const GET = async () => {
  const favicon = await fetch('https://somecms.com').then((res) => res.arrayBuffer());

  return new Response(favicon, {
    headers: {
      'Content-Type': 'image/x-icon',
    },
  });
};

export const dynamic = 'force-static';
```

### DIFF
--- a/packages/open-next/src/build/createAssets.ts
+++ b/packages/open-next/src/build/createAssets.ts
@@ -44,7 +44,9 @@ export function createStaticAssets(options: buildHelper.BuildOptions) {
 
   const faviconPath = path.join(appPath, appSrcPath, "favicon.ico");
 
-  if (fs.existsSync(faviconPath)) {
+  // We need to check if the favicon is either a file or directory.
+  // If it's a directory, we assume it's a route handler and ignore it.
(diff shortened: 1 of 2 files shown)


packages/open-next/src/build/createAssets.ts
```diff
@@ -44,7 +44,7 @@ export function createStaticAssets(options: buildHelper.BuildOptions) {
 
   const faviconPath = path.join(appPath, appSrcPath, "favicon.ico");
 
-  if (fs.existsSync(faviconPath)) {
+  if (fs.existsSync(faviconPath) && fs.lstatSync(faviconPath).isFile()) {
```
vicb, line 47: +1 for the change.
Maybe add a comment why this is needed.

chanceaclark: Done!

vicb (approved):

PR #634, opened 2024-11-19
refactor: minor updates

vicb:


packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -203,25 +202,36 @@ File ${fullFilePath} does not exist
...
   // We then need to copy all the files at the root of server
 
   mkdirSync(path.join(outputNextDir, "server"), { recursive: true });
 
-  readdirSync(path.join(standaloneNextDir, "server")).forEach((f) => {
-    if (statSync(path.join(standaloneNextDir, "server", f)).isDirectory())
-      return;
-    if (f !== "server.js") {
+  readdirSync(path.join(standaloneNextDir, "server"))
+    .filter(
```
conico974, line 222: We could do it in 1 filter instead of 2, it avoids having to iterate over everything twice

conico974 (approved):
Just a little comment, other than that LGTM

PR #635, opened 2024-11-19
fix(cloudflare): bind waitUntil to ctx

vicb:
Fix the error "TypeError: Illegal invocation: function called with incorrect `this` reference."


conico974 (approved):
LGTM Thanks

PR #637, opened 2024-11-20
add: pnpm catalog

sommeeeer:
This PR will add [pnpm catalog](https://pnpm.io/catalogs) for deps that is reused through the repo. Making it easier to maintain. I will open this in draft mode so we can continue working on it. The alternative was using something like [syncpack](https://jamiemason.github.io/syncpack/guide/getting-started/), but that would add another dep so I decided to go with pnpm catalogs instead.

Any suggestions on what more we could add? I have considered using catalog for the other shared deps in `examples/*`, like f.ex `tailwindcss`/`typescript` or `postcss`. Not sure if we want that, and also should we consider using a caret or tilde in any of the catalog versions? 

Btw, I have tested building and deploying the examples apps with this PR, and it works as it should.

### DIFF
--- a/examples/app-pages-router/package.json
+++ b/examples/app-pages-router/package.json
@@ -21,9 +21,9 @@
     "@types/node": "catalog:",
     "@types/react": "catalog:",
     "@types/react-dom": "catalog:",
-    "autoprefixer": "10.4.15",
-    "postcss": "8.4.27",
-    "tailwindcss": "3.3.3",
-    "typescript": "^4.9.3"
+    "autoprefixer": "catalog:",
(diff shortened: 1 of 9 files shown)


conico974:
@sommeeeer Yeah we should add these 3 deps and bump typescript version as well.
And we don't want to use `^` or `~`, we should always use exact where we can

sommeeeer:
> Yeah we should add these 3 deps and bump typescript version as well.

i didnt add catalog to the `open-next` package, could however use it for `@types/node` and `typescript`. not sure what we want?

conico974 (approved):
LGTM Thanks

PR #638, opened 2024-11-21
fix(http): Set content-length only if body is present

vicb:


packages/open-next/src/http/request.ts
```diff
@@ -6,19 +6,15 @@
...
 
 export class IncomingMessage extends http.IncomingMessage {
-  constructor({
-    method,
-    url,
-    headers,
-    body,
-    remoteAddress,
-  }: {
+  constructor(request: {
```
vicb, line 9: I had to name the parameter because of a formatting issue.

```
    // Set the content length when there is a body.
    // See https://httpwg.org/specs/rfc9110.html#field.content-length
    if (body) {
      headers["content-length"] ??= String(Buffer.byteLength(body));
    }
```

When saving the file with the previous structure the line `if (body) {` would be deleted otherwise 🤨
This would not happen if the comments are moved after the if.
Weird bug but still happens after ESLint and Biome extensions are de-activated in VSCode.

🤷

conico974: It's probably worth to add a comment explaining this

conico974 (approved):
Just a little comment, otherwise LGTM Thanks

packages/open-next/src/http/request.ts
vicb: Well... amending the PR to revert the change worked.

To prove I'm not completely 🤪, here is a video when I start from the original file:

https://github.com/user-attachments/assets/2446b208-084c-47b3-bab6-931400c36ac6

sommeeeer: > To prove I'm not completely 🤪, here is a video when I start from the original file:

does it happen if you use `/* ... */`?

vicb: Yep, tried that as well as a few different workarounds and it results in the same weird bug.

PR #641, opened 2024-11-22
fix(cache): make dummy cache error ignorable

vicb:
All "dummy" cache methods throw by design.
This PR makes the corresponding errors ignorable so that the logs are less verbose and always use the DEBUG level.

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -1,3 +1,4 @@
+import { IgnorableError, isIgnorableError } from "utils/error";
 import { isBinaryContentType } from "./binary";
 import { debug, error, warn } from "./logger";
 
@@ -192,7 +193,11 @@ export default class S3Cache {
       } as CacheHandlerValue;
     } catch (e) {
       // We can usually ignore errors here as they are usually due to cache not being found
-      debug("Failed to get fetch cache", e);
+      if (isIgnorableError(e)) {
+        debug("Failed to get fetch cache", e.message);
+      } else {
+        debug("Failed to get fetch cache", e);
+      }
       return null;
     }
   }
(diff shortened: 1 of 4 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -192,7 +193,11 @@ export default class S3Cache {
       } as CacheHandlerValue;
     } catch (e) {
       // We can usually ignore errors here as they are usually due to cache not being found
-      debug("Failed to get fetch cache", e);
+      if (isIgnorableError(e)) {
```
conico974, line 196: Do we want to do this here, `error` should already be handling this https://github.com/opennextjs/opennextjs-aws/blob/4a500eeb609e1ed3be97d744e5d762d65dafd883/packages/open-next/src/adapters/logger.ts#L43-L67
Having only the `error` function deciding which error to log is probably better (and easier to read)

conico974 (commented):
We may want to introduce another env variable `OPEN_NEXT_ERROR_LOG_LEVEL` to decide if we log the `IgnorableError` even in debug. 
All of this would be decided inside the error function
By default it would only log level 1 error and above

packages/open-next/src/adapters/cache.ts
conico974: I think `IgnorableError` alone are enough in most cases, debug mode should never be used outside of when you debug so you should never see such errors in production.
The `OPEN_NEXT_ERROR_LOG_LEVEL` has 1 main use imo, it's for adding the ability to mask `RecoverableError` logs, but it could also be used to mask `IgnorableError` in debug mode, which could be useful for us when testing something that has nothing to do with these errors

vicb:
See https://github.com/opennextjs/opennextjs-aws/pull/649

PR #644, opened 2024-11-25
fix(cloudflare): cloudflare-streaming do not use the edge runtime

vicb:
Discussed at https://github.com/opennextjs/opennextjs-aws/pull/642#discussion_r1853958888
I'll add the validation in a follow-up PR

### DIFF
--- a/packages/open-next/src/overrides/wrappers/cloudflare.ts
+++ b/packages/open-next/src/overrides/wrappers/cloudflare.ts
@@ -17,7 +17,7 @@ interface WorkerContext {
 
 const handler: WrapperHandler<
   InternalEvent,
-  InternalResult | ({ type: "middleware" } & MiddlewareOutputEvent)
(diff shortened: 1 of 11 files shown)


packages/open-next/src/overrides/wrappers/cloudflare.ts
```diff
@@ -65,4 +65,7 @@ export default {
   name: "cloudflare",
   supportStreaming: true,
   edgeRuntime: true,
-};
+} satisfies Wrapper<
+  InternalEvent,
+  InternalResult | ({ type: "middleware" } & MiddlewareOutputEvent)
```
conico974, line 70: We should probably add `type:middleware` to the `MiddlewareOutputEvent` directly

conico974 (approved):
Just a little comment, but other than that LGTM

vicb:
@conico974 PTAL, I have tried to clarify the types

PR #645, opened 2024-11-25
refactor(cloudflare): rename the "cloudflare" wrapper to "cloudflare-edge"

vicb:
The commit message do not mention "cloudflare-streaming" -> "cloudflare-node" as this was never released officially.

### DIFF
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -5,7 +5,10 @@ import url from "node:url";
 
 import type { BuildOptions as ESBuildOptions } from "esbuild";
 import { build as buildAsync, buildSync } from "esbuild";
-import type { OpenNextConfig } from "types/open-next.js";
+import type {
+  DefaultOverrideOptions,
+  OpenNextConfig,
+} from "types/open-next.js";
 
 import logger from "../logger.js";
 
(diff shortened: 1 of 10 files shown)


packages/open-next/src/build/helper.ts
```diff
@@ -356,3 +359,18 @@ export function initOutputDir(options: BuildOptions) {
...
+ * @returns Whether the edge runtime is used
+ */
+export async function isEdgeRuntime(
+  overrides: DefaultOverrideOptions | undefined,
+) {
+  if (!overrides?.wrapper) {
+    return false;
+  }
+  if (typeof overrides.wrapper === "string") {
+    return overrides.wrapper.startsWith("cloudflare");
```
conico974, line 373: This is weird, `cloudflare-node` does run in an `edge` like runtime but is meant for the `node` runtime of next.

conico974 (commented):
I think we should change the `Wrapper` type so that it is clearer (and would allow for better validation). It could become 
```ts
export type Wrapper<
  E extends BaseEventOrResult = InternalEvent,
  R extends BaseEventOrResult = InternalResult,
> = BaseOverride & {
  wrapper: WrapperHandler<E, R>;
  supportStreaming: boolean;
  supportExternalMiddleware?: boolean
  edgeLikeRuntime?: boolean;
};
```

`edgeLikeRuntime` would be used at build time to apply the necessary fix for cloudflare. It would also help distinguish this from Next `edge` runtime. This would be set for both cloudflare wrappers.
I don't like this name a lot, maybe you can find something better. Maybe something with Worker or WinterCG 

`supportExternalMiddleware` would only be used for validation because external middleware need to be able to handle both `InternalResult` and `MiddlewareOutputEvent`. This one would be set to true for the `aws-cloudfront` wrapper as well.

All of this could be done in a different PR if you prefer.

vicb:
> All of this could be done in a different PR if you prefer.

Yes, I would prefer a different PR as I think we probably need to refactor the code.
See https://github.com/opennextjs/opennextjs-aws/issues/646

packages/open-next/src/build/helper.ts
vicb: fixed

conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review!

PR #649, opened 2024-11-26
error

vicb:


packages/open-next/src/adapters/logger.ts
```diff
@@ -42,28 +42,33 @@ const isDownplayedErrorLog = (errorLog: AwsSdkClientCommandErrorLog) =>
...
-    if (error.logLevel === 0) {
-      debug(...args);
+    const error = args.find((arg) => isOpenNextError(arg))!;
+    if (error.logLevel < getOpenNextErrorLogLevel()) {
       return;
     }
+    if (error.logLevel === 0) {
+      // Display the name and the message instead of full Open Next errors.
+      return debug(
+        ...args.map((arg) =>
```
vicb, line 56: @conico974 the code does not display the full exception here and for level 1. Let me know what you think about that.

conico974: That's good, as long as we have the message it's fine

packages/open-next/src/adapters/logger.ts
```diff
@@ -42,28 +42,33 @@ const isDownplayedErrorLog = (errorLog: AwsSdkClientCommandErrorLog) =>
...
-    ) as BaseOpenNextError;
-    if (error.logLevel === 0) {
-      debug(...args);
+    const error = args.find((arg) => isOpenNextError(arg))!;
+    if (error.logLevel < getOpenNextErrorLogLevel()) {
       return;
     }
+    if (error.logLevel === 0) {
+      // Display the name and the message instead of full Open Next errors.
+      return debug(
```
conico974, line 55: We should probably add some kind of warning just before this line in case `openNextDebug` is not set, otherwise there will be no error logged and it might be confusing for people.
Or we just use `console.log` or `console.error` and rely entirely on `OPEN_NEXT_ERROR_LOG_LEVEL`

conico974 (approved):
One little thing, but otherwise LGTM

packages/open-next/src/adapters/logger.ts
vicb: Good idea.
+1 to drop openNextDebug

conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review!

packages/open-next/src/adapters/logger.ts
```diff
@@ -1,4 +1,4 @@
-import type { BaseOpenNextError } from "utils/error";
+import { type BaseOpenNextError, isOpenNextError } from "utils/error";
```
sommeeeer, line 1: that `BaseOpenNextError` is not used anywhere?

PR #651, opened 2024-11-26
fix: type of the edge converter

vicb:
I had missed the edge converter in https://github.com/opennextjs/opennextjs-aws/pull/644


conico974 (approved):
LGTM Thanks

PR #655, opened 2024-11-27
feat(build): add BuildOptions#minify to enable code minification

vicb:
Default to !BuildOptions#debug

This will be useful for cloudflare as we don't want to minify the code to be able to apply string replace patches (see [this issue](https://github.com/opennextjs/opennextjs-cloudflare/issues/143))

With this change, we can decide not to minify the code whether or not debug is enabled (by overriding the BuildOptions).

Notes:
- `wrangler` will ultimately minify the code anyway
- we plan to migrate the patches to AST transforms to not depend on whether or not the code is minified


vicb:
Thanks for the review!

PR #656, opened 2024-11-28
Add override for esbuild plugin

conico974:


packages/open-next/src/core/util.ts
```diff
@@ -28,7 +28,11 @@ import {
 overrideNextjsRequireHooks(NextConfig);
 applyNextjsRequireHooksOverride();
 //#endOverride
+
+//#override cacheHandlerPath
 const cacheHandlerPath = require.resolve("./cache.cjs");
```
conico974, line 33: After some thought, i don't see how setting the `NEXT_CACHE_HANDLER_PATH` env would benefit us that much, at least not here. 
The only way this would be useful would be at build time so that the cache is available during `next build`, but then it would mean that we'll have to either include open-next in the bundle (or at least a precompiled version of the cache in the `node_modules`) or still override it at build time like we do today.
It also means that both the tag and incremental cache would need to be available at build time

packages/open-next/src/core/util.ts
```diff
@@ -28,7 +28,11 @@ import {
 overrideNextjsRequireHooks(NextConfig);
```
conico974, line 28: @vicb These 2 uses `require.resolve` and are present on next >=13.4.13 and next <=13.5.1

conico974:
@vicb PTAL, with these changes you should be able to just add these 2 overrides in the deletes part of the replacement plugin and remove the OpenNext patch

packages/open-next/src/core/requestHandler.ts
```diff
@@ -19,7 +19,9 @@ import { requestHandler, setNextjsPrebundledReact } from "./util";
 // This is used to identify requests in the cache
 globalThis.__openNextAls = new AsyncLocalStorage();
 
+//#override patchAsyncStorage
```
vicb, line 22: Long term, we should not delete it but have a cloudflare specific impl.
Maybe we could then override instead of deleting but at the same time the temporary patch in cloudflare is good enough before we have a long term solution.

WDYT?

packages/open-next/src/core/util.ts
vicb: Here again we can not delete this but we should override or it will break the build.

vicb (commented):
I think there is no rush to merge this before we give more thoughts about how to implement things into cf.
+1 to keep this open as a reminder though.

packages/open-next/src/core/requestHandler.ts
conico974: This one will hopefully be removed in the future in OpenNext as well.
Once this PR https://github.com/vercel/next.js/pull/72082 gets merged it will not be necessary anymore.
I just have to figure out how to test with the new `use cache` stuff from next

conico974:
> I think there is no rush to merge this before we give more thoughts about how to implement things into cf.
+1 to keep this open as a reminder though.

I'll change it to draft then

PR #659, opened 2024-12-02
refactor: remove unused imports

sommeeeer:
Should we consider turning on this [no-unused-imports](https://biomejs.dev/linter/rules/no-unused-imports/) in `biome.jsonc`?

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -33,6 +33,9 @@
       },
       "performance": {
         "noDelete": "off"
+      },
+      "correctness": {
+        "noUnusedImports": "error"
       }
     }
   },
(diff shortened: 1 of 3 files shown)


vicb:
Oops, sorry that I haven't seen the PR earlier, I think I fixed a bunch of that already

sommeeeer:
> Oops, sorry that I haven't seen the PR earlier, I think I fixed a bunch of that already

i saw that, i guess we can close this. should we consider adding that rule to `biome`? what level would you prefer?

vicb:
> should we consider adding that rule to biome? what level would you prefer?

I was thinking about that when fixing the imports. I think error is good.

vicb (approved):
Thanks!

PR #660, opened 2024-12-02
fix: partially reverts 644

vicb:


packages/open-next/src/core/routingHandler.ts
```diff
@@ -20,14 +25,17 @@ import { handleMiddleware } from "./routing/middleware";
 export const MIDDLEWARE_HEADER_PREFIX = "x-middleware-response-";
 export const MIDDLEWARE_HEADER_PREFIX_LEN = MIDDLEWARE_HEADER_PREFIX.length;
 
-export interface MiddlewareOutputEvent {
-  type: "middleware";
+export interface RoutingResult {
```
conico974, line 28: I'm wondering if it would not make sense to define both `RoutingResult` and `MiddlewareResult` in `types/open-next` instead

conico974 (approved):
LGTM Thanks

packages/open-next/src/core/routingHandler.ts
vicb: Great idea, added a commit.
Thanks!

conico974 (commented):
It is failing to build https://github.com/opennextjs/opennextjs-aws/actions/runs/12135452530/job/33834536340?pr=660

vicb:
fixed!
Thanks @conico974 for all the reviews this morning 🙏

PR #661, opened 2024-12-03
fix(cloudflare): fix the node wrapper

vicb:
Returns the response converter from the internal event. 
It fixes a "This ReadableStream did not return bytes" workerd error.

### DIFF
--- a/packages/open-next/src/overrides/wrappers/cloudflare-node.ts
+++ b/packages/open-next/src/overrides/wrappers/cloudflare-node.ts
@@ -31,9 +31,6 @@ const handler: WrapperHandler<InternalEvent, InternalResult> =
     const url = new URL(internalEvent.url);
     (internalEvent.url as string) = url.href.slice(url.origin.length);
 
-    const { promise: promiseResponse, resolve: resolveResponse } =
-      Promise.withResolvers<Response>();
-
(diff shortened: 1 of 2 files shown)


packages/open-next/src/overrides/wrappers/cloudflare-node.ts
```diff
@@ -52,17 +49,15 @@ const handler: WrapperHandler<InternalEvent, InternalResult> =
...
         return Writable.fromWeb(writable);
       },
       onWrite: () => {},
       onFinish: (_length: number) => {},
     };
 
-    ctx.waitUntil(handler(internalEvent, streamCreator));
-
-    return promiseResponse;
+    const internalResult = await handler(internalEvent, streamCreator);
```
conico974, line 59: It may fix it, but it also means that streaming will not work.
The handler in node returns only after the response is fully sent (and the streamCreator is not useful in this case)

Maybe using a custom `Writable` with a custom `ReadableStream` would help figure out what's the issue here (or resolving the response in `onWrite` (this may break other things, but at least it'll help figure out what's going on here)

vicb:
superseded by https://github.com/opennextjs/opennextjs-aws/pull/664

PR #663, opened 2024-12-03
refactor: move StreamCreator to types/open-next

vicb:
And make onFinish optional

### DIFF
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -8,22 +8,12 @@ import type { Socket } from "node:net";
 import type { TransformCallback, Writable } from "node:stream";
 import { Transform } from "node:stream";
 
+import type { StreamCreator } from "types/open-next";
 import { debug } from "../adapters/logger";
 import { parseCookies, parseHeaders } from "./util";
 
 const SET_COOKIE_HEADER = "set-cookie";
 const CANNOT_BE_USED = "This cannot be used in OpenNext";
-
-export interface StreamCreator {
-  writeHeaders(prelude: {
-    statusCode: number;
-    cookies: string[];
-    headers: Record<string, string>;
-  }): Writable;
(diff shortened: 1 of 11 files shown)


packages/open-next/src/http/openNextResponse.ts
```diff
@@ -92,9 +82,7 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
   }
 
   get finished() {
-    return Boolean(
-      this.writableFinished && this.responseStream?.writableFinished,
-    );
+    return this.writableFinished && (this.responseStream?.writableFinished ?? true);
```
vicb, line 85: This is a change in behavior but I believe this is what we want?

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -92,8 +82,8 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
   }
 
   get finished() {
-    return Boolean(
-      this.writableFinished && this.responseStream?.writableFinished,
+    return (
+      this.writableFinished && (this.responseStream?.writableFinished ?? true)
```
conico974, line 86: I'm not sure this one is correct actually. 
Your fix is good in case we don't provide a `ResponseStream`, but in case we do, we may end up in a situation where `OpenNextNodeResponse` is finished, but not `this.responseStream`. I'm not 100% sure of that though and not sure what could be the consequences either.

I think we could replace it with 
```ts
this.responseStream ? this.responseStream.writableFinished : this.writableFinished
```

conico974 (commented):
Not sure about the `finished()` change. The rest is good

packages/open-next/src/http/openNextResponse.ts
vicb: swapped for your suggestion, thanks!

conico974 (approved):
LGTM Thanks

PR #665, opened 2024-12-04
Feat Automatic cdn invalidation

### LINKED ISSUE #658: [Feature] Automatic CDN invalidation for On Demand Revalidation
We should provide a way to automatically invalidate the CDN.

It should be a custom override and the default should be to use a dummy one that will do nothing (Cloudfront automatic CDN invalidation could incur very high cost if not done properly)

conico974:
Fix #658 

~~Still a bit of work to do on this one, it will require a refactor of `InternalEvent` (or `RoutingResult`) and the routingHandler so that we have access to the information we need (Or maybe we can use the async hook context)~~

Should be ready for review

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -433,6 +433,25 @@ export default class Cache {
 
         // Update all keys with the given tag with revalidatedAt set to now
         await globalThis.tagCache.writeTags(toInsert);
+
+        // We can now invalidate all paths in the CDN
+        const uniquePaths = Array.from(
+          new Set(
+            toInsert
+              // We need to filter fetch cache key as they are not in the CDN
+              .filter((t) => t.tag.startsWith("_N_T_/"))
+              .map((t) => t.path),
+          ),
+        );
+        if (uniquePaths.length > 0) {
+          await globalThis.cdnInvalidationHandler.invalidatePaths(
+            uniquePaths.map((path) => ({
+              path,
+              // Here we can be sure that the path is from app router
+              isAppRouter: true,
+            })),
+          );
+        }
       }
     } catch (e) {
       error("Failed to revalidate tag", e);
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -447,9 +447,15 @@ export default class Cache {
         if (uniquePaths.length > 0) {
           await globalThis.cdnInvalidationHandler.invalidatePaths(
             uniquePaths.map((path) => ({
-              path,
-              // Here we can be sure that the path is from app router
-              isAppRouter: true,
+              initialPath: path,
+              rawPath: path,
+              resolvedRoutes: [
+                {
+                  route: path,
+                  // TODO: ideally here we should check if it's an app router page or route
+                  type: "app",
+                },
+              ],
             })),
           );
         }
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -435,6 +435,7 @@ export default class Cache {
         await globalThis.tagCache.writeTags(toInsert);
 
         // We can now invalidate all paths in the CDN
+        // This only applies to `revalidateTag`, not to `res.revalidate()`
         const uniquePaths = Array.from(
           new Set(
             toInsert
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -440,7 +440,7 @@ export default class Cache {
             toInsert
               // We need to filter fetch cache key as they are not in the CDN
               .filter((t) => t.tag.startsWith("_N_T_/"))
-              .map((t) => t.path),
+              .map((t) => `/${t.path}`),
           ),
         );
         if (uniquePaths.length > 0) {
--- a/packages/open-next/src/overrides/cdnInvalidation/cloudfront.ts
+++ b/packages/open-next/src/overrides/cdnInvalidation/cloudfront.ts
@@ -14,7 +14,7 @@ export default {
         ? [`${path}`, `${path}?_rsc=*`]
         : [
             `${path}`,
-            `_next/data/${process.env.BUILD_ID}${path === "/" ? "/index" : path}.json*`,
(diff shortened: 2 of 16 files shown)


packages/open-next/src/overrides/cdnInvalidation/cloudfront.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { CDNInvalidationHandler } from "types/overrides";
+
+const cloudfront = new CloudFrontClient({});
+export default {
+  name: "cloudfront",
+  invalidatePaths: async (paths) => {
+    //TODO: test the constructed paths
+    const constructedPaths = paths.flatMap(({ path, isAppRouter }) =>
+      isAppRouter
+        ? [`${path}`, `${path}?_rsc=*`]
```
sommeeeer, line 14: I think only [`${path}*`] is enough here? I will do some testing tommorow btw.

conico974: It's on purpose ` ${path}*` would invalidate every route that start with `path`. For example if you have a `products` and a `products/[slug]` route (using `${path}*`) it would  be impossible to invalidate only `products`

sommeeeer: I think it also invalidates the `products` if you use the `${path}*` one actually.

conico974: That's not what i'm saying, it invalidates `products` but also every single `products/[slug]` route, which may not be what you want

sommeeeer: but yeah im guessing we dont want that. its probably wise to only invalidate the path.

in my projects im just used to doing the star, cause normally you would want `products` and `products/[slug]` invalidated at the same time.

sommeeeer: However, after ive done some testing this morning the paths need to have a leading `/` else an error will be thrown: `Failed to revalidate tag InvalidArgument: Your request contains one or more invalid invalidation paths.`

sommeeeer: The paths need to have a leading `/` else an error will be thrown: `Failed to revalidate tag InvalidArgument: Your request contains one or more invalid invalidation paths.`

conico974: @sommeeeer It's only for `revalidateTag` right ? You didn't test `res.revalidate` ?

sommeeeer: > @sommeeeer It's only for `revalidateTag` right ? You didn't test `res.revalidate` ?
yeah, i will test `res.revalidate` now

sommeeeer: > It's only for `revalidateTag` right ? You didn't test `res.revalidate` ?

i tested `revalidatePath('/products');`

sommeeeer: This works:
```ts
const constructedPaths = paths.flatMap(({ path, isAppRouter }) =>
  isAppRouter
    ? [`/${path}`, `/${path}?_rsc=*`]
    : [
        `/${path}`,
        `/_next/data/${process.env.BUILD_ID}${path === "/" ? "/index" : path}.json*`,
      ],
);```

File: packages/open-next/src/adapters/cache.ts [hunk 2]
```diff
@@ -433,6 +433,32 @@ export default class Cache {
 
         // Update all keys with the given tag with revalidatedAt set to now
         await globalThis.tagCache.writeTags(toInsert);
+
+        // We can now invalidate all paths in the CDN
+        // This only applies to `revalidateTag`, not to `res.revalidate()`
+        const uniquePaths = Array.from(
```
vicb, line 439: nit: you don't need another array, you can use `Set#size` and `Set#values()`

packages/open-next/src/overrides/cdnInvalidation/cloudfront.ts
```diff
@@ -0,0 +1,37 @@
...
+  invalidatePaths: async (paths) => {
+    const constructedPaths = paths.flatMap(
+      ({ initialPath, resolvedRoutes }) => {
+        const isAppRouter = resolvedRoutes.some(
+          (route) => route.type === "app",
+        );
+        // revalidateTag doesn't have any leading slash, remove it just to be sure
+        const path = initialPath.replace(/^\//, "");
+        return isAppRouter
+          ? [`/${path}`, `/${path}?_rsc=*`]
```
vicb, line 19: style: I find nested ternary hard to read, not important

packages/tests-unit/tests/adapters/cache.test.ts
```diff
@@ -470,4 +477,72 @@ describe("CacheHandler", () => {
...
+      expect(tagCache.writeTags).not.toHaveBeenCalled();
+    });
+
+    it("Should call tagCache.writeTags", async () => {
+      globalThis.tagCache.getByTag.mockResolvedValueOnce(["/path"]);
+      await cache.revalidateTag("tag");
+
+      expect(globalThis.tagCache.getByTag).toHaveBeenCalledWith("tag");
+
+      expect(tagCache.writeTags).toHaveBeenCalled();
```
vicb, line 508: can this be removed?

vicb (approved):
LGTM, great PR!

packages/open-next/src/adapters/cache.ts
conico974: Yeah, but i find it easier to work with an array. 
`globalThis.cdnInvalidationHandler.invalidatePaths` expect to receive an array, not an iterable, so we'd also have to change the typing there

packages/tests-unit/tests/adapters/cache.test.ts
conico974: Yeah we could, it could probably be replaced with `toHaveBeenCalledTimes(1)` so that we ensure that it does not try to write multiple times

PR #666, opened 2024-12-06
chore: bump react and next in catalog

sommeeeer:
React 19 is stable now. I have successfully deployed the `examples` apps and ran the e2e test. Should work fine.


sommeeeer:
- Bumped next to `15.1.x` aswell
- Had to update the code as `after` is stable now
- `api/after/ssg` was cached by cloudfront so i added it to the condition in `middleware.ts`. 
- Ran the e2e test from my computer, should work fine now.

conico974 (approved):
LGTM Thanks a lot

PR #668, opened 2024-12-11
fix: ignore type checking in image-optimization.ts

sommeeeer:
This will make the build work again. We can safely ignore type checking in the entire file, cause its only used for Next <= `14.1`


conico974 (approved):
LGTM Thanks

PR #669, opened 2024-12-11
chore: add e2e for forbidden and unauthorized

sommeeeer:
- Add e2e test for `forbidden` and `unauthorized`
- Next will set `cache-control` to `s-maxage=31536000` on those routes, so i added 403 and 401 in `fixHeadersForError()` so they don't get cached


conico974 (commented):
We should probably wait a bit before merging this one, not sure what's the point of using `forbidden` or `unauthorized` on SSG route ( And they don't seem to work in middleware either )

PR #674, opened 2024-12-12
fix crash when not found is not there

conico974:
Should fix https://github.com/opennextjs/opennextjs-cloudflare/issues/167


dario-piotrowicz (approved):
Great, I can confirm that this does fix my issue 🥳 

Thanks a lot @conico974 😄

Screenshot 2024-12-12 at 18 27 01

PR #675, opened 2024-12-12
fix: import correct logger

sommeeeer:
There was an user in our Discord that complained about his lambda were outputting tons of logging, but he didn't have `OPEN_NEXT_DEBUG` set to any truthy value. Turns out that the `node.ts` module from the `proxyExternalRequest` override imported debug, and error from `node:console` instead of our own in `adapters/logger`.


conico974 (approved):
LGTM Thanks

PR #683, opened 2024-12-20
fix(cloudflare): No body for 101, 204, 205, or 304 status codes

vicb:
Fixes:

```
✘ [ERROR] TypeError: Response with null body status (101, 204, 205, or 304) cannot have a body.
```

For cached responses


conico974 (approved):
LGTM Thanks

vicb:
Thanks for the quick review!

PR #684, opened 2024-12-20
fix: fetch cache does not depend on tag cache being enabled

vicb:
That should help with https://github.com/opennextjs/opennextjs-cloudflare/pull/194

### DIFF
--- a/.changeset/many-cooks-breathe.md
+++ b/.changeset/many-cooks-breathe.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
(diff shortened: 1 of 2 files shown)


conico974 (approved):
LGTM Thanks.
FYI at the moment there is no way to manually invalidate a fetch cache entry without the tag cache

vicb:
Thanks for all the reviews today :)

vicb:
> FYI at the moment there is no way to manually invalidate a fetch cache entry without the tag cache

We will work on the tag cache next. Hopefully early next year.

Issue #624, opened 2024-11-08
Cannot find module './cjs/react.development.js'


conico974:
When doing function splitting with only app router component on lambda, this issue can happen.
This file is included for every app router route/component, including `_app` and `_document`, which are bundled everywhere except on app router only splitted function. (it should not be included btw)

The import is done before the `setNodeEnv` function call on lambda and so NODE_ENV is undefined and react try to load the dev stuff.

## FIX

There is a very easy fix for those running into this set the `NODE_ENV` env variable to `production`.
This should also be done on sst by default

asynchroza:
I'm using SST which calls open-next under the hood but it seems I cannot propagate the environment variable even when using `cross-env`. Do you have any clue what the issue here might be?

sommeeeer:
> I'm using SST which calls open-next under the hood but it seems I cannot propagate the environment variable even when using `cross-env`. Do you have any clue what the issue here might be?

whats in your `sst.aws.Nextjs` and are you using splitted server functions?

asynchroza:
> > I'm using SST which calls open-next under the hood but it seems I cannot propagate the environment variable even when using `cross-env`. Do you have any clue what the issue here might be?
> 
> whats in your `sst.aws.Nextjs` and are you using splitted server functions?

For anybody else who is using SST, after some discussions with the team @sommeeeer and @conico974, I got it running by updating my sst configuration as follows:

```ts
// eslint-disable-next-line @typescript-eslint/triple-slash-reference
/// <reference path="./.sst/platform/config.d.ts" />

export default $config({
    app(input) {
        return {
            name: 'pronto',
            removal: input?.stage === 'production' ? 'retain' : 'remove',
            protect: ['production'].includes(input?.stage),
            home: 'aws',
        };
    },
    async run() {
        new sst.aws.Nextjs('Core', {
            environment: {
                // ! @link https://github.com/opennextjs/opennextjs-aws/issues/624
                // TODO: Look into moving to a newer version when this is merged @link https://github.com/opennextjs/opennextjs-aws/pull/800
                NODE_ENV: 'production',
            },
        });
    },
}
```

conico974:
Fixed in https://github.com/opennextjs/opennextjs-aws/pull/800

PR #686, opened 2024-12-28
change `setNodeEnv` code to avoid warnings when using the Cloudflare adapter

dario-piotrowicz:
This is a quick change to avoid warnings such as:
Screenshot 2024-12-27 at 18 20 06
when using the Cloudflare adapter

I am not a huge fan of this change but, for a quick fix which doesn't require too much refactoring in the aws package, I think it's either this or patching the aws code on the Cloudflare adapter (as I've done in https://github.com/opennextjs/opennextjs-cloudflare/pull/211).

(A more proper fix would be to make the aws code configurable so that when used by the Cloudflare adapter the `setNodeEnv` is either never called, thus three shaken away, or becomes a no-op function, but that as I referred to before would be a much bigger change, which might not be worth it right now)

### DIFF
--- a/packages/open-next/src/adapters/util.ts
+++ b/packages/open-next/src/adapters/util.ts
@@ -1,7 +1,11 @@
 //TODO: We should probably move all the utils to a separate location
 
 export function setNodeEnv() {
-  process.env.NODE_ENV = process.env.NODE_ENV ?? "production";
+  // Note: we create a `processEnv` variable instead of just using `process.env` directly
+  //       because build tools can substitute `process.env.NODE_ENV` on build making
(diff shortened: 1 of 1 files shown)


vicb (approved):

conico974 (approved):
LGTM
Not a big fan of this change either to be honest. We need to find a better way of defining this (or delegate it to the platform itself).
Order of import can cause this issue https://github.com/opennextjs/opennextjs-aws/issues/624

dario-piotrowicz:
> LGTM Not a big fan of this change either to be honest. We need to find a better way of defining this (or delegate it to the platform itself). Order of import can cause this issue #624

thanks 🙂

yeah I agree it does feel like this could be improved 😕

PR #689, opened 2025-01-04
Feat: Add dev overrides

conico974:
This PR adds a bunch of overrides that is useful to run OpenNext locally.
They can be used in the config file: 
```ts
import type {OpenNextConfig} from "@opennextjs/aws/types/open-next.js"

export default {
  default: {
    override: {
      wrapper: "express-dev",
      converter: "node",
      incrementalCache: "fs-dev",
      queue: "direct",
      tagCache: "fs-dev"
    }
  },

  imageOptimization: {
    override: {
      wrapper: "dummy",
      converter: "dummy"
    },
    loader: "fs-dev",
    // This part is not needed on arm linux, and image optimization will only work in linux
    install: {
      arch: "x64",
      packages: ["sharp"]
    }
  },
} satisfies OpenNextConfig
```

`"direct"` queue could also be used to remove the need to have a queue. It should not be used where `waitUntil` is not properly supported as it means that ISR will wait for the revalidation to be done before returning the response. Revalidation request will also not be deduped

### DIFF
--- a/packages/open-next/src/overrides/imageLoader/fs-dev.ts
+++ b/packages/open-next/src/overrides/imageLoader/fs-dev.ts
@@ -1,0 +1,17 @@
+import type { ImageLoader } from "types/overrides";
+import fs from "node:fs/promises";
+import { Readable } from "node:stream";
+
+export default {
+  name: "fs-dev",
+  load: async (url: string) => {
+    const basePath = "../../assets";
+    const fileData = await fs.readFile(`${basePath}/${url}`);
+    const contentType = url.endsWith(".png") ? "image/png" : "image/jpeg";
+    return {
+      body: Readable.from(fileData),
+      contentType,
+      cacheControl: "public, max-age=31536000, immutable",
+    };
+  },
+} satisfies ImageLoader;
--- a/packages/open-next/src/overrides/imageLoader/fs-dev.ts
+++ b/packages/open-next/src/overrides/imageLoader/fs-dev.ts
@@ -1,6 +1,6 @@
-import type { ImageLoader } from "types/overrides";
 import fs from "node:fs/promises";
 import { Readable } from "node:stream";
+import type { ImageLoader } from "types/overrides";
 
 export default {
   name: "fs-dev",
--- a/packages/open-next/src/overrides/imageLoader/fs-dev.ts
+++ b/packages/open-next/src/overrides/imageLoader/fs-dev.ts
@@ -1,15 +1,14 @@
-import fs from "node:fs/promises";
-import { Readable } from "node:stream";
+import fs from "node:fs";
 import type { ImageLoader } from "types/overrides";
 
 export default {
   name: "fs-dev",
   load: async (url: string) => {
     const basePath = "../../assets";
-    const fileData = await fs.readFile(`${basePath}/${url}`);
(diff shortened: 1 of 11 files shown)


packages/open-next/src/overrides/imageLoader/fs-dev.ts
```diff
@@ -0,0 +1,17 @@
+import fs from "node:fs/promises";
+import { Readable } from "node:stream";
+import type { ImageLoader } from "types/overrides";
+
+export default {
+  name: "fs-dev",
+  load: async (url: string) => {
+    const basePath = "../../assets";
+    const fileData = await fs.readFile(`${basePath}/${url}`);
```
vicb, line 9: nit: use `fs.createReadStream()` instead ? (no a big deal for dev anyway)

packages/open-next/src/overrides/queue/direct.ts
```diff
@@ -0,0 +1,21 @@
+import type { Queue } from "types/overrides.js";
+
+const queue: Queue = {
+  name: "dev-queue",
```
vicb, line 4: nit: "direct-queue" for consistency ?

packages/open-next/src/overrides/tagCache/fs-dev.ts
```diff
@@ -0,0 +1,53 @@
...
+    const revalidatedTags = tags.filter(
+      (tagPathMapping) =>
+        tagPathMapping.path.S === path &&
+        Number.parseInt(tagPathMapping.revalidatedAt.N) > (lastModified ?? 0),
+    );
+    return revalidatedTags.length > 0 ? -1 : (lastModified ?? Date.now());
+  },
+  writeTags: async (newTags) => {
+    const unchangedTags = tags.filter((tagPathMapping) =>
+      newTags.some(
```
vicb, line 37: is this missing a `!` 🤔

vicb (commented):
NIce!

A few nits

packages/open-next/src/overrides/tagCache/fs-dev.ts
conico974: Yeah good catch

packages/open-next/src/overrides/tagCache/fs-dev.ts
```diff
@@ -0,0 +1,54 @@
...
+      (tagPathMapping) =>
+        tagPathMapping.path.S === path &&
+        Number.parseInt(tagPathMapping.revalidatedAt.N) > (lastModified ?? 0),
+    );
+    return revalidatedTags.length > 0 ? -1 : (lastModified ?? Date.now());
+  },
+  writeTags: async (newTags) => {
+    const unchangedTags = tags.filter(
+      (tagPathMapping) =>
+        !newTags.some(
```
vicb, line 38: one way to avoid the nested loop would be to create a Set of `${tag.length}-${tag}-${path}` and use `Set#has`

conico974: I guess we can do that. Probably not a big deal though as this override should only be used in dev.
I'll update this tomorrow

vicb (approved):
The `Set` thing was only a suggestion, up to you to apply or not (I agree it's not a big deal)

PR #692, opened 2025-01-05
e2e: add open-graph image tests

sommeeeer:
This PR will add e2e testing for `next/og` and `opengraph-image.tsx` in the app router

- An empty page on the route `/og` that should have an opengraph image present in the metatags.
- An API route on `/api/og` that returns an `ImageResponse` from `next/og`.

### DIFF
--- a/packages/tests-e2e/tests/appRouter/og.test.ts
+++ b/packages/tests-e2e/tests/appRouter/og.test.ts
@@ -1,4 +1,12 @@
+import { createHash } from "node:crypto";
 import { expect, test } from "@playwright/test";
+
+const OG_MD5 = "6e5e794ac0c27598a331690f96f05d00";
+const API_OG_MD5 = "cac95fc3e2d4d52870c0536bb18ba85b";
+
+function validateMd5(data: Buffer, expectedHash: string) {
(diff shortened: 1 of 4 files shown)


packages/tests-e2e/tests/appRouter/og.test.ts
```diff
@@ -0,0 +1,59 @@
+import { createHash } from "node:crypto";
+import { expect, test } from "@playwright/test";
+
+const OG_MD5 = "6e5e794ac0c27598a331690f96f05d00";
```
vicb, line 4: Do you think a comment on how to update those checksums would help (in case it changes)?

sommeeeer: > Do you think a comment on how to update those checksums would help (in case it changes)?

Good point, i'll add them now.

sommeeeer:
Perhaps we should consider adding another test for checking if something custom is passed to the HTTP response in `ImageResponse` [options](https://nextjs.org/docs/app/api-reference/functions/image-response)?

conico974 (approved):
LGTM Thanks

PR #694, opened 2025-01-08
fix: host not in route handler url when using external middleware

james-elicx:
ref: https://github.com/opennextjs/opennextjs-cloudflare/pull/234#discussion_r1907320812


conico974 (approved):
LGTM Thanks for that

Issue #621, opened 2024-11-06
[RFC] Incremental Cache improvements


khuezy:
https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/overrides/incrementalCache/s3.ts#L43-L58

I had a conversation with the Vercel devs and they said that their cacheHandle makes use of `lru-cache` to prevent extraneous calls to S3 (or any external datastore). When I was using `sst NextjsSite`, I noticed lots of S3 hits too:

```
async get(key, isFetch) {
    const result = await s3Client.send(
      new GetObjectCommand({
        Bucket: CACHE_BUCKET_NAME,
        Key: buildS3Key(key, isFetch ? "fetch" : "cache"),
      }),
    );

    const cacheData = JSON.parse(
      (await result.Body?.transformToString()) ?? "{}",
    );
    return {
      value: cacheData,
      lastModified: result.LastModified?.getTime(),
    };
  },
 ```

Whenever a page is reloaded, it would make the S3 call due to the `get`. Is this correct behavior? Or should we place `lru-cache` here before hitting S3? eg if cache is null or is stale.

CC: @conico974 @vicb

conico974:
From what i've understood of what they're saying and also observed it myself the way it works in our case is exactly how we want it to work.

Except for PPR cases (which are special), this will get called in 2 situations 
- Either for an ISR/SSG route (only if not cached by the CDN)
- Or for the fetch/unstable cache if used (can be the case in SSR if configured to do so)

It can become costly when you use the fetch cache a lot (especially in SSR) or very low revalidation time with a lot of pages, but that's kind of expected.

As far as i know they use an LRU cache only in `standalone` and in single instance.

The problem with an In-Memory LRU cache is with on demand revalidation. There is simply no way to tell all the alive lambda/function about that without some custom runtime and a continously running lambda, and even then, is it worth it ?
The only way to improve this would be to reduce the correctness.

There is an alternative that would be to create a multi tiered cache system, it should probably be in 3-tier :
- A very fast in-memory cache that become stale after a very short amount of time (by default 0s since it could break stuff)
- Something in DDB or redis or similar where we could easily retrieve metadata (like lastUpdatedTime), this would be faster and cheaper than S3 and we could probably even store some small data (small fetch for example). In case data is not stale in DDB then we would return either by order of pref : In-Memory, DDB, S3.
- And the S3 or equivalent where we would finally retrieve data.

[...]

PR #699, opened 2025-01-14
Feat: Multi-tiered cache for aws

conico974:
As discussed in #621.
The new incremental cache work this way : 
- First it checks the local in memory LRU cache and that it's not expired ( default TTL is 0 to avoid issue with On Demand revalidation )
- If found but expired, it checks DynamoDB metadata to check if the data has been updated somewhere else ( If it's not been updated it will return the locally retrieved entry )
- Lastly it'll check S3 cache and populate the local LRU cache

It requires both an S3 Bucket and a DynamoDb table
It uses the same env variable as both the default S3 and Tag cache as well as 2 new optional env `OPEN_NEXT_LOCAL_CACHE_TTL` (default is 0) and `OPEN_NEXT_LOCAL_CACHE_SIZE` (default is 1000, in nbr of entries)

> [!WARNING] 
This implementation is **NOT** strongly consistent
If some error happens while writing to DDB, some instance may use outdated local data

### DIFF
--- a/packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
+++ b/packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
@@ -1,0 +1,172 @@
+import type { CacheValue, IncrementalCache } from "types/overrides";
+import S3Cache, { getAwsClient } from "./s3-lite";
+import { customFetchClient } from "utils/fetch";
+import { debug } from "../../adapters/logger";
+
+// TTL for the local cache in milliseconds
+const localCacheTTL = process.env.OPEN_NEXT_LOCAL_CACHE_TTL
+  ? Number.parseInt(process.env.OPEN_NEXT_LOCAL_CACHE_TTL)
+  : 0;
+// Maximum size of the local cache in nb of entries
+const maxCacheSize = process.env.OPEN_NEXT_LOCAL_CACHE_SIZE
+  ? Number.parseInt(process.env.OPEN_NEXT_LOCAL_CACHE_SIZE)
+  : 1000;
+
+class LRUCache {
+  private cache: Map<
+    string,
+    {
+      value: CacheValue<boolean>;
+      lastModified: number;
+    }
+  > = new Map();
+  private maxSize: number;
+
+  constructor(maxSize: number) {
+    this.maxSize = maxSize;
+  }
+
+  // isFetch is not used here, only used for typing
+  get<T extends boolean = false>(key: string, isFetch?: T) {
+    return this.cache.get(key) as {
+      value: CacheValue<T>;
+      lastModified: number;
+    };
+  }
+
+  set(key: string, value: any) {
+    if (this.cache.size >= this.maxSize) {
+      const firstKey = this.cache.keys().next().value;
+      if (firstKey) {
+        this.cache.delete(firstKey);
+      }
+    }
+    this.cache.set(key, value);
+  }
+
+  delete(key: string) {
+    this.cache.delete(key);
+  }
+}
+
+const localCache = new LRUCache(maxCacheSize);
+
+const awsFetch = (body: RequestInit["body"], type: "get" | "set" = "get") => {
+  const { CACHE_BUCKET_REGION } = process.env;
+  const client = getAwsClient();
+  return customFetchClient(client)(
+    `https://dynamodb.${CACHE_BUCKET_REGION}.amazonaws.com`,
+    {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/x-amz-json-1.0",
+        "X-Amz-Target": `DynamoDB_20120810.${
+          type === "get" ? "GetItem" : "PutItem"
+        }`,
+      },
+      body,
+    },
+  );
+};
+
+const buildDynamoKey = (key: string) => {
+  const { NEXT_BUILD_ID } = process.env;
+  return `__meta_${NEXT_BUILD_ID}_${key}`;
+};
+
+/**
+ * This cache implementation uses a multi-tier cache with a local cache, a DynamoDB metadata cache and an S3 cache.
+ * It uses the same DynamoDB table as the default tag cache and the same S3 bucket as the default incremental cache.
+ * It will first check the local cache.
+ * If the local cache is expired, it will check the DynamoDB metadata cache to see if the local cache is still valid.
+ * Lastly it will check the S3 cache.
+ */
+const multiTierCache: IncrementalCache = {
+  name: "multi-tier-ddb-s3",
+  async get(key, isFetch) {
+    // First we check the local cache
+    const localCacheEntry = localCache.get(key, isFetch);
+    if (localCacheEntry) {
+      if (Date.now() - localCacheEntry.lastModified < localCacheTTL) {
+        debug("Using local cache without checking ddb");
+        return localCacheEntry;
+      }
+      try {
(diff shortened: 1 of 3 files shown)


packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
```diff
@@ -0,0 +1,172 @@
...
+    {
+      value: CacheValue<boolean>;
+      lastModified: number;
+    }
+  > = new Map();
+  private maxSize: number;
+
+  constructor(maxSize: number) {
+    this.maxSize = maxSize;
+  }
```
vicb, line 27:
```suggestion
  constructor(private maxSize: number) {
  }
```

packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  // isFetch is not used here, only used for typing
+  get<T extends boolean = false>(key: string, isFetch?: T) {
+    return this.cache.get(key) as {
```
vicb, line 31: To really have a LRU, you would need to delete and re-set the key. For now it's a FIFO.

packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
```diff
@@ -0,0 +1,172 @@
+import type { CacheValue, IncrementalCache } from "types/overrides";
+import { customFetchClient } from "utils/fetch";
+import { debug } from "../../adapters/logger";
+import S3Cache, { getAwsClient } from "./s3-lite";
+
+// TTL for the local cache in milliseconds
+const localCacheTTL = process.env.OPEN_NEXT_LOCAL_CACHE_TTL
+  ? Number.parseInt(process.env.OPEN_NEXT_LOCAL_CACHE_TTL)
```
vicb, line 8: - parseInt(..., 10) and below
- OPEN_NEXT_LOCAL_CACHE_TTL_MS ?

packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
conico974: Ho yeah you're right, i'll make the change

packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
```diff
@@ -0,0 +1,172 @@
...
+  : 0;
+// Maximum size of the local cache in nb of entries
+const maxCacheSize = process.env.OPEN_NEXT_LOCAL_CACHE_SIZE
+  ? Number.parseInt(process.env.OPEN_NEXT_LOCAL_CACHE_SIZE)
+  : 1000;
+
+class LRUCache {
```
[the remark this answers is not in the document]
conico974: Probably a better idea yeah, even though it's only used here for now

packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
```diff
@@ -0,0 +1,139 @@
...
+          const data = await result.json();
+          const hasBeenDeleted = data.Item?.deleted?.BOOL;
+          if (hasBeenDeleted) {
+            localCache.delete(key);
+            return { value: undefined, lastModified: 0 };
+          }
+          // If the metadata is older than the local cache, we can use the local cache
+          // If it's not found we assume that no write has been done yet and we can use the local cache
+          const lastModified = data.Item?.revalidatedAt?.N
+            ? Number.parseInt(data.Item.revalidatedAt.N)
```
vicb, line 83:
```suggestion
            ? Number.parseInt(data.Item.revalidatedAt.N, 10)
```

packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
```diff
@@ -0,0 +1,139 @@
...
+      localCache.set(key, {
+        value: result.value,
+        lastModified: result.lastModified ?? Date.now(),
+      });
+    }
+    return result;
+  },
+  async set(key, value, isFetch) {
+    const revalidatedAt = Date.now();
+    await S3Cache.set(key, value, isFetch);
```
vicb, line 105: use `Promise.allSetlled()` to parallelize?

packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
```diff
@@ -0,0 +1,139 @@
...
+      }),
+      "set",
+    );
+    localCache.set(key, {
+      value,
+      lastModified: revalidatedAt,
+    });
+  },
+  async delete(key) {
+    await S3Cache.delete(key);
```
vicb, line 123: Same comment: as for set: `Promise.allSettled`

packages/open-next/src/utils/lru.ts
```diff
@@ -0,0 +1,28 @@
...
+      this.cache.delete(key);
+      this.cache.set(key, result);
+    }
+    return result;
+  }
+
+  set(key: string, value: any) {
+    if (this.cache.size >= this.maxSize) {
+      const firstKey = this.cache.keys().next().value;
+      if (firstKey) {
```
vicb, line 18: nit (empty string)

```suggestion
      if (firstKey !== undefined) {
```

packages/open-next/src/utils/lru.ts
```diff
@@ -0,0 +1,28 @@
+export class LRUCache<T> {
+  private cache: Map<string, T> = new Map();
+
+  constructor(private maxSize: number) {}
+
+  get(key: string) {
+    const result = this.cache.get(key);
+    if (result) {
+      this.cache.delete(key);
```
vicb, line 9: nit: maybe add a comment to explain why, it might be surprising at first sight.

packages/open-next/src/utils/lru.ts
```diff
@@ -0,0 +1,28 @@
+export class LRUCache<T> {
+  private cache: Map<string, T> = new Map();
+
+  constructor(private maxSize: number) {}
+
+  get(key: string) {
+    const result = this.cache.get(key);
```
vicb, line 7: `.has` would be allow storing `null` but as we don't need that, maybe only add a comment on the class?

vicb (approved):
LGTM, a few nits

packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
conico974: This is on purpose actually, given how it works we have 3 choice for handling write error failure: 
- We could do as it is here, set on S3, then set on DDB, which means that if it fails in DDB, instance that don't have local cache will work as expected. ( But those with local cache will still serve outdated data )
- We could first write in DDB and then S3, which means that in case S3 fail, new instance will fetch outdated data, and existing instance will fetch outdated data.
- Or we could use `allSettled` but then the behavior will be unpredictable in case one of the 2 fails

I should have added a comment explaining this. One other thing we could do is to let the user chose the behavior they'd want.

I'll update and merge the PR tomorrow in case we should chose another option

vicb: Oh you're right 👍
Thanks for explaining very clearly

PR #701, opened 2025-01-16
refactor: use utility for cross-platform path regex construction

james-elicx:
- Create a new utility for constructing regular expressions with cross-platform paths.
- Update usage to use the new utility.

### DIFF
--- a/packages/open-next/src/utils/regex.ts
+++ b/packages/open-next/src/utils/regex.ts
@@ -14,8 +14,8 @@ export function getCrossPlatformPathRegex(
   opts: { escape: boolean } = { escape: true },
 ) {
   const newExpr = (
-    opts.escape ? regex.replace(/([[\]().*+?^$|])/g, "\\$1") : regex
-  ).replaceAll("/", "(?:\\/|\\\\)");
+    opts.escape ? regex.replace(/([[\]().*+?^$|{}\\])/g, "\\$1") : regex
+  ).replaceAll("/", String.raw`(?:\/|\\)`);
 
   return new RegExp(newExpr, "g");
 }
--- a/packages/open-next/src/utils/regex.ts
+++ b/packages/open-next/src/utils/regex.ts
@@ -1,3 +1,14 @@
+/**
+ * Constructs a regular expression for a path that supports separators for multiple platforms
+ *  - Uses posix separators (`/`) as the input that should be made cross-platform.
+ *  - Special characters are escaped by default but can be controlled through opts.escape.
+ *
+ * @example
+ * ```ts
+ * getCrossPlatformPathRegex("./middleware.mjs")
+ * getCrossPlatformPathRegex("\\./middleware\\.(mjs|cjs)", { escape: false })
+ * ```
+ */
 export function getCrossPlatformPathRegex(
   regex: string,
   opts: { escape: boolean } = { escape: true },
--- a/packages/open-next/src/utils/regex.ts
+++ b/packages/open-next/src/utils/regex.ts
@@ -1,0 +1,10 @@
+export function getCrossPlatformPathRegex(
+  regex: string,
+  opts: { escape: boolean } = { escape: true },
+) {
+  const newExpr = (
+    opts.escape ? regex.replace(/([[\]().*+?^$|])/g, '\\$1') : regex
+  ).replaceAll('/', '(?:\\/|\\\\)');
+
+  return new RegExp(newExpr, 'g');
+}
--- a/packages/open-next/src/utils/regex.ts
+++ b/packages/open-next/src/utils/regex.ts
@@ -2,6 +2,7 @@
  * Constructs a regular expression for a path that supports separators for multiple platforms
  *  - Uses posix separators (`/`) as the input that should be made cross-platform.
  *  - Special characters are escaped by default but can be controlled through opts.escape.
+ *  - Posix separators are always escaped.
  *
  * @example
  * ```ts
--- a/packages/open-next/src/utils/regex.ts
+++ b/packages/open-next/src/utils/regex.ts
@@ -6,7 +6,7 @@
  * @example
  * ```ts
  * getCrossPlatformPathRegex("./middleware.mjs")
- * getCrossPlatformPathRegex("\\./middleware\\.(mjs|cjs)", { escape: false })
+ * getCrossPlatformPathRegex(String.raw`\./middleware\.(mjs|cjs)`, { escape: false })
  * ```
  */
 export function getCrossPlatformPathRegex(
--- a/packages/open-next/src/utils/regex.ts
+++ b/packages/open-next/src/utils/regex.ts
@@ -1,3 +1,8 @@
+type Options = {
+  escape?: boolean;
+  flags?: string;
+};
+
 /**
  * Constructs a regular expression for a path that supports separators for multiple platforms
  *  - Uses posix separators (`/`) as the input that should be made cross-platform.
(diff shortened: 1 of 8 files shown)


packages/open-next/src/utils/regex.ts
```diff
@@ -0,0 +1,10 @@
+export function getCrossPlatformPathRegex(
+  regex: string,
+  opts: { escape: boolean } = { escape: true },
```
conico974, line 3: Why the escape options ? It's not used anywhere.
My take on this is that we should only escape path related stuff here (i.e. `/` the starting `./` and file extensions)

packages/tests-unit/tests/utils/regex.test.ts
```diff
@@ -0,0 +1,22 @@
+import { getCrossPlatformPathRegex } from "@opennextjs/aws/utils/regex.js";
+
+const specialChars = "^([123]+|[123]*)?$";
+
+describe("getCrossPlatformPathRegex", () => {
```
conico974, line 5: Ideally here i'd like to have (in addition to more generic test like the one you made) more realistic test, it's a bit more readable.
We could test things like `core/resolve.js` and `./middleware.mjs` against both windows and unix path

packages/open-next/src/utils/regex.ts
vicb: > Why the escape options ?

That's something I asked James.

If you only want to escape a path then you would use the default value (`true`)

But sometimes (i.e. ESBuild filter), you might want to have more control on the pattern and do not escape special chars.

packages/open-next/src/utils/regex.ts
```diff
@@ -0,0 +1,10 @@
+export function getCrossPlatformPathRegex(
```
vicb, line 1: Maybe add a comment:

- The regex is in posix format (uses "/" directory separator)
- escape controls whether the special char are escaped.

Would be nice to have 2 examples:
- `getCrossPlatformPathRegex("./middleware.mjs")`
- getCrossPlatformPathRegex("\./middleware\.(mjs|cjs)", { escape: false })

packages/open-next/src/utils/regex.ts
```diff
@@ -0,0 +1,10 @@
...
+  regex: string,
+  opts: { escape: boolean } = { escape: true },
+) {
+  const newExpr = (
+    opts.escape ? regex.replace(/([[\]().*+?^$|])/g, "\\$1") : regex
```
vicb, line 6: The set looks to be missing `{`, `}`, `\`

packages/open-next/src/utils/regex.ts
```diff
@@ -0,0 +1,10 @@
...
+  ).replaceAll("/", "(?:\\/|\\\\)");
```
vicb, line 7: nit/personal pref (`\\\\` makes me 😱)
```suggestion
  ).replaceAll("/", String.raw`(?:\/|\\)`);
```

vicb (commented):
@james-elicx thanks for the change, LGTM.

Let's have more discussion with @conico974 about the exact behavior this should have

packages/open-next/src/utils/regex.ts
conico974: > `getCrossPlatformPathRegex("./middleware.(mjs|cjs)", { escape: false })`

This is why i don't like the escape options, this one is technically not correct it should be `getCrossPlatformPathRegex("\./middleware\.(mjs|cjs)", { escape: false })`. 
It makes it kind of weird (when `{escape:false}` is set) to escape some part but not others (even one that you may miss like the starting `.`)

vicb:
@conico974 replying here as the convo is resolved:

I understand your point but sounds like "only escape path related stuff here (i.e. / the starting ./ and file extensions)" is not without surprises either.

For now I feel like we have those solutions:
- only handle the separator
- handle the separator and escape the special chars (do not offer `{ escape: false }` - I will come with a different function for my needs
- use the impl in this PR.

I need some think to think about it.

Notes:
- `/(\/|\\)edgeFunctionHandler.js/g` was wrongly escaped before this PR (only to say there is probably no ideal solution, as you mentioned we can get things wrong after this PR too)
- I think you replied before I updated [my message](https://github.com/opennextjs/opennextjs-aws/pull/701#discussion_r1919709531)

packages/tests-unit/tests/utils/regex.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    [
+      ["core/resolve.js", String.raw`core(?:\/|\\)resolve\.js`],
+      ["./middleware.mjs", String.raw`\.(?:\/|\\)middleware\.mjs`],
+    ].forEach(([input, output]) =>
+      expect(getCrossPlatformPathRegex(input).source).toEqual(output),
+    );
+  });
+
+  it("should return cross-platform paths without escaping special characters", () => {
+    const regex = getCrossPlatformPathRegex("\\./middleware\\.(mjs|cjs)", {
```
vicb, line 35: String.rwa would be nice here too :)

conico974:
@vicb As you said there is probably no ideal solution for this ( unless with some typescript wizardry, but it's not worth the effort ).
Personally i would have just made it escape path related stuff, and if we need more complex stuff we just write our own regex. 
But i'm fine with the current implementation as long as we provide enough explanation as to what needs to be done for these special cases.

packages/open-next/src/utils/regex.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Constructs a regular expression for a path that supports separators for multiple platforms
+ *  - Uses posix separators (`/`) as the input that should be made cross-platform.
+ *  - Special characters are escaped by default but can be controlled through opts.escape.
```
conico974, line 4: It should be added that `/` will always be escaped even when setting `escape:false`

james-elicx:
```suggestion
 *  - Special characters are escaped by default but can be controlled through opts.escape.
 *  - Posix separators are always escaped.
```

packages/open-next/src/utils/regex.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function getCrossPlatformPathRegex(
+  regex: string,
+  opts: { escape: boolean } = { escape: true },
+) {
+  const newExpr = (
+    opts.escape ? regex.replace(/([[\]().*+?^$|{}\\])/g, "\\$1") : regex
+  ).replaceAll("/", String.raw`(?:\/|\\)`);
+
+  return new RegExp(newExpr, "g");
```
vicb, line 21: I don't think we really need "g" here?
At least it doesn't harm.
Do we want to be able to set that via `{escape = true, flags = "g"}`

packages/open-next/src/utils/regex.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @example
+ * ```ts
+ * getCrossPlatformPathRegex("./middleware.mjs")
+ * getCrossPlatformPathRegex(String.raw`\./middleware\.(mjs|cjs)`, { escape: false })
+ * ```
+ */
+export function getCrossPlatformPathRegex(
+  regex: string,
+  opts: { escape: boolean } = { escape: true },
```
vicb, line 15: I usually use:
 
```suggestion
  { escape = true },
```

vicb (approved):
Looks good to me, thanks for adding a clear doc and tests.

conico974 (approved):
LGTM, Thanks

PR #702, opened 2025-01-17
refactor(cache): deprecate global disableDynamoDBCache and disableIncrementalCache

vicb:
@conico974 Actually, this might be a breaking change...
So maybe I should deprecate instead and add a changeset then we can remove later.
WDYT?

### DIFF
--- a/.changeset/sweet-kiwis-agree.md
+++ b/.changeset/sweet-kiwis-agree.md
@@ -1,0 +1,12 @@
+---
+"@opennextjs/aws": patch
+---
+
+refactor(cache): deprecate global disableDynamoDBCache and disableIncrementalCache
+
+In the cache adapter:
+
+- `globalThis.disableDynamoDBCache` is deprecated and will be removed.
+  use `globalThis.openNextConfig.dangerous?.disableTagCache` instead.
+- `globalThis.disableIncrementalCache` is deprecated and will be removed.
+  use `globalThis.openNextConfig.dangerous?.disableIncrementalCache` instead.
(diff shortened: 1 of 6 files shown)


packages/open-next/src/build/compileCache.ts
```diff
@@ -28,15 +28,7 @@ export function compileCache(
...
-        js: [
-          `globalThis.disableIncrementalCache = ${
-            config.dangerous?.disableIncrementalCache ?? false
-          };`,
-          `globalThis.disableDynamoDBCache = ${
-            config.dangerous?.disableTagCache ?? false
-          };`,
-          `globalThis.isNextAfter15 = ${isAfter15};`,
-        ].join(""),
+        js: `globalThis.isNextAfter15 = ${isAfter15};`,
```
conico974, line 31: We should probably leave it for now here, in case some people rely on this in their custom overrides.
Just add a comment that we need to remove it in the future

conico974 (commented):
This is kind of a breaking change, some people may rely on it (especially if they have a custom tag cache that they reuse across multiple apps).
Fully removing this in a future minor (with a warning) is probably more than enough, there is likely very very few people that rely on this ( if any at all )

packages/open-next/src/build/compileCache.ts
vicb: I added a `@deprecated` annotation.

conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review

PR #704, opened 2025-01-20
make sure edge function entries are properly awaited

dario-piotrowicz:
the values in `self._ENTRIES` can be promises so they need to be awaited before using them, AFAICT that's also what Next.js  does: [source code](
https://github.com/vercel/next.js/blob/6d17cc6b0f3c153d2405ac97e862da2e9008258f/packages/next/src/server/web/sandbox/sandbox.ts#L119)

this PR makes sure that such values are properly awaited

### DIFF
--- a/.changeset/smart-carrots-build.md
+++ b/.changeset/smart-carrots-build.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: make sure edge function entries are properly awaited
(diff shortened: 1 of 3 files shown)


conico974 (approved):
LGTM Thanks.
Could you just add a changeset though ?

vicb (approved):
LGTM.

Is there a simple way to add a test?

dario-piotrowicz:
> LGTM Thanks. Could you just add a changeset though ?

Thanks 🙂, sorry my bad, here I've added one: e4c87e332 let me know if it's ok or if it should contain more details 🙂

dario-piotrowicz:
> Is there a simple way to add a test?

good point, let me have a look 👍

PR #705, opened 2025-01-20
add e2e test for top-level await in middleware

conico974:
This PR add test for #704
I've added a `@ts-expect-error` for the top level await, i didn't want to go down the rabbit hole of finding the issue here (middleware is somehow considered a cjs file and if set as a module, `next/server` import does not work anymore)
There is also a build time warning, but it still works both with `next start` or in the e2e test

### DIFF
--- a/examples/app-pages-router/middleware.ts
+++ b/examples/app-pages-router/middleware.ts
@@ -1,6 +1,14 @@
 import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
 
+// Needed to test top-level await
+// @ts-expect-error - It will cause a warning at build time, but it should just work
+const topLevelAwait = await new Promise<string>((resolve) => {
+  setTimeout(() => {
+    resolve("top-level-await");
+  }, 150);
+});
+
 export function middleware(request: NextRequest) {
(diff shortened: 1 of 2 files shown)


khuezy (approved):

examples/app-pages-router/middleware.ts
```diff
@@ -1,6 +1,14 @@
 import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
 
+// Needed to test top-level await
+// @ts-expect-error - It will cause a warning at build time, but it should just work
+const topLevelAwait = await new Promise<string>((resolve) => {
+  setTimeout(() => {
+    resolve("top-level-await");
+  }, 150);
```
khuezy, line 9: Interesting - does it have to have a timeout? Would `await Promise.resolve('top-level-await')` work w/o the timeout?

conico974: Not necessarily, it should work with `Promise.resolve`. I just wanted to emulate some wait time and be sure that we are not running on the same loop of the event loop.
We can probably make it 5 or 10 ms, no need for 150

PR #708, opened 2025-01-22
fix: return when an error is downplayed

sommeeeer:
People were reporting getting this error in SST Console: 
```Name: 'S3Client', commandName: 'GetObjectCommand', input: { Bucket: 'myapp-dev-myappassets-bvaaaaa', Key: '_cache/kbhQ5yYbaaaaaaaAa/dk.cache' }, error: NoSuchKey: The specified key does not exist.```

The error gets downplayed correctly but turns out we are not returning when an error is downplayed so this [line](https://github.com/opennextjs/opennextjs-aws/blob/9e770b53701e96f2007926b86450699696763bae/packages/open-next/src/adapters/logger.ts#L72) gets run aswell. This PR fixes this.

### DIFF
--- a/.changeset/orange-impalas-repeat.md
+++ b/.changeset/orange-impalas-repeat.md
@@ -2,4 +2,8 @@
 "@opennextjs/aws": patch
 ---
 
-Early return when an error gets downplayed
+fix: add early return for downplayed aws-sdk errors
+
+In the logger adapter:
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: .changeset/orange-impalas-repeat.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Early return when an error gets downplayed
```
vicb, line 5: Nit: maybe tell what you fixed instead of how?

vicb (approved):

conico974 (approved):
LGTM Thanks

PR #709, opened 2025-01-23
fix #648

### LINKED ISSUE #648: When did incremental became stable?
@petebacondarwin noted this code in https://github.com/opennextjs/opennextjs-aws/blob/2202f36ce0f87357b249bd127cdd5e84d6deffd3/packages/open-next/src/build/createServerBundle.ts

```ts
  const isAfter141 =
    buildHelper.compareSemver(options.nextVersion, "14.0.4") >= 0;
// ...

    openNextReplacementPlugin({
      name: `utilOverride ${name}`,
      target: /core(\/|\\)util\.js/g,
      deletes: [
        ...(isAfter141
          ? ["experimentalIncrementalCacheHandler"]
          : ["stableIncrementalCache"]),
      ],
    }),
```

Which of 14.0.4 or 14.1 is the correct version?

conico974:
fix #648

### DIFF
--- a/.changeset/unlucky-olives-try.md
+++ b/.changeset/unlucky-olives-try.md
@@ -2,4 +2,4 @@
 "@opennextjs/aws": patch
 ---
 
-use correct version for applying stableIncrementalCache (It failed for next 14.0.4)
+fix: stableIncrementalCache is only used for Next.js >= 14.1
--- a/.changeset/unlucky-olives-try.md
+++ b/.changeset/unlucky-olives-try.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+use correct version for applying stableIncrementalCache (It failed for next 14.0.4)
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -277,6 +277,7 @@ export function getNextVersion(appPath: string): string {
 
 /**
  * Compare two semver versions.
+ *
  * @param v1 - First version. Can be "latest", otherwise it should be a valid semver version in the format of `major.minor.patch`. Usually is the next version from the package.json without canary suffix. If minor or patch are missing, they are considered 0.
  * @param v2 - Second version. Should not be "latest", it should be a valid semver version in the format of `major.minor.patch`. If minor or patch are missing, they are considered 0.
  * @see
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -290,7 +290,7 @@ export function compareSemver(v1: string, v2: string): number {
 
   if (major1 !== major2) return major1 - major2;
   if (minor1 !== minor2) return minor1 - minor2;
-  return patch1 - patch2;
+  return (patch1 ?? 0) - (patch2 ?? 0);
 }
 
 export function copyOpenNextConfig(
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -275,6 +275,15 @@ export function getNextVersion(appPath: string): string {
   return version.split("-")[0];
 }
 
+/**
+ * Compare two semver versions.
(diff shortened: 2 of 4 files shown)


khuezy (approved):

conico974:
Oops, i was about to merge and i noticed an issue @khuezy PTAL

packages/open-next/src/build/helper.ts
```diff
@@ -290,7 +290,7 @@ export function compareSemver(v1: string, v2: string): number {
 
   if (major1 !== major2) return major1 - major2;
   if (minor1 !== minor2) return minor1 - minor2;
-  return patch1 - patch2;
+  return (patch1 ?? 0) - (patch2 ?? 0);
```
vicb, line 293: don't you have the same issue with minor?

vicb:
Not sure if important but:

```ts
if (v1 === "latest") return 1;

// ->

if (v1 === "latest") {
return v2 === "latest" ? 0 : 1;
}
```

packages/open-next/src/build/helper.ts
conico974: Technically we could have this issue for minor2 only.
We only use this to compare with next version, and the next version comes from `next/package.json` (so should always contain major, minor and patch)

v2 is the one we provide to test, i don't see in which scenario we'd use latest

I should add a comment above the function so that we know about its limitation and how it should be used

khuezy: The function assumes that `M, m, p` exists. 

```
 const [major1 = 0, minor1 = 0, patch1 = 0] = v1.split(".").map(Number);
 const [major2 = 0, minor2 = 0, patch2 = 0] = v2.split(".").map(Number);
 ```
 
 This should default those values.

**Review by khuezy — APPROVED** [review]
LGTM thanks!

File: packages/open-next/src/build/helper.ts [hunk 2]
```diff
@@ -275,6 +275,15 @@ export function getNextVersion(appPath: string): string {
   return version.split("-")[0];
 }
 
+/**
+ * Compare two semver versions.
+ * @param v1 - First version. Can be "latest", otherwise it should be a valid semver version in the format of `major.minor.patch`. Usually is the next version from the package.json without canary suffix. If minor or patch are missing, they are considered 0.
+ * @param v2 - Second version. Should not be "latest", it should be a valid semver version in the format of `major.minor.patch`. If minor or patch are missing, they are considered 0.
+ * @see
```
vicb, line 282: I can't see anything :)

packages/open-next/src/build/helper.ts
```diff
@@ -275,6 +275,15 @@ export function getNextVersion(appPath: string): string {
   return version.split("-")[0];
 }
 
+/**
+ * Compare two semver versions.
```
vicb, line 279:
```suggestion
/**
 * Compare two semver versions.
 *
```

.changeset/unlucky-olives-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+use correct version for applying stableIncrementalCache (It failed for next 14.0.4)
```
vicb, line 5: fix: stableIncrementalCache is only used for Next.js >= 14.1

vicb (approved):
LGTM!

Thought.

When I see

```ts
buildHelper.compareSemver(options.nextVersion, "14.1") >= 0;
```

I awlays think that there would be a nice way:

```ts
buildHelper.compareSemver(options.nextVersion, ">=", 14.1);
```

conico974:
>LGTM!
>
>Thought.
>
>When I see
>
> ```ts 
>buildHelper.compareSemver(options.nextVersion, "14.1") >= 0;
>```
>I always think that there would be a nicer way:
> ```ts
>buildHelper.compareSemver(options.nextVersion, ">=", 14.1);
>```

I guess we could refactor it later, i'll merge it for now

PR #710, opened 2025-01-24
feat: support new bun lock file

cuquo:
adds support of the new non-binary lock file 

ref: https://bun.sh/docs/install/lockfile

### DIFF
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -69,6 +69,7 @@ function findMonorepoRoot(appPath: string) {
   let currentPath = appPath;
   while (currentPath !== "/") {
     const found = [
+      // bun can generate yaml lock files (`bun install --yarn`) so bun should be before yarn
       { file: "bun.lockb", packager: "bun" as const },
       { file: "bun.lock", packager: "bun" as const },
       { file: "package-lock.json", packager: "npm" as const },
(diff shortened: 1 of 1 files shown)


khuezy (approved):
LGTM thanks!

packages/open-next/src/build/helper.ts
```diff
@@ -73,6 +73,7 @@ function findMonorepoRoot(appPath: string) {
       { file: "yarn.lock", packager: "yarn" as const },
       { file: "pnpm-lock.yaml", packager: "pnpm" as const },
       { file: "bun.lockb", packager: "bun" as const },
+      { file: "bun.lock", packager: "bun" as const },
```
vicb, line 76: It looks like `bun install --yarn` will also generate a `yarn` lock file so maybe the bun entries should be moved before yarn?

vicb, line 76:
```suggestion
      { file: "bun.lockb", packager: "bun" as const },
      { file: "bun.lock", packager: "bun" as const },
      { file: "yarn.lock", packager: "yarn" as const },
      { file: "pnpm-lock.yaml", packager: "pnpm" as const },
```

vicb (approved):

packages/open-next/src/build/helper.ts
```diff
@@ -69,10 +69,11 @@ function findMonorepoRoot(appPath: string) {
   let currentPath = appPath;
   while (currentPath !== "/") {
     const found = [
+      { file: "bun.lockb", packager: "bun" as const },
```
vicb, line 72:
```suggestion
      // bun can generate yaml lock files (`bun install --yarn`) so bun should be before yarn
      { file: "bun.lockb", packager: "bun" as const },
```

conico974 (approved):
LGTM Thanks

PR #717, opened 2025-01-31
Feat: tag cache mode

### LINKED ISSUE #706: [Feature] Add another mode for the tag cache
Right now the tag cache is made to work mostly with DynamoDB and does not work in the same way as in next standalone. We should add a different mode for the tag cache (and support both) that would work in the same way as in next standalone.

We should probably add a `dynamodb-nextmode` override and add some docs explaining the difference and which one you should choose

---

#### On `get`

We just check for the cache key in the tag cache. If it has been revalidated we just return `null`, otherwise we continue

#### On `set`

We have to write both the incremental cache and check the tag cache for non existing tag/key combination. For non existing tag/key combination, we have to add them

#### On `revalidateTag` for each tag

We have to update every possible combination for the requested tag

#### Pros
- Very fast on read
- Only one query per `get` (On DynamoDB it's a lot cheaper)
- Can allow for automatic cdn invalidation on `revalidateTag`
#### Cons
- Lots of write on `set` and `revalidateTag`
- Needs to be prepopulated at build time to work properly

---

#### On `get`

We have to check for every tag (after reading the incremental cache) that they have not been revalidated. 

In DynamoDB, this would require 1 `GetItem` per tag (including internal one), more realistically 1 `BatchGetItem` per `get` (In terms of pricing, it would be billed as multiple single `GetItem`)

#### On `set`

We don't have to do anything here

#### On `revalidateTag` for each tag

We have to update a single entry for this tag

#### Pros
- No need to prepopulate DDB
- Very little write
#### Cons
- Might be slower on read 
- One page request (i.e. `GET` request) could require to check a lot of tags (And some of them multiple time when used with the fetch cache)
- Almost impossible to do automatic cdn revalidation by itself

conico974:
Should close #706 
Everything should be explained in the issue already

One thing we could do that could improve perf in some cases would be to introduce a new optional method in `incrementalCache`: `getMetadata` that would retrieve `lastModified` as well as `tags`. If provided it would allow us to check the tag cache without having to load the cache value (i.e. probably useless for S3 only incremental cache). 
This should probably be done in another PR

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -66,7 +66,7 @@ export default class Cache {
       const _hasBeenRevalidated = await hasBeenRevalidated(
         key,
         _tags,
-        cachedEntry?.lastModified,
+        cachedEntry,
       );
 
       if (_hasBeenRevalidated) return null;
@@ -85,7 +85,7 @@ export default class Cache {
           const hasPathBeenUpdated = await hasBeenRevalidated(
             path.replace("_N_T_/", ""),
             [],
-            cachedEntry.lastModified,
+            cachedEntry,
           );
           if (hasPathBeenUpdated) {
             // In case the path has been revalidated, we don't want to use the fetch cache
@@ -121,7 +121,7 @@ export default class Cache {
       const _hasBeenRevalidated = await hasBeenRevalidated(
         key,
         tags,
-        _lastModified,
+        cachedEntry,
       );
       if (cacheData === undefined || _hasBeenRevalidated) return null;
 
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -1,5 +1,6 @@
 import { isBinaryContentType } from "../utils/binary";
 import { debug, error, warn } from "./logger";
+import { getTagFromValue, hasBeenRevalidated } from "utils/cache";
 
 interface CachedFetchValue {
   kind: "FETCH";
@@ -134,14 +135,15 @@ export default class Cache {
 
       if (cachedEntry?.value === undefined) return null;
 
-      const _lastModified = await globalThis.tagCache.getLastModified(
+      const _tags = [...(tags ?? []), ...(softTags ?? [])];
+      const _lastModified = cachedEntry.lastModified ?? Date.now();
+      const _hasBeenRevalidated = await hasBeenRevalidated(
         key,
+        _tags,
         cachedEntry?.lastModified,
       );
-      if (_lastModified === -1) {
-        // If some tags are stale we need to force revalidation
-        return null;
-      }
+
+      if (_hasBeenRevalidated) return null;
 
       // For cases where we don't have tags, we need to ensure that the soft tags are not being revalidated
       // We only need to check for the path as it should already contain all the tags
@@ -154,11 +156,12 @@ export default class Cache {
             !tag.endsWith("page"),
         );
         if (path) {
-          const pathLastModified = await globalThis.tagCache.getLastModified(
+          const hasPathBeenUpdated = await hasBeenRevalidated(
             path.replace("_N_T_/", ""),
+            [],
             cachedEntry.lastModified,
           );
-          if (pathLastModified === -1) {
+          if (hasPathBeenUpdated) {
             // In case the path has been revalidated, we don't want to use the fetch cache
             return null;
           }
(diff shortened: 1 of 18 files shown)


packages/open-next/src/types/overrides.ts
```diff
@@ -81,16 +86,31 @@ export type IncrementalCache = {
 
 // Tag cache
 
-export type TagCache = {
+type BaseTagCache = {
+  name: string;
+};
+
+export type NextModeTagCache = BaseTagCache & {
+  mode: "nextMode";
```
conico974, line 94: I really don't like this name but i couldn't find something better. Maybe someone else will find a better idea

packages/open-next/src/adapters/cache.ts
```diff
@@ -465,4 +400,28 @@ export default class Cache {
       error("Failed to revalidate tag", e);
     }
   }
+
+  private async updateTagsOnSet(key: string, derivedTags: string[]) {
+    if (
+      globalThis.openNextConfig.dangerous?.disableTagCache ||
+      globalThis.tagCache.mode === "nextMode"
+    ) {
+      return;
```
vicb, line 409: Maybe move this to the call site to avoid computing the derivedTags

packages/open-next/src/overrides/tagCache/constants.ts
```diff
@@ -1,4 +1,5 @@
 export const MAX_DYNAMO_BATCH_WRITE_ITEM_COUNT = 25;
+export const MAX_DYNAMO_BATCH_GET_ITEM_COUNT = 100;
```
vicb, line 2: unused?

vicb (commented):
Finally would some time to look at the changes, nice!
I have added a few comments.

packages/open-next/src/types/open-next.ts
```diff
@@ -148,6 +148,7 @@ export type IncludedTagCache =
   | "dynamodb"
   | "dynamodb-lite"
   | "fs-dev"
+  | "ddb-nextMode"
   | "dummy";
```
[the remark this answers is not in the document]
conico974: Good idea, but it means i need to change the name of the file as well (Because of how the resolve plugin works)

packages/open-next/src/types/overrides.ts
```diff
@@ -81,16 +86,31 @@ export type IncrementalCache = {
 
 // Tag cache
 
-export type TagCache = {
+type BaseTagCache = {
+  name: string;
+};
+
+export type NextModeTagCache = BaseTagCache & {
```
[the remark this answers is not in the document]
conico974: Probably a good idea yeah, i'll add that

packages/open-next/src/adapters/cache.ts
```diff
@@ -135,16 +62,15 @@ export default class Cache {
...
+        _tags,
         lastModified,
       );
-      if (_lastModified === -1) {
-        // If some tags are stale we need to force revalidation
-        return null;
-      }
 
-      if (value === undefined) return null;
+      if (value === undefined || _hasBeenRevalidated) return null;
```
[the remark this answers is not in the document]
conico974: Yeah i made the same suggestion to dario the other day on another PR. I need to rebase this PR

packages/open-next/src/overrides/tagCache/constants.ts
conico974: Yeah, i'll remove it.
Technically you can fetch a max of 100 item with a `BatchGetItem`, but it should never be used in such case. I'll add a `console.error` for cases with more than 100 tags

packages/open-next/src/adapters/cache.ts
```diff
@@ -185,14 +112,15 @@ export default class Cache {
...
+      const _hasBeenRevalidated = await hasBeenRevalidated(
         key,
+        tags,
         lastModified,
       );
-      if (_lastModified === -1) {
-        // If some tags are stale we need to force revalidation
-        return null;
-      }
+      if (cacheData === undefined || _hasBeenRevalidated) return null;
```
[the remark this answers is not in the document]
vicb: Should this be addressed in this PR or in a follow up?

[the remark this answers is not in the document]
conico974: Ho my bad it's already tested above, i just forgot to remove it

packages/open-next/src/adapters/cache.ts
```diff
@@ -363,32 +295,8 @@ export default class Cache {
...
-            path: key,
-            tag: tag,
-            // In case the tags are not there we just need to create them
-            // but we don't want them to return from `getLastModified` as they are not stale
-            revalidatedAt: 1,
-          })),
-        );
-      }
+
+      await this.updateTagsOnSet(key, data, ctx);
```
vicb, line 299: delete will fallback here, is that expected?

vicb (approved):
LGTM, 2 minor comments.

Thanks!

packages/open-next/src/adapters/cache.ts
conico974: Good catch, i'll add something so that it's a no op.
Ideally we should handle deleting/updating tag in this function as well, i'll adress it in another PR later.

Issue #720, opened 2025-02-03
The CI is broken


vicb:
We have this error:

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
```

See https://github.com/opennextjs/opennextjs-aws/actions/runs/13109063485/job/36568909192

Note that there is no direct dependency on `actions/upload-artifact@3`, it must be a transitive dependency

vicb:
Oops, there is actually a direct dependency...

PR #721, opened 2025-02-03
fix the e2e CI workflow

vicb:
fixes https://github.com/opennextjs/opennextjs-aws/issues/720

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -163,7 +163,7 @@ jobs:
 
       - name: Archive
         if: failure()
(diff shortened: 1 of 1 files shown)


vicb:
@conico974 could you please run the E2E action on this branch To make sure it fixes the issue?
 _(I can not do that)_

conico974 (approved):
LGTM
I can't run the CI either actually (without merging), the e2e is restricted to running on this repo and only on main.
We probably need to change that as well in the future

vicb:
Let's try then :)
Thanks for the review

PR #722, opened 2025-02-03
fix(cloudflare): enable trailiing slash redirect

vicb:
event.url is a valid URL within cloudflare vs a path otherwise

### DIFF
--- a/packages/open-next/src/core/routing/matcher.ts
+++ b/packages/open-next/src/core/routing/matcher.ts
@@ -255,8 +255,8 @@ export function handleRewrites<T extends RewriteDefinition>(
 function handleTrailingSlashRedirect(
   event: InternalEvent,
 ): false | InternalResult {
-  const url = new URL(event.url, "http://localhost");
-  const emptyBody = emptyReadableStream();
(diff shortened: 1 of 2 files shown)


packages/open-next/src/core/routing/matcher.ts
```diff
@@ -255,8 +255,8 @@ export function handleRewrites<T extends RewriteDefinition>(
 function handleTrailingSlashRedirect(
   event: InternalEvent,
 ): false | InternalResult {
-  const url = new URL(event.url, "http://localhost");
-  const emptyBody = emptyReadableStream();
+  // When rawPath is `//domain`, `url.host` would be `domain`.
```
conico974, line 258:
```suggestion
  // When rawPath is `//domain`, `url.host` would be `domain`.
  // https://github.com/opennextjs/opennextjs-aws/issues/355
```
Adding the issue here may help understanding why we need to do that

conico974 (approved):
LGTM. Just a little comment

vicb:
Thanks @conico974

PR #726, opened 2025-02-05
allow `null` to be returned as an incremental cache get result

dario-piotrowicz:
Very minor but I think it would be nice if we could return `null` as the incremental cache `get` result instead of being forced to return an empty object:
https://github.com/opennextjs/opennextjs-cloudflare/blob/642c3d2e654ef830b7957faecb1cbe26bae469ed/packages/cloudflare/src/api/kvCache.ts#L46

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -132,6 +132,8 @@ export default class Cache {
     try {
       const cachedEntry = await globalThis.incrementalCache.get(key, true);
 
+      if (cachedEntry?.value === undefined) return null;
+
       const _lastModified = await globalThis.tagCache.getLastModified(
         key,
         cachedEntry?.lastModified,
@@ -140,8 +142,6 @@ export default class Cache {
         // If some tags are stale we need to force revalidation
         return null;
       }
-
-      if (cachedEntry?.value === undefined) return null;
 
       // For cases where we don't have tags, we need to ensure that the soft tags are not being revalidated
       // We only need to check for the path as it should already contain all the tags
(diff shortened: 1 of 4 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -130,21 +130,18 @@ export default class Cache {
...
-        lastModified,
+        cachedEntry?.lastModified,
       );
       if (_lastModified === -1) {
         // If some tags are stale we need to force revalidation
         return null;
       }
 
-      if (value === undefined) return null;
+      if (cachedEntry?.value === undefined) return null;
```
conico974, line 144: For efficiency we should move this a few lines up, before checking the tag cache. 
There is no point in checking the tag cache when we should return null anyway

packages/open-next/src/adapters/cache.ts
```diff
@@ -181,18 +178,18 @@ export default class Cache {
 
   async getIncrementalCache(key: string): Promise<CacheHandlerValue | null> {
     try {
-      const { value: cacheData, lastModified } =
-        await globalThis.incrementalCache.get(key, false);
+      const cachedEntry = await globalThis.incrementalCache.get(key, false);
 
-      const meta = cacheData?.meta;
+      const meta = cachedEntry?.value?.meta;
```
conico974, line 183: Here we should return null early as well if there is no cache entry

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -157,7 +157,7 @@ export async function cacheInterceptor(
     try {
       const cachedData = await globalThis.incrementalCache.get(localizedPath);
       debug("cached data in interceptor", cachedData);
-      if (cachedData.value?.type === "app") {
+      if (cachedData?.value?.type === "app") {
```
conico974, line 160: Same, we should break early, but in this case we should return `event` not `null`

conico974 (approved):
LGTM, Just a few minor things that i noticed and that could be improved

dario-piotrowicz:
Thanks a bunch for the review @conico974 

I've applied your suggestions please have another quick look and let me know if you're happy with the changes 🙂

conico974 (approved):
Perfect Thanks

PR #728, opened 2025-02-06
use dynamic import for prerender manifest in e2e

james-elicx:
Upstreaming changes being made in cloudflare adapter.

- Uses dynamic import for pre-render manifest.
- Uses http if the host is localhost.

### DIFF
--- a/examples/app-router/app/api/isr/route.ts
+++ b/examples/app-router/app/api/isr/route.ts
@@ -1,5 +1,3 @@
-import fs from "node:fs/promises";
-import path from "node:path";
 import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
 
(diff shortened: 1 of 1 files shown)


examples/app-router/app/api/isr/route.ts
```diff
@@ -1,24 +1,32 @@
...
-    path.join(cwd, ".next/prerender-manifest.json"),
-    "utf-8",
-  );
-  const manifest = JSON.parse(prerenderManifest);
+  let manifest: { preview: { previewModeId: string } };
+  // This fails at build time when next.js tries to evaluate the route
+  try {
+    const prerenderManifest = await import(
+      // @ts-expect-error
+      /* webpackIgnore: true */ "./.next/prerender-manifest.json"
```
conico974, line 13: This won't work here, the file are not at the right place.
Probably something like that `"../../../../prerender-manifest.json"` should do the trick.

conico974 (changes requested):
This fails e2e, you should be able to test that with `next start`, it should behave the same

examples/app-router/app/api/isr/route.ts
james-elicx: To be honest I didn't run it locally because the e2e readme said the first step was to deploy the app...

If I run it with next start, it still returns a 500 for the route with this - I presume it's because of how Next.js serves the app. Unsure what will happen in the OpenNext build though as I can't run e2es locally.

vicb: What if we fallback on readFile on error and add a comment that the import is specific for `@opennextjs/cloudflare`?

If not we can diverge an reconcile later (add a comment in our version of the tests)

conico974: I found the issue actually, they were 2 issues locally : 
`host` is undefined with `next start`
and you need to add `{ assert: { type: "json" } }` in the import

james-elicx: Interesting. Thanks for taking a look. Although we'll want to use `with` instead of `assert` as it's [deprecated](https://v8.dev/features/import-attributes#deprecation-and-eventual-removal-of-assert).

examples/app-router/app/api/isr/route.ts
```diff
@@ -1,24 +1,34 @@
...
-    "utf-8",
-  );
-  const manifest = JSON.parse(prerenderManifest);
+  let manifest: { preview: { previewModeId: string } };
+  // This fails at build time when next.js tries to evaluate the route
+  try {
+    const prerenderManifest = await import(
+      // @ts-expect-error
+      /* webpackIgnore: true */ "../../../../prerender-manifest.json",
+      { assert: { type: "json" } }
```
james-elicx, line 14:
```suggestion
      { with: { type: "json" } }
```

examples/app-router/app/api/isr/route.ts
```diff
@@ -11,7 +11,7 @@ export async function GET(request: NextRequest) {
     const prerenderManifest = await import(
       // @ts-expect-error
       /* webpackIgnore: true */ "../../../../prerender-manifest.json",
-      { assert: { type: "json" } }
+      { with: { type: "json" } }
```
conico974, line 14: Looks like next doesn't like it 
image

james-elicx: :/

james-elicx: Both next build and next start work fine for me

PR #730, opened 2025-02-07
fix: make `patchAsyncStorage` work with non-mutable `node:module` modules

dario-piotrowicz:
This change allows `patchAsyncStorage` to work with the cloudflare adapter allowing us to get rid of the [code that comments the function out](
https://github.com/opennextjs/opennextjs-cloudflare/blob/6a3e35e3f79ad10607ae0acc7f88dbe9a89e8d6f/packages/cloudflare/src/cli/build/bundle-server.ts#L202-L208)

The problem with the current patch is that it updates `mod._resolveFilename` which doesn't work in the cloudflare adapter since the worker gets build by esbuild making the exports of the module (which we polyfill using unenv) non-settable:
Screenshot 2025-02-07 at 17 46 41

> [!WARNING]
> __Disclamer__: I did test the Cloudflare adapter with this change and this does allow us to use `patchAsyncStorage` without getting the runtime error anymore, I could however not see any other difference in the app I tested this with (the `app-router` app), if anyone would have suggestions on how/if I can check some effects that the patch has that would be very much appreciated 🙏

### DIFF
--- a/packages/open-next/src/core/patchAsyncStorage.ts
+++ b/packages/open-next/src/core/patchAsyncStorage.ts
@@ -1,9 +1,9 @@
-const mod = require("node:module");
+let mod = require("node:module");
 
 const resolveFilename = mod._resolveFilename;
 
 export function patchAsyncStorage() {
-  mod._resolveFilename = ((
+  const _resolveFilename = ((
     originalResolveFilename: typeof resolveFilename,
     request: string,
     parent: any,
@@ -29,4 +29,8 @@ export function patchAsyncStorage() {
 
     // We use `bind` here to avoid referencing outside variables to create potential memory leaks.
   }).bind(null, resolveFilename);
+  mod = {
+    ...mod,
+    _resolveFilename,
+  };
 }
--- a/.changeset/seven-mirrors-sleep.md
+++ b/.changeset/seven-mirrors-sleep.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: make `patchAsyncStorage` work with non-mutable `node:module` modules


packages/open-next/src/core/patchAsyncStorage.ts
```diff
@@ -1,9 +1,9 @@
-const mod = require("node:module");
+let mod = require("node:module");
 
 const resolveFilename = mod._resolveFilename;
 
 export function patchAsyncStorage() {
-  mod._resolveFilename = ((
+  const _resolveFilename = ((
```
vicb, line 6: I don't think this is correct path on workers.

`_resolveFilename` is not used on workers ([it throws](https://github.com/unjs/unenv/blob/b18778548152d2a9a019b8cb42a3367508d8960d/src/runtime/node/module.ts#L127-L129)) as we do not use dynamic require - requires/imports are resolved at build time.

vicb: Maybe we can use a build time patch instead, replacing the Next version with the patch

conico974 (commented):
I'm fine with this change as long as it works (i can't test it today)
FYI this patch is meant to make the fetch cache works when we do ISR revalidation, without it fetch cache will just be ignored.
BTW i just realized that i didn't make any e2e test for this

packages/open-next/src/core/patchAsyncStorage.ts
dario-piotrowicz: ah ok I see, yes that's a good call thanks 🙂👍 

I'm happy to close the PR if we don't see the point in this then (however it is a bit too bad that aws does something that then we need to un-do, I wonder if that could be solved somehow at some point 😕)

dario-piotrowicz:
Thanks for having a look @conico974 🫶 

But I'm going to close the PR as per @vicb's suggestion, the fix should probably be done in our adapter

PS: if you were to add an e2e test please do let me know as I would be interested to see the effect of this function, if this function is definitely necessary for fetch cache I wonder why my change was working in our adapter (unless our adapter was throwing but swallowing the errors?! 🤔)

PR #731, opened 2025-02-09
move generate message id util to own file

james-elicx:
Had to move this utility to its own file so that I could import it into a queue without getting build errors due to the original file trying to import from the configs stuff.


conico974 (approved):
LGTM Thanks

PR #732, opened 2025-02-10
refactor: make onLoad async

vicb:
ESBuild recommends callbacks [to be async](https://esbuild.github.io/plugins/#on-load) because many can run in parallel


conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review!

Issue #713, opened 2025-01-27
[Bug] Concurrent requests with the cloudflare wrappers


vicb:
There is a bug when using the cloudflare wrappers (`cloudflare-node` and `cloudflare-edge`)

It will manifest with the following logf

```
workerd/io/worker.c++:1195: warning: NOSENTRY Warning: Cross Request Promise Resolve

conico974:
Hmm i see, not sure we can easily reuse the current OpenNext ALS without some breaking change. Reusing it probably means that we'd have to initialize the ALS on every wrapper and that's not something that i'd like to do.

Maybe it would make sense to define the cloudflare context inside the wrapper directly ?
Another option would be to create one ALS only if you can't find the already defined one (i.e. `(globalThis as any)[Symbol.for("__cloudflare-context__")]`)

PR #733, opened 2025-02-10
refactor: `waitUntil` passed around via ALS

vicb:
@conico974 this a POC to fix https://github.com/opennextjs/opennextjs-aws/issues/713

I tested locally that
- `waitUntil` is available for middleware and the server
- `waitUntil` is fixed when there is a nested request inside a long running request

The handler API is changed so this is a breaking change.

What do you think?

### DIFF
--- a/.changeset/pink-papayas-smoke.md
+++ b/.changeset/pink-papayas-smoke.md
@@ -1,0 +1,25 @@
+---
+"@opennextjs/aws": minor
+---
+
+refactor: `waitUntil` passed around via ALS.
+
+BREAKING CHANGE: `waitUntil` is passed around via ALS to fix #713.
+
+`globalThis.openNextWaitUntil` is no more available, you can access `waitUntil`
+on the ALS context: `globalThis.__openNextAls.getStore()`
+
+The `OpenNextHandler` signature has changed: the second parameter was a `StreamCreator`.
+It was changed to be of type `OpenNextHandlerOptions` which has both a `streamCreator` key
+and a `waitUntil` key.
+
+If you use a custom wrapper, you need to update the call to the handler as follow:
+
+```ts
+// before
+globalThis.openNextWaitUntil = myWaitUntil;
+handler(internalEvent, myStreamCreator);
+
+// after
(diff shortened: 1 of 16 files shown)


conico974 (approved):
LGTM. 
It should probably be in a minor release though, a big warning in the release note about the breaking change for custom wrapper should be enough. 
We'll need to update the docs as well

vicb:
> LGTM. It should probably be in a minor release though, a big warning in the release note about the breaking change for custom wrapper should be enough. We'll need to update the docs as well

Thanks for the review.

I'll polish the code, add a BREAKING CHANGE changeset, and prepare a doc PR before this can be merged.

vicb:
@conico974 added a commit to polish the code and add a changeset, PTAL

packages/open-next/src/utils/promise.ts
```diff
@@ -60,30 +61,30 @@ export class DetachedPromiseRunner {
...
+  if (store?.waitUntil) {
+    store.waitUntil(promisesToAwait);
     return;
   }
   await promisesToAwait;
 }
 
 function provideNextAfterProvider() {
-  /** This should be considered unstable until `unstable_after` is stablized. */
+  /** This should be considered unstable until `unstable_after` is stabilized. */
```
conico974, line 76: `next/after` is considered stable in next 15, this comment can be removed

.changeset/pink-papayas-smoke.md
```diff
@@ -0,0 +1,25 @@
+---
+"@opennextjs/aws": minor
+---
+
+refactor: `waitUntil` passed around via ALS.
```
conico974, line 5:
```suggestion
refactor: `waitUntil` passed around via ALS  and `OpenNextHandler` signature has changed
```

conico974 (approved):
If you could just fix the biome linting issue and see my 2 little comments, other than that LGTM

packages/open-next/src/utils/promise.ts
vicb: removed

vicb:
Thanks for the review @conico974, I have addressed the comment.
I'll work on the doc  PR before this is released.

PR #734, opened 2025-02-10
feat: invoke a remove method on the queue after cache set

james-elicx:
The idea behind this is that if we have an in-memory queue that wants to de-dupe on a per-isolate basis, we could invoke a method after the incremental cache set operation to remove the path from the list of active revalidations. Incremental cache set operation use the path as the key, so this should be a decent strategy for per-isolate queues.

includes https://github.com/opennextjs/opennextjs-aws/pull/731 because i dont have permission to merge prs

### DIFF
--- a/.changeset/stupid-gifts-cough.md
+++ b/.changeset/stupid-gifts-cough.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": minor
+---
+
+Invoke a remove method on the queue after an incremental cache set for a route takes place.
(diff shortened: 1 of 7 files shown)


conico974 (commented):
I'm not sure how useful this will be once deployed honestly. Revalidation involves a `HEAD` request, how likely is it that the same isolate will send and receive the `HEAD` request.

And just awaiting the fetch call and removing there is probably as effective (and at least you're certain that you are running on the same isolate)

james-elicx:
Fair point. Makes sense.

PR #737, opened 2025-02-12
update docs for global

vicb:
Minor edits.

I don't think "Only available in main functions or in the external middleware with `enableCacheInterception` set to `true`." were accurate?

### DIFF
--- a/packages/open-next/src/types/global.ts
+++ b/packages/open-next/src/types/global.ts
@@ -66,14 +66,12 @@ declare global {
   // Needed in the cache adapter
   /**
    * The cache adapter for incremental static regeneration.
-   * Only available in main functions or in the external middleware with `enableCacheInterception` set to `true`.
-   * Defined in `createMainHandler` or in `adapter/middleware.ts`.
+   * Defined in `createMainHandler` and in `adapters/middleware.ts`.
    */
   var incrementalCache: IncrementalCache;
   /**
    * The cache adapter for the tag cache.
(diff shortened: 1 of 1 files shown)


packages/open-next/src/types/global.ts
```diff
@@ -66,14 +66,12 @@ declare global {
   // Needed in the cache adapter
   /**
    * The cache adapter for incremental static regeneration.
-   * Only available in main functions or in the external middleware with `enableCacheInterception` set to `true`.
```
conico974, line 69: It may not be very clear, but it should be accurate.
It is only defined if you're inside `server-functions` (not edge) or inside the external middleware.
It's probably the `main functions` that is confusing, if you have any suggestions

vicb: Oh actually I wanted to drop  ``with `enableCacheInterception` set to `true`.`` because it does not seem to affect whether it is available?

conico974: It should https://github.com/opennextjs/opennextjs-aws/blob/cc4f5f4ef4d04f93a4a138d8848ae472e571b297/packages/open-next/src/adapters/middleware.ts#L38-L50. These lines are supposed to be removed on build if `enableCacheInterception` is false 
On cloudflare right now it's a bit different since the external middleware is bundled with the default server

vicb: Oh right.
I'll revert those changes.

conico974 (approved):
LGTM Thanks

PR #741, opened 2025-02-14
Generic code patching

conico974:
This plugin allow to apply patch to traced files at build time, as well as provide a way to pass esbuild plugins when needed.
Probably still WIP but it's getting really close.

It also includes this patch https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/build/patches/plugins/fetch-cache-wait-until.ts as an example on how it can be used : https://github.com/opennextjs/opennextjs-aws/blob/c54167a04e0c1e81c62b2ac743a955080ecaf048/packages/open-next/src/build/patch/patchFetchCacheWaitUntil.ts

It is only available internally as of now, but we could introduce later something like an `open-next.build.ts` file that would allow users or adapter based on us to use this more easily.

@vicb @dario-piotrowicz @james-elicx i'd appreciate some early feedback on this. It should be possible to migrate a lot of the patch in cloudflare to use this, as well as use `createServerBundle` from aws directly

### DIFF
--- a/packages/open-next/src/build/createMiddleware.ts
+++ b/packages/open-next/src/build/createMiddleware.ts
@@ -81,6 +81,7 @@ export async function createMiddleware(
       additionalExternals: config.edgeExternals,
       onlyBuildOnce: forceOnlyBuildOnce === true,
       name: "middleware",
+      additionalPlugins: [],
     });
 
     installDependencies(outputPath, config.middleware?.install);
@@ -96,6 +97,7 @@ export async function createMiddleware(
       options,
       onlyBuildOnce: true,
       name: "middleware",
+      additionalPlugins: [],
     });
   }
 }
--- a/packages/open-next/src/build/createMiddleware.ts
+++ b/packages/open-next/src/build/createMiddleware.ts
@@ -81,7 +81,7 @@ export async function createMiddleware(
       additionalExternals: config.edgeExternals,
       onlyBuildOnce: forceOnlyBuildOnce === true,
       name: "middleware",
-      additionalPlugins: [],
+      additionalPlugins: () => [],
     });
 
     installDependencies(outputPath, config.middleware?.install);
@@ -97,7 +97,7 @@ export async function createMiddleware(
       options,
       onlyBuildOnce: true,
       name: "middleware",
-      additionalPlugins: [],
+      additionalPlugins: () => [],
     });
   }
 }
--- a/packages/open-next/src/build/createServerBundle.ts
+++ b/packages/open-next/src/build/createServerBundle.ts
@@ -6,6 +6,7 @@ import type { FunctionOptions, SplittedFunctionOptions } from "types/open-next";
 import type { Plugin } from "esbuild";
 import logger from "../logger.js";
 import { minifyAll } from "../minimize-js.js";
+import { ContentUpdater } from "../plugins/content-updater.js";
 import { openNextReplacementPlugin } from "../plugins/replacement.js";
 import { openNextResolvePlugin } from "../plugins/resolve.js";
 import { getCrossPlatformPathRegex } from "../utils/regex.js";
@@ -22,7 +23,7 @@ interface CodeCustomization {
   additionalCodePatches: CodePatcher[];
   // These plugins are meant to apply during the esbuild bundling process.
   // This will only apply to OpenNext code.
-  additionalPlugins: Plugin[];
+  additionalPlugins: (contentUpdater: ContentUpdater) => Plugin[];
 }
 
 export async function createServerBundle(
@@ -207,7 +208,9 @@ async function generateBundle(
 
   const disableRouting = isBefore13413 || config.middleware?.external;
 
-  const additionalPlugins = codeCustomization?.additionalPlugins ?? [];
+  const updater = new ContentUpdater(options);
+
+  const additionalPlugins = codeCustomization?.additionalPlugins(updater) ?? [];
 
   const plugins = [
     openNextReplacementPlugin({
@@ -235,6 +238,7 @@ async function generateBundle(
       overrides,
     }),
     ...additionalPlugins,
+    updater.plugin,
   ];
 
   const outfileExt = fnOptions.runtime === "deno" ? "ts" : "mjs";
--- a/packages/open-next/src/build/createServerBundle.ts
+++ b/packages/open-next/src/build/createServerBundle.ts
@@ -3,6 +3,7 @@ import path from "node:path";
 
 import type { FunctionOptions, SplittedFunctionOptions } from "types/open-next";
 
+import type { Plugin } from "esbuild";
 import logger from "../logger.js";
 import { minifyAll } from "../minimize-js.js";
 import { openNextReplacementPlugin } from "../plugins/replacement.js";
@@ -14,11 +15,19 @@ import { copyTracedFiles } from "./copyTracedFiles.js";
 import { generateEdgeBundle } from "./edge/createEdgeBundle.js";
 import * as buildHelper from "./helper.js";
 import { installDependencies } from "./installDeps.js";
-import { applyCodePatches, type CodePatcher } from "./patch/codePatcher.js";
+import { type CodePatcher, applyCodePatches } from "./patch/codePatcher.js";
+
+interface CodeCustomization {
+  // These patches are meant to apply on user and next generated code
+  additionalCodePatches: CodePatcher[];
+  // These plugins are meant to apply during the esbuild bundling process.
+  // This will only apply to OpenNext code.
+  additionalPlugins: Plugin[];
+}
 
 export async function createServerBundle(
   options: buildHelper.BuildOptions,
-  additionalCodePatches: CodePatcher[] = [],
+  codeCustomization?: CodeCustomization,
 ) {
   const { config } = options;
   const foundRoutes = new Set<string>();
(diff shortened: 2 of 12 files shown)


vicb:
Glanced at the code, looks great.
I'll take a deeper look early next week.

packages/open-next/src/build/createServerBundle.ts
```diff
@@ -36,7 +40,7 @@ export async function createServerBundle(options: buildHelper.BuildOptions) {
     if (fnOptions.runtime === "edge") {
       await generateEdgeBundle(name, options, fnOptions);
```
vicb, line 41: Should we pass patchers for the edge bundle too?

packages/open-next/src/build/createServerBundle.ts
```diff
@@ -153,14 +158,26 @@ async function generateBundle(
...
     outputPath,
     fnOptions.routes ?? ["app/page.tsx"],
     isBundled,
   );
 
+  await applyCodePatches(options, Array.from(tracedFiles), manifests, [
+    // TODO: create real code patchers here
+    {
+      name: "fakePatchChunks",
+      filter: /chunks\/\d*\.js/,
```
vicb, line 173:
```suggestion
      filter: /chunks\/\d+\.js/,
```
Also use platform independent regex

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,114 @@
...
+  filePath: string;
+  // All js files that will be included in this specific server function
+  tracedFiles: string[];
+  // All next.js manifest that are present at runtime - Key relative to `.next` folder
+  manifests: Record<string, any>;
+}) => Promise<string>;
+
+export interface CodePatcher {
+  name: string;
+  filter: RegExp | VersionedField<RegExp>[];
```
vicb, line 25: We use `filter` to match ESBuild but `pathFilter` would be nice IMO (vs `contentFilter`)

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,114 @@
...
+  contentFilter?: RegExp | VersionedField<RegExp>[];
+  patchCode: PatchCodeFn | VersionedField<PatchCodeFn>[];
```
vicb, line 27: I'm afraid having multiple `VersionnedField` (with potentially different versions) might be a mess, should `before/after` be member of `CodePatcher` to make things simpler?

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,114 @@
...
+): T[] {
+  const result: T[] = [];
+  for (const field of fields) {
+    if (
+      field.before &&
+      field.after &&
+      buildHelper.compareSemver(version, field.before) >= 0 &&
+      buildHelper.compareSemver(version, field.after) < 0
+    ) {
+      result.push(field.field);
```
vicb, line 42: Nits:
Does the result need to be an array?
You can also return early to avoid `else`

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export async function applyCodePatches(
+  buildOptions: buildHelper.BuildOptions,
+  tracedFiles: string[],
+  manifests: Record<string, any>,
+  codePatcher: CodePatcher[],
+) {
+  await Promise.all(
+    tracedFiles.map(async (filePath) => {
+      const nextVersion = buildOptions.nextVersion;
```
vicb, line 66: nit: move that before the loop

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,114 @@
...
+      // We check the filename against the filter to see if we should apply the patch
+      const patchToPotentiallyApply = codePatcher.filter((patch) => {
+        const filters = Array.isArray(patch.filter)
```
vicb, line 69: `filters = [patch.filter].flat()`

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,114 @@
...
+  buildOptions: buildHelper.BuildOptions,
+  tracedFiles: string[],
+  manifests: Record<string, any>,
+  codePatcher: CodePatcher[],
+) {
+  await Promise.all(
+    tracedFiles.map(async (filePath) => {
+      const nextVersion = buildOptions.nextVersion;
+      // We check the filename against the filter to see if we should apply the patch
+      const patchToPotentiallyApply = codePatcher.filter((patch) => {
```
vicb, line 68: nit: `patchToPotentiallyApply` -> `patchMatchingPath`

vicb (commented):
That's quite nice and generic!

I understand that `VersionnedField[]` allow for a single patch to cover different versions but I wonder if it's worth the burden (maintenance, readability).

Maybe only adding `before/after` to a patch could be enough?

Then i.e. `pathAsyncLocalStroage` could be an array of code patches for different versions

vicb:
One thing that would be nice is for patched to be able to report a status (messages, stats).
It could be handy to debug the build.
That's something that could be added in a follow up PR.

packages/open-next/src/build/createServerBundle.ts
conico974: I don't think so, or at least not in its current form. For the edge runtime there is not really a concept of traced files, and given that we are already using esbuild for all the files it probably make more sense to use esbuild plugin there.
In fact we could also allow to pass additional esbuild plugin in server bundle (i think that's all that's really needed for you to use `createServerBundle` directly right ?

packages/open-next/src/build/patch/codePatcher.ts
conico974: The idea behind it is to more easily accomodate small changes while keeping the reason why we need things like that.
For example if next just move a file, we just need to make a small change to `filter` and it makes it very clear why.
But i admit that it might be a little bit harder to understand at first glance than having just multiple `CodePatcher`

packages/open-next/src/build/createServerBundle.ts
vicb: Makes sense.

> i think that's all that's really needed for you to use createServerBundle directly right ?

One other thing is that we include the edge ON config.

I am thinking that maybe we should have a `deployTo: cloudflare|node` in the config so that we could handle in this in a generic way. I still need to think a bit more about that.

FYI what I'd like to look at next:
- change `internalEvent.url` to always be resolved URL (vs a relative URL) - this is blocking us from using bundled middleware because external middleware needs a resolved URL vs server bundle require a path. Using a bundled middleware would reduce the bundle size when the middleware and server are in the same worker
- What I'm thinking is that we could compile an `init.js` and use that instead of the ESBuild banners. That init would depend on the platform (i.e. `deployTo`)
- With cloudflare workers, the main reason why I'm looking at this is to allow accessing `process.env` and `getCloudflareContext()` at top level.

Sorry if this is no very detailed and a little confusing, I'm still thinking about how all of this should be done. I know where I want to go but not exactly sure about the path. It should hopefully become clearer in the next few days.

conico974: IMO this shouldn't be in the config but something in the wrapper itself, right now it returns things like `supportStreaming`, we could add a `deployTo` or `platformRuntime` field there. The wrapper is already dependent on the runtime it is deployed to anyway

packages/open-next/src/build/createMiddleware.ts
```diff
@@ -66,6 +66,7 @@ export async function createMiddleware(
       additionalExternals: config.edgeExternals,
       onlyBuildOnce: forceOnlyBuildOnce === true,
       name: "middleware",
+      additionalPlugins: [],
```
vicb, line 69: one caveat with ESBuild plugins is that `onLoad`s are called only until one returned a value (!== undefined).
We implemented a [`ContentUpdated`](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/build/patches/plugins/content-updater.ts).It would allow updating a file i.e. according to the Next version and the platform (workers vs node) with 2 different plugins.

conico974: I'm not sure i totally understand what you suggest here. You'd want to move `ContentUpdater` here with a few additional check like `VersionedField` ?  And instead of `additionalPlugins` being `Plugin[]`, it would be `((updater: ContentUpdater) => Plugin)[]` ?

Anyway after merging this and using it in cloudflare, you'll likely don't need the `ContentUpdater` anymore, pretty much all the plugins that use it today (and some that don't) could be migrated to `CodePatcher`

packages/open-next/src/build/patch/astCodePatcher.ts
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Patches the code from by applying the rule.
+ *
+ * This function is mainly for on off edits and tests,
+ * use `getRuleEdits` to apply multiple rules.
+ *
+ * @param code The source code
+ * @param rule The astgrep rule (yaml or NapiConfig)
+ * @param lang The language used by the source code
+ * @param lang Whether to apply the rule only once
```
sommeeeer, line 96: `once`

conico974:
Replaced by #781

PR #745, opened 2025-02-18
fix: add correct typing to tagCache in initializationFunction

sommeeeer:
Turns out we didn't have the correct typing for `tagCache` in `initializationFunction`. We also didn't provide the override to `compileTagCacheProvider`.


conico974 (approved):
LGTM Thanks

PR #747, opened 2025-02-18
Update package.json exports for types imports

nhardy:
Updates the `package.json` `exports` field to split into `default` and `types` to fix issue with TypeScript being unable to reference types such as:
```ts
import type { OpenNextConfig } from "open-next/types/open-next";
```
Error message: `Cannot find module 'open-next/types/open-next' or its corresponding type declarations. ts(2307)`

### DIFF
--- a/packages/open-next/package.json
+++ b/packages/open-next/package.json
@@ -22,13 +22,6 @@
       "types": "./dist/*.d.ts"
     }
   },
-  "typesVersions": {
-    "*": {
-      "*": [
-        "dist/*"
-      ]
-    }
-  },
   "keywords": [],
   "author": "",
   "files": [
--- a/packages/open-next/package.json
+++ b/packages/open-next/package.json
@@ -17,7 +17,8 @@
     "dev": "tsc -w"
   },
   "exports": {
-    "./*": "./dist/*"
+    "default": "./dist/*",
+    "types": "./dist/*.d.ts"
   },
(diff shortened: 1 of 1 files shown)


packages/open-next/package.json
```diff
@@ -17,7 +17,8 @@
     "dev": "tsc -w"
   },
   "exports": {
-    "./*": "./dist/*"
+    "default": "./dist/*",
+    "types": "./dist/*.d.ts"
```
vicb, line 21: Shouldn't this be
```suggestion
    "./*": {
      "default": "./dist/*",
      "types": "./dist/*.d.ts"
    }
```

packages/open-next/package.json
```diff
@@ -17,7 +17,8 @@
...
   },
   "typesVersions": {
```
vicb, line 23: Is `typesVersion` useful here? 
Maybe we can drop that while we are updating this file

packages/open-next/package.json
nhardy: You are totally correct. Missed copying the full diff from my patch-package patch. 😅

packages/open-next/package.json
nhardy: You're right here too. Given the version is wildcard, the exports types cover this.

conico974 (approved):
LGTM Thanks

PR #748, opened 2025-02-19
return meta files from createAssets()

james-elicx:
I want to re-use the value instead of importing the generated file.

I also had to add back the `typesVersions` that was removed in https://github.com/opennextjs/opennextjs-aws/pull/747 as it broke builds and imports for the cloudflare adapter.

image
image

### DIFF
--- a/packages/open-next/package.json
+++ b/packages/open-next/package.json
@@ -22,6 +22,13 @@
       "types": "./dist/*.d.ts"
     }
   },
+  "typesVersions": {
+    "*": {
+      "*": [
+        "dist/*"
+      ]
+    }
+  },
   "keywords": [],
   "author": "",
   "files": [
--- a/packages/open-next/src/build/createAssets.ts
+++ b/packages/open-next/src/build/createAssets.ts
@@ -2,7 +2,7 @@ import fs from "node:fs";
 import path from "node:path";
 
 import logger from "../logger.js";
-import type { TagCacheMetaFile } from "../types/open-next.js";
+import type { TagCacheMetaFile } from "../types/cache.js";
 import { isBinaryContentType } from "../utils/binary.js";
 import * as buildHelper from "./helper.js";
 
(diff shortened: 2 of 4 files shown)


packages/open-next/package.json
```diff
@@ -22,6 +22,13 @@
       "types": "./dist/*.d.ts"
     }
   },
+  "typesVersions": {
```
vicb, line 25: Would `"types": "dist/*"` work here?

Could the problem be that we import `path.js` instead of `path` (would be awesome to be able to drop the `.js` but it might be a breaking change)

Anyway do not spend too much time on this if you have a working config.

packages/open-next/src/types/open-next.ts
vicb, line 1: Should this rather go in `types/cache.ts`?

packages/open-next/src/build/createAssets.ts
```diff
@@ -244,5 +241,5 @@ export function createCacheAssets(options: buildHelper.BuildOptions) {
     ({ relativePath }) => !relativePath.endsWith(".cache"),
   );
 
-  return { useTagCache };
+  return { useTagCache, metaFiles };
```
vicb, line 244: please update the comment on `createCacheAssets`

vicb (approved):
LGTM with a few nits

packages/open-next/src/types/open-next.ts
conico974: I think so too

packages/open-next/package.json
conico974: > Could the problem be that we import path.js instead of path (would be awesome to be able to drop the .js but it might be a breaking change)

I don't think so, i've tested this on a test repo that i had and it worked. I was even able to drop the `.js`.

If we need `typesVersions` to not break compatibility i'm fine with it

conico974 (approved):
Same comment as vic, but other than that LGTM

packages/open-next/package.json
james-elicx: unfortunately same problem

packages/open-next/src/types/open-next.ts
james-elicx: done. i thought that one was supposed to be things that came from next.js instead of the custom open-next types so i put them in the other one at first.

PR #749, opened 2025-02-20
Relax the type of OpenNextConfig.middleware.external

vicb:
minor change but there should be not real reason to force this to be true?


conico974 (commented):
Actually there is a reason, all the other options in middleware here will have no effect if used without the middleware being external. 
It was to make it clear that this should only be used for bundling the middleware separately (looks like i failed :smile: )

vicb:
makes sense!

PR #750, opened 2025-02-21
add: s3-lite override for imageLoader

sommeeeer:
- Uses `aws4fetch`
- New env variable `BUCKET_REGION`. Falls back to `AWS_REGION` ?? `AWS_DEFAULT_REGION` if undefined. This way no change in IAC is needed for most users. 

Add an `open-next.config.ts` with these options: 
```ts
const config = {
  default: {},
   imageOptimization: {
     loader: 's3-lite',
   },
} satisfies OpenNextConfig;

export default config;
```

### DIFF
--- a/packages/open-next/src/build/createImageOptimizationBundle.ts
+++ b/packages/open-next/src/build/createImageOptimizationBundle.ts
@@ -89,6 +89,7 @@ export async function createImageOptimizationBundle(
           "const require = topLevelCreateRequire(import.meta.url);",
           "import bannerUrl from 'url';",
           "const __dirname = bannerUrl.fileURLToPath(new URL('.', import.meta.url));",
+          "globalThis.internalFetch = fetch;",
         ].join("\n"),
       },
     },
--- a/packages/open-next/src/build/createImageOptimizationBundle.ts
+++ b/packages/open-next/src/build/createImageOptimizationBundle.ts
@@ -89,7 +89,6 @@ export async function createImageOptimizationBundle(
           "const require = topLevelCreateRequire(import.meta.url);",
           "import bannerUrl from 'url';",
           "const __dirname = bannerUrl.fileURLToPath(new URL('.', import.meta.url));",
-          "globalThis.internalFetch = fetch;",
         ].join("\n"),
       },
     },
--- a/packages/open-next/src/overrides/imageLoader/s3-lite.ts
+++ b/packages/open-next/src/overrides/imageLoader/s3-lite.ts
@@ -1,0 +1,69 @@
+import { AwsClient } from "aws4fetch";
+import { Readable } from "node:stream";
+import type { ReadableStream } from "node:stream/web";
+
+import type { ImageLoader } from "types/overrides";
+import { FatalError } from "utils/error";
+import { customFetchClient } from "utils/fetch";
+
+
+let awsClient: AwsClient | null = null;
+
+const getAwsClient = () => {
+  if (awsClient) {
+    return awsClient;
+  }
+  awsClient = new AwsClient({
+    accessKeyId: process.env.AWS_ACCESS_KEY_ID!,
+    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY!,
+    sessionToken: process.env.AWS_SESSION_TOKEN,
+    region: process.env.BUCKET_REGION,
+  });
+  return awsClient;
+};
+
+const { BUCKET_NAME, BUCKET_KEY_PREFIX, BUCKET_REGION } = process.env;
+
+function ensureEnvExists() {
+  if (!BUCKET_NAME || !BUCKET_REGION || BUCKET_KEY_PREFIX) {
+    throw new Error("Bucket name, region and key prefix must be defined!");
+  }
+}
+
+const awsFetch = async (key: string, options: RequestInit) => {
+  const client = getAwsClient();
+  const url = `https://${BUCKET_NAME}.s3.${BUCKET_REGION}.amazonaws.com/${key}`;
+  return customFetchClient(client)(url, options);
+};
+
+
+const s3Loader: ImageLoader = {
+  name: "s3",
+  load: async (key: string) => {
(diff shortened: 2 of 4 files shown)


packages/open-next/src/build/createImageOptimizationBundle.ts
```diff
@@ -89,6 +89,7 @@ export async function createImageOptimizationBundle(
           "const require = topLevelCreateRequire(import.meta.url);",
           "import bannerUrl from 'url';",
           "const __dirname = bannerUrl.fileURLToPath(new URL('.', import.meta.url));",
+          "globalThis.internalFetch = fetch;",
```
conico974, line 92: No need for that

packages/open-next/src/overrides/imageLoader/s3-lite.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const getAwsClient = () => {
+  if (awsClient) {
+    return awsClient;
+  }
+  awsClient = new AwsClient({
+    accessKeyId: process.env.AWS_ACCESS_KEY_ID!,
+    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY!,
+    sessionToken: process.env.AWS_SESSION_TOKEN,
+    region: process.env.BUCKET_REGION,
```
conico974, line 19: You could fallback to the default aws region https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime, this way no change needed in IAC for most user

packages/open-next/src/overrides/imageLoader/s3-lite.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+const awsFetch = async (key: string, options: RequestInit) => {
+  const client = getAwsClient();
+  const url = `https://${BUCKET_NAME}.s3.${BUCKET_REGION}.amazonaws.com/${key}`;
+  return customFetchClient(client)(url, options);
+};
+
+const s3Loader: ImageLoader = {
+  name: "s3",
```
conico974, line 39:
```suggestion
  name: "s3-lite",
```

packages/open-next/src/overrides/imageLoader/s3-lite.ts
```diff
@@ -0,0 +1,67 @@
...
+    region: process.env.BUCKET_REGION,
+  });
+  return awsClient;
+};
+
+const { BUCKET_NAME, BUCKET_KEY_PREFIX, BUCKET_REGION } = process.env;
+
+function ensureEnvExists() {
+  if (!(BUCKET_NAME || BUCKET_REGION || BUCKET_KEY_PREFIX)) {
+    throw new Error("Bucket name, region and key prefix must be defined!");
```
conico974, line 28:
```suggestion
    throw new FatalError("Bucket name, region and key prefix must be defined!");
```

packages/open-next/src/overrides/imageLoader/s3-lite.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+const awsFetch = async (key: string, options: RequestInit) => {
+  const client = getAwsClient();
+  const url = `https://${BUCKET_NAME}.s3.${BUCKET_REGION}.amazonaws.com/${key}`;
+  return customFetchClient(client)(url, options);
```
conico974, line 35: No need for the custom fetch client here, it is only used to avoid issues with the fetch cache (which does not exist here). Just do a `client.fetch()`

packages/open-next/src/overrides/imageLoader/s3-lite.ts
```diff
@@ -0,0 +1,67 @@
...
+    const keyPrefix = BUCKET_KEY_PREFIX?.replace(/^\/|\/$/g, "");
+    const response = await awsFetch(
+      keyPrefix
+        ? `${keyPrefix}/${key.replace(/^\//, "")}`
+        : key.replace(/^\//, ""),
+      {
+        method: "GET",
+      },
+    );
+
```
conico974, line 51: You need to handle cases where it's not a 200 here with an appropriate error (i.e. not found) https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_Errors
Doing something similar to this https://github.com/opennextjs/opennextjs-aws/blob/9f3720f4309d95b484ae6ae3e9ed7f7342424339/packages/open-next/src/overrides/incrementalCache/s3-lite.ts#L55-L60 is probably enough

packages/open-next/src/overrides/imageLoader/s3-lite.ts
```diff
@@ -49,6 +50,15 @@ const s3Loader: ImageLoader = {
       },
     );
 
+    if (response.status === 404) {
+      throw new IgnorableError("The specified key does not exist.");
```
conico974, line 58: The error should not be ignored here. All these errors are `FatalError`, they can be ignored in the incremental cache, not here

packages/open-next/src/overrides/imageLoader/s3-lite.ts
```diff
@@ -3,11 +3,14 @@ import type { ReadableStream } from "node:stream/web";
...
 import type { ImageLoader } from "types/overrides";
-import { FatalError } from "utils/error";
-import { customFetchClient } from "utils/fetch";
+import { FatalError, IgnorableError, RecoverableError } from "utils/error";
 
 let awsClient: AwsClient | null = null;
 
+const { BUCKET_NAME, BUCKET_KEY_PREFIX } = process.env;
+const BUCKET_REGION =
+  process.env.BUCKET_REGION ?? process.env.AWS_DEFAULT_REGION;
```
conico974:
```suggestion
  process.env.BUCKET_REGION ?? process.env.AWS_REGION ?? process.env.AWS_DEFAULT_REGION;
```

conico974 (approved):
LGTM Thanks

Issue #719, opened 2025-02-03
Make event.url consistent across platforms (URL vs path)


vicb:
We have an issue with trailing slashes, see https://github.com/opennextjs/opennextjs-cloudflare/issues/312

The reason why this doesn't work is this code:

https://github.com/opennextjs/opennextjs-aws/blob/14b81827f9078b98e32115fb5cfe706d03d64537/packages/open-next/src/core/routing/matcher.ts#L258-L269

in cloudflare, `event.url` is `http://localhost:8787/...` so `url.host !== "localhost"` is true (the `host` is `localhost:8787`).

I think that `event.url` is not usually an URL but a path in `@opennextjs/aws`.

However in `@opennextjs/cloudflare`, `event.url` is actually an URL.

See also this code in `cloudflare-node.ts`:

https://github.com/opennextjs/opennextjs-aws/blob/14b81827f9078b98e32115fb5cfe706d03d64537/packages/open-next/src/overrides/wrappers/cloudflare-node.ts#L33-L38

@conico974 do you have an idea on how to best fix this?

Thanks!

conico974:
Maybe just changing this line would be enough to fix it temporarily https://github.com/opennextjs/opennextjs-aws/blob/14b81827f9078b98e32115fb5cfe706d03d64537/packages/open-next/src/core/routing/matcher.ts#L258
We don't really care about the actual value here, we could replace it with something like `http://n` (same way next is doing it in some places)

I think a more permanent fix would be to ensure that url always have a protocol and an host (that we could replace by dummy values if we can't infer them ). We probably need to change a bunch of things in the routing layer (and the converters) for this to work

vicb:
IIUC, I don't think that would help?

<img width="817" alt="Image" src="https://github.com/user-attachments/assets/29954d8d-5162-4cb8-b16f-6e4909855a2f" />

conico974:
Ho sorry, yeah, we'll need to change the if below as well (i.e. `if url.host !== "n"` instead of localhost)
The issue we had with that is that by using trailing slash redirect stuff, an attacker could trigger a redirect to an external url (because in lambda url has no protocol/host).

vicb:
I don't think that would work as `url.host` is `localhost:8787` on my example above :)

conico974:
Ho yeah nevermind.
Replacing with `event.rawPath` in this same line should work though

vicb:
> Replacing with `event.rawPath` in this same line should work though

[...]

PR #752, opened 2025-02-25
refactor: internalEvent.url is a full URL

vicb:
fixes https://github.com/opennextjs/opennextjs-aws/issues/719

edit: One thing I need to check is the base path, we should probably add it when constructing URLs (i.e. 404, 500)

@conico974 could you please review and maybe run the aws integration tests when this looks ok?
cf tests are green with this PR.
Thanks!

### DIFF
--- a/packages/open-next/src/core/routing/matcher.ts
+++ b/packages/open-next/src/core/routing/matcher.ts
@@ -14,6 +14,7 @@ import { emptyReadableStream, toReadableStream } from "utils/stream";
 import { debug } from "../../adapters/logger";
 import { localizePath } from "./i18n";
 import {
+  constructNextUrl,
   convertFromQueryString,
   convertToQueryString,
   escapeRegex,
@@ -171,7 +172,7 @@ export function handleRewrites<T extends RewriteDefinition>(
   event: InternalEvent,
   rewrites: T[],
 ) {
-  const { rawPath, headers, query, cookies } = event;
+  const { rawPath, headers, query, cookies, url } = event;
   const localizedRawPath = localizePath(event);
   const matcher = routeHasMatcher(headers, cookies, query);
   const computeHas = computeParamHas(headers, cookies, query);
@@ -185,7 +186,7 @@ export function handleRewrites<T extends RewriteDefinition>(
   });
   let finalQuery = query;
 
-  let rewrittenUrl = rawPath;
+  let rewrittenUrl = url;
   const isExternalRewrite = isExternal(rewrite?.destination);
   debug("isExternalRewrite", isExternalRewrite);
   if (rewrite) {
@@ -202,16 +203,12 @@ export function handleRewrites<T extends RewriteDefinition>(
     // %2B does not get interpreted as a space in the URL thus breaking the query string.
     const encodePlusQueryString = queryString.replaceAll("+", "%20");
     debug("urlParts", { pathname, protocol, hostname, queryString });
-    const toDestinationPath = compile(escapeRegex(pathname ?? "") ?? "");
-    const toDestinationHost = compile(escapeRegex(hostname ?? "") ?? "");
-    const toDestinationQuery = compile(
-      escapeRegex(encodePlusQueryString ?? "") ?? "",
-    );
+    const toDestinationPath = compile(escapeRegex(pathname));
+    const toDestinationHost = compile(escapeRegex(hostname));
+    const toDestinationQuery = compile(escapeRegex(encodePlusQueryString));
     const params = {
       // params for the source
-      ...getParamsFromSource(match(escapeRegex(rewrite?.source) ?? ""))(
-        pathToUse,
-      ),
+      ...getParamsFromSource(match(escapeRegex(rewrite.source)))(pathToUse),
       // params for the has
       ...rewrite.has?.reduce((acc, cur) => {
         return Object.assign(acc, computeHas(cur));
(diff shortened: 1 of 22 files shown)


packages/open-next/src/core/routing/matcher.ts
```diff
@@ -245,7 +245,10 @@ export function handleRewrites<T extends RewriteDefinition>(
     internalEvent: {
       ...event,
       rawPath: rewrittenUrl,
-      url: `${rewrittenUrl}${convertToQueryString(finalQuery)}`,
+      url: new URL(
+        `${rewrittenUrl}${convertToQueryString(finalQuery)}`,
+        new URL(event.url),
```
conico974, line 250: Here it fails in e2e because `event.url` is incorrect.

packages/open-next/src/overrides/converters/aws-apigw-v1.ts
```diff
@@ -57,13 +57,14 @@ async function convertFromAPIGatewayProxyEvent(
   event: APIGatewayProxyEvent,
 ): Promise<InternalEvent> {
   const { path, body, httpMethod, requestContext, isBase64Encoded } = event;
+  const headers = normalizeAPIGatewayProxyEventHeaders(event);
   return {
     type: "core",
     method: httpMethod,
     rawPath: path,
-    url: path + normalizeAPIGatewayProxyEventQueryParams(event),
+    url: `https://${headers.host ?? "on"}${path}${normalizeAPIGatewayProxyEventQueryParams(event)}`,
```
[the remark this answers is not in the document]
vicb: Ok I'll apply the change in aws converters.
Could the code in request handler (your link) be deleted then?

packages/open-next/src/core/routing/matcher.ts
vicb: Do you know what test fails?

conico974: https://github.com/opennextjs/opennextjs-aws/blob/674dce6937266ae7f838d6343616cbb254a174af/packages/tests-e2e/tests/appRouter/config.redirect.test.ts#L17
https://github.com/opennextjs/opennextjs-aws/blob/674dce6937266ae7f838d6343616cbb254a174af/packages/tests-e2e/tests/appRouter/config.redirect.test.ts#L40
https://github.com/opennextjs/opennextjs-aws/blob/674dce6937266ae7f838d6343616cbb254a174af/packages/tests-e2e/tests/appRouter/config.redirect.test.ts#L51
And these 2 https://github.com/opennextjs/opennextjs-aws/blob/674dce6937266ae7f838d6343616cbb254a174af/packages/tests-e2e/tests/pagesRouter/redirect.test.ts#L11-L25

packages/open-next/src/overrides/converters/aws-apigw-v1.ts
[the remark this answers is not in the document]
conico974: Honestly i'm not sure how to handle this one properly. We need the headers to be updated, but only if there is a `x-forwarded-host`

packages/open-next/src/overrides/converters/aws-apigw-v1.ts
```diff
@@ -62,7 +62,7 @@ async function convertFromAPIGatewayProxyEvent(
     type: "core",
     method: httpMethod,
     rawPath: path,
-    url: `https://${headers.host ?? "on"}${path}${normalizeAPIGatewayProxyEventQueryParams(event)}`,
+    url: `https://${headers["x-forwarded-host"] ?? "on"}${path}${normalizeAPIGatewayProxyEventQueryParams(event)}`,
```
conico974, line 65: This is not 100% correct either, we could create an helper function that would retrieve the host from the headers.
If this header is not set (which can be the case), we should fallback to the `host` header

packages/open-next/src/overrides/converters/aws-apigw-v1.ts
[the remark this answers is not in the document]
vicb: This could be done in the converter as it is aws specific, no?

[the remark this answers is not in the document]
conico974: Not entirely, this is kind of a standard https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host.
And this will likely cause trouble when used with the `node` converter and wrapper as well.
Maybe a better solution would be to modify the url in the request handler as well. If we have a `x-forwarded-host` we change both the host header and the url to use this header. 
We may have to move this to the routing handler though (in case of an external middleware)

packages/open-next/src/overrides/converters/aws-apigw-v1.ts
vicb: Do you mean an helper to `headers["x-forwarded-host"] ?? headers.on ?? "on"` ?

conico974: That was my original idea, but i think there is a better solution (https://github.com/opennextjs/opennextjs-aws/pull/752#discussion_r1969749991). In the converters we keep `headers.host ?? "on"`, but in the routing handler if we have a `x-forwarded-host` header, we change both the host header and the `internalEvent.url`

vicb:
I have pushed some commits to integrate the feedback.

@conico974 Could you please run the test in debug mode and report the `rewrittenUrl` is they are still failing?

```ts
debug("rewrittenUrl", { rewrittenUrl, finalQuery, isUsingParams });
```

Thanks!

conico974 (approved):
I've run the e2e test and everything is working as expected.
LGTM Thanks

vicb:
Thanks a lot @conico974, I will double check the basePath thing before merging.

vicb:
I tested using a `basePath` on cloudflare and it works the same as with `next dev`, merging

Issue #763, opened 2025-03-03
[e2e] Add a test for API routes on the pages router


vicb:
See: https://github.com/opennextjs/opennextjs-cloudflare/pull/431/files#r1976591704

conico974:
Fixed in #764 and #765

PR #764, opened 2025-03-03
add: e2e for api routes in pages router

sommeeeer:
For #763, add e2e for API routes in `pagesRouter` example app:

- Hello world in `/api/hello.ts` 
- Dynamic API route in `/api/dynamic/[slug].ts` 
- Catch all route in `/api/dynamic/catch-all/[...slug].ts` 
- Catch all optional route in `/api/dynamic/catch-all-optional/[[...slug]].ts` 

I also wanted to have a last predefined route which would take precedence over those above as stated in Next docs [caveats](https://nextjs.org/docs/pages/building-your-application/routing/api-routes#caveats). Turns out its not taking precedence as stated in the docs. Same behavior in `next start`. I will open an issue in Next for this. For now I just added `.skip()` in the e2e.

### DIFF
--- a/examples/pages-router/src/pages/api/dynamic/[slug].ts
+++ b/examples/pages-router/src/pages/api/dynamic/[slug].ts
@@ -1,0 +1,6 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+
+export default function handler(req: NextApiRequest, res: NextApiResponse) {
+  const { slug } = req.query;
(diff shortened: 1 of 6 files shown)


vicb (approved):
Thanks 🙏
Would be nice to have the next bug ID before merging

conico974 (approved):
LGTM Thanks for that

PR #765, opened 2025-03-04
fix: api routes in pages router e2e

sommeeeer:
Turns out I was using `route.ts`(app router) instead of `index.ts`(pages router). The precedence route works as expected now, both deployed to AWS and in `next start`. I closed the issue in Next aswell.


conico974 (approved):
LGTM Thanks

PR #767, opened 2025-03-04
Fix rewrite query not forwarded and api destination

### LINKED ISSUE #766: next.config rewrites don't get handled properly
Me again!

When setting up `next.config` rewrites, they don't seem to be playing nicely with query parameters with pages router (Could be app router as well, but untested)

## Versions
**Next version:** 5.2.1
**Open-next version:** 3.5.1 

Here's an example rewrites I've put together;

http://github.com/nicholas-c/next-15-open-next-example/

```
  reactStrictMode: true,
  async rewrites() {
    return [
      {
        source: "/about",
        destination: "/",
      },
      {
        source: "/account/register",
        destination: "/api/auth?tab=register",
      },
      {
        source: "/account/login",
        destination: "/api/auth",
      },
    ];
  }
}
```

Image

Locally this works as intended, both with `yarn dev` and `yarn start` and yields the expected results.

However when built and deployed with Open-next the query parameter gets lost.

Image

https://next-15-en.nich-carter-qza3ng.gymshark-sandbox.tools/account/register
https://next-15-en.nich-carter-qza3ng.gymshark-sandbox.tools/api/auth?tab=register

Additionally to this, because I'm piggy backing on a previous test repo I have for this; When I add back in i18n config into `next.config` `/account/register` starts to 404 in the example above.

conico974:
This PR fix #766

They were 2 different issues described there: 
- Query params defined in the destination of a rewrite were not forwarded as expected
-  Internal destination that are api route should be stripped from the appended locale. (If `locale` is not set to false, next will always transform in rewrite with an appended locale, but it expects the api route to be not locale prefixed)

@nicholas-c Could you confirm that it fixes your issue

### DIFF
--- a/packages/open-next/src/core/routing/matcher.ts
+++ b/packages/open-next/src/core/routing/matcher.ts
@@ -227,11 +227,23 @@ export function handleRewrites<T extends RewriteDefinition>(
       rewrittenHost = unescapeRegex(toDestinationHost(params));
       rewrittenQuery = unescapeRegex(toDestinationQuery(params));
     }
+
+    // We need to strip the locale from the path if it's a local api route
+    if (NextConfig.i18n && !isExternalRewrite) {
+      const strippedPathLocale = rewrittenPath.replace(
+        new RegExp(`^/(${NextConfig.i18n.locales.join("|")})`),
+        "",
+      );
+      if (strippedPathLocale.startsWith("/api/")) {
+        rewrittenPath = strippedPathLocale;
+      }
+    }
+
     rewrittenUrl = isExternalRewrite
       ? `${protocol}//${rewrittenHost}${rewrittenPath}`
       : new URL(rewrittenPath, event.url).href;
 
-    // Should we merge the query params or use only the ones from the rewrite?
+    // We merge query params from the source and the destination
     finalQuery = {
       ...query,
       ...convertFromQueryString(rewrittenQuery),
@@ -243,6 +255,7 @@ export function handleRewrites<T extends RewriteDefinition>(
   return {
     internalEvent: {
       ...event,
+      query: finalQuery,
       rawPath: new URL(rewrittenUrl).pathname,
       url: rewrittenUrl,
     },
--- a/.changeset/red-starfishes-wave.md
+++ b/.changeset/red-starfishes-wave.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Fix api rewrite destination with i18n and query not forwarded on rewrite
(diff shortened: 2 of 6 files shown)


nicholas-c:
> @nicholas-c Could you confirm that it fixes your issue

Absolutely, I'll patch my test link in an hour or two - Will update here when done

packages/open-next/src/core/routing/matcher.ts
```diff
@@ -227,11 +227,23 @@ export function handleRewrites<T extends RewriteDefinition>(
       rewrittenHost = unescapeRegex(toDestinationHost(params));
       rewrittenQuery = unescapeRegex(toDestinationQuery(params));
     }
+
+    // We need to strip the locale from the path if it's a local api route
+    if (NextConfig.i18n && !isExternalRewrite) {
+      const strippedPathLocale = rewrittenPath.replace(
+        new RegExp(`^/(${NextConfig.i18n.locales.join("|")})`),
```
khuezy, line 234: I forget what `locales` are, is it an array of language codes like `['en', 'es', 'fr']`? This might unintentionally replace urls that contains substrings of those codes?

If so, the regex might want to limit the matching, eg:

```
`^/(${NextConfig.i18n.locales.join("|")})(/|$)`);
  
```

I could be wrong.

conico974: I don't think it's possible here, because we only replace the url if the strippedPath is starting with `/api/`

khuezy (approved):
LGTM! SHIP IT after @nicholas-c confirms the fix.

khuezy:
cc: @sommeeeer
Should we probably add an i18n app for e2e to catch these things.

conico974:
@khuezy page router already has some i18n config.
We just didn't test these specific cases

sommeeeer:
> cc: @sommeeeer Should we probably add an i18n app for e2e to catch these things.

yeah, most definitely. we have discussed having some sort of logic that builds different next apps with different configurations (basePath, i18n etc) and then tests them in a Github Action locally. without deploying the actual apps. since most of it can probably be tested that way, and dont really need a deploy.  perhaps, until we have that up and going, perhaps we should consider adding an app with some `i18n` and locales?

nicholas-c:
@conico974 working perfectly 👌 

API route working:
https://next-15-en.nich-carter-qza3ng.gymshark-sandbox.tools/account/register

i18n config working: 
https://next-15-es.nich-carter-qza3ng.gymshark-sandbox.tools/account/register
https://next-15-es.nich-carter-qza3ng.gymshark-sandbox.tools/fr/account/register

Thanks again!

PR #769, opened 2025-03-12
fix: ignore @img in copyTracedFiles

sommeeeer:
In Next `15.2.2` (and maybe earlier minor and patch) sharp seems have some deps under the scope `@img/*`. (i.e [@img/sharp-linux-x64](https://www.npmjs.com/package/@img/sharp-linux-x64)) 

The issue was discovered in this [comment](https://github.com/sst/sst/issues/5534#issuecomment-2713713849). 

Without this fix the default server function might end up being 32mb+ larger than it has to.

### DIFF
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -198,7 +198,10 @@ File ${fullFilePath} does not exist
       //TODO: we need to figure which packages we could safely remove
       from.includes(path.join("node_modules", "caniuse-lite")) ||
       // from.includes("jest-worker") || This ones seems necessary for next 12
-      from.includes(path.join("node_modules", "sharp"))
+      from.includes(path.join("node_modules", "sharp")) ||
+      // This seems to be only in Next 15
+      // Some of sharp deps are under the @img scope
+      from.includes(path.join("node_modules", "@img"))
     ) {
       return;
     }
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -19,6 +19,23 @@ import { MIDDLEWARE_TRACE_FILE } from "./constant.js";
 
 const __dirname = url.fileURLToPath(new URL(".", import.meta.url));
 
+//TODO: we need to figure which packages we could safely remove
+const EXCLUDED_PACKAGES = [
+  "caniuse-lite",
+  "sharp",
+  // This seems to be only in Next 15
+  // Some of sharp deps are under the @img scope
+  "@img",
+];
+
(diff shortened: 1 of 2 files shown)


vicb (approved):

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -198,7 +198,10 @@ File ${fullFilePath} does not exist
       //TODO: we need to figure which packages we could safely remove
       from.includes(path.join("node_modules", "caniuse-lite")) ||
       // from.includes("jest-worker") || This ones seems necessary for next 12
-      from.includes(path.join("node_modules", "sharp"))
+      from.includes(path.join("node_modules", "sharp")) ||
+      // This seems to be only in Next 15
+      // Some of sharp deps are under the @img scope
+      from.includes(path.join("node_modules", "@img"))
```
conico974, line 204: All of those should probably be replaced by a proper regex. 
This would be true for `node_modules/@imgx/whatever`, same thing for sharp and caniuse-lite

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -19,6 +19,23 @@ import { MIDDLEWARE_TRACE_FILE } from "./constant.js";
...
+  "caniuse-lite",
+  "sharp",
+  // This seems to be only in Next 15
+  // Some of sharp deps are under the @img scope
+  "@img",
+];
+
+function isExcluded(srcPath: string) {
+  return EXCLUDED_PACKAGES.some((excluded) =>
+    new RegExp(`${path.sep}node_modules${path.sep}${excluded}${path.sep}`).test(
```
vicb, line 33: - use `getCrossPlatformPathRegex`
- use `str.match(regexp)` instead of `.test` (which updates/uses `re.lastIndex` - it shouldn't cause issue here but it did for us in the cloudflare adapter so it would be better to avoid the construct)

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -30,9 +31,7 @@ const EXCLUDED_PACKAGES = [
 
 function isExcluded(srcPath: string) {
   return EXCLUDED_PACKAGES.some((excluded) =>
-    new RegExp(`${path.sep}node_modules${path.sep}${excluded}${path.sep}`).test(
-      srcPath,
-    ),
+    srcPath.match(getCrossPlatformPathRegex(`/node_modules/${excluded}/`)),
```
sommeeeer, line 34: would you prefer me to wrap this in `Boolean()`?

vicb: If something, I would prefer `!== null` over `Boolean()` but I don't think it's necessary.

PR #771, opened 2025-03-13
add timeout in revalidatePath test as locally can run too fast

james-elicx:
I was seeing some flakiness when running the e2es locally in Cloudflare purely due to the fact that the test ran in less than 1 second and so the seconds in the date did not change.

image

### DIFF
--- a/packages/tests-e2e/tests/appRouter/revalidateTag.test.ts
+++ b/packages/tests-e2e/tests/appRouter/revalidateTag.test.ts
@@ -79,6 +79,7 @@ test("Revalidate path", async ({ page, request }) => {
   elLayout = page.getByText("Date:");
   const initialDate = await elLayout.textContent();
 
+  // Wait so that enough time passes for the data on the page to update when revalidating
   await page.waitForTimeout(2000);
 
   // Send revalidate path request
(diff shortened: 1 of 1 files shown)


khuezy (approved):

packages/tests-e2e/tests/appRouter/revalidateTag.test.ts
```diff
@@ -79,6 +79,8 @@ test("Revalidate path", async ({ page, request }) => {
   elLayout = page.getByText("Date:");
   const initialDate = await elLayout.textContent();
 
+  await page.waitForTimeout(2000);
```
vicb, line 82: nit: add a comment about why this is needed/what it solves?

vicb (approved):

conico974 (approved):

packages/tests-e2e/tests/appRouter/revalidateTag.test.ts
james-elicx:
```suggestion
  // Wait so that enough time passes for the data on the page to update when revalidating
  await page.waitForTimeout(2000);
```

PR #773, opened 2025-03-13
Feat Add additional params to the queue override

conico974:
Add additional params to the queue
Also add a `.js` to the import in `logger` to fix an issue with test in cloudflare

### DIFF
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -324,10 +324,10 @@ export async function revalidateIfRequired(
 
       // For some weird cases, lastModified is not set, haven't been able to figure out yet why
       // For those cases we add the etag to the deduplication id, it might help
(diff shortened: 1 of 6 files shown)


packages/open-next/src/core/routing/util.ts
```diff
@@ -324,10 +324,10 @@ export async function revalidateIfRequired(
 
       // For some weird cases, lastModified is not set, haven't been able to figure out yet why
       // For those cases we add the etag to the deduplication id, it might help
-      const etag = headers.etag ?? headers.ETag ?? "";
+      const etag = headers.etag ?? `${headers.ETag}` ?? "";
```
vicb, line 327: Nit: wrap the whole expression?

vicb: Nit rename etag to eTag?

PR #776, opened 2025-03-13
fix: add cookies in express-dev wrapper

sommeeeer:
we didn't `set-cookie` in `express-dev` wrapper. should we call: 
```ts
res.flushHeaders();
res.uncork();
```
aswell? like in the `node` wrapper?


conico974 (commented):
We should just add `res.flushHeaders()`

conico974 (approved):
LGTM Thanks

Issue #772, opened 2025-03-13
set-cookie in middleware


sommeeeer:
if you `set-cookie` in middleware Next uses an internal header called `x-middleware-set-cookie`. We need to remove that. https://github.com/opennextjs/opennextjs-aws/blob/656c7096f9a9a4023c9ec5cacbe57fb7361dafee/examples/app-router/middleware.ts#L68-L73 

these lines results in this header on every route in the `app-router` example:

![Image](https://github.com/user-attachments/assets/caa63f17-c290-4760-8eee-a732a2088375)

im not sure if we want that `x-middleware-next` either?

khuezy:
Was this new header recently added? I don't recall seeing it back in 14ish.

sommeeeer:
> Was this new header recently added? I don't recall seeing it back in 14ish.

i think it got added via this [PR](https://github.com/vercel/next.js/pull/65008/files)

can't we just remove them here? 

https://github.com/opennextjs/opennextjs-aws/blob/7e23eee5b93491b938cbf2433d5bed04809a808b/packages/open-next/src/core/routing/middleware.ts#L111-L112

khuezy:
Oh that was about a year ago.

sommeeeer:
https://github.com/vercel/next.js/pull/75869/files

they filter it out here, but i guess since we bundle our own middleware and inject `.next/server/src/middleware.js` into it here:
https://github.com/opennextjs/opennextjs-aws/blob/daa330b22e8b69b1a515e5a5cacfe8a9da418ba7/packages/open-next/src/build/edge/createEdgeBundle.ts#L96-L100

we might need to remove them in `src/core/routing/middleware.ts` like i said above

PR #777, opened 2025-03-17
fix: remove internal next headers in middleware response

sommeeeer:
For #772

Internally Next uses `x-middleware-set-cookie` and `x-middleware-next`. These are then removed on the response by Next code, but since we have a custom middleware we need to remove it our self.


conico974 (approved):
LGTM Thanks

PR #780, opened 2025-03-18
add more e2e for route handlers in app-router

sommeeeer:
This introduces more general E2E for all the different route handlers available. Let me know if I should include anything more in this PR.

It will test:
- All supported HTTP methods (`GET`, `POST`, `PUT`, `PATCH`, `DELETE`, `HEAD`, and `OPTIONS`)
- That formData works in `POST` route handler
- That revalidation (ISR) works in `GET` route handler
- That you can cache a `GET` route with `force-static`
- That you can set cookies in a route handler
- That you should be able to redirect in a route handler
- Dynamic segments in a route handler
- Query parameters

Reference: https://nextjs.org/docs/app/building-your-application/routing/route-handlers

btw, i had to turn off this in `biome.jsonc`: https://biomejs.dev/linter/rules/use-literal-keys as it was not happy with using  `headers["allow"]` or similar.

### DIFF
--- a/packages/tests-e2e/tests/appRouter/methods.test.ts
+++ b/packages/tests-e2e/tests/appRouter/methods.test.ts
@@ -155,7 +155,7 @@ test("should be able to set cookies in route handler", async ({ request }) => {
   const postData = await postRes.json();
   expect(postData.message).toBe("ok");
   const cookies = postRes.headers()["set-cookie"];
-  expect(cookies).toBe("auth_session=SUPER_SECRET_SESSION_ID_1234");
+  expect(cookies).toContain("auth_session=SUPER_SECRET_SESSION_ID_1234");
 });
 
 test("should be able to redirect in route handler", async ({ request }) => {


biome.jsonc
```diff
@@ -46,6 +46,9 @@
         "rules": {
           "suspicious": {
             "noRedeclare": "off"
+          },
+          "complexity": {
+            "useLiteralKeys": "off"
```
conico974, line 51: Can't you just do `headers.allow` everywhere ?
I really don't like the idea of turning off lint rules, especially for test

sommeeeer: I can, but I only turned off those lint rules as an override in `packages/tests-unit/*` and `packages/tests-e2e/*`. personally i dont like literal keys when you get f.ex:
```ts
headers['content-type'];
headers['content-length'];
headers.allow;
```

anyways, i'll just use literal keys if you want.

packages/tests-e2e/tests/appRouter/methods.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    expect(headers.special).toBe("OpenNext is the best :) :] :> :D");
+  });
+});
+
+test("formData should work in POST route handler", async ({ request }) => {
+  const formData = new FormData();
+  formData.append("name", "OpenNext [] () %&#!%$#");
+  formData.append("email", "opennext@opennext.com");
+  const postRes = await request.post("/methods/post/formdata", {
+    form: formData,
```
vicb, line 95: Posting a `FormData` doesn't work for Cloudflare, posting an object is fine (as in the cookies test). Wondering why it works here?

vicb: nvm updating Playwright fixes this

PR #781, opened 2025-03-18
Feat: Generic Code Patching

conico974:
This PR replaces #741.
This PR allow to apply patch to traced files at build time, as well as provide a way to pass esbuild plugins when needed.

2 PR will depend on this one : #783 and #782

### DIFF
--- a/packages/open-next/src/build/createMiddleware.ts
+++ b/packages/open-next/src/build/createMiddleware.ts
@@ -81,6 +81,7 @@ export async function createMiddleware(
       additionalExternals: config.edgeExternals,
       onlyBuildOnce: forceOnlyBuildOnce === true,
       name: "middleware",
+      additionalPlugins: [],
     });
 
     installDependencies(outputPath, config.middleware?.install);
@@ -96,6 +97,7 @@ export async function createMiddleware(
       options,
       onlyBuildOnce: true,
       name: "middleware",
+      additionalPlugins: [],
     });
   }
 }
--- a/packages/open-next/src/build/createMiddleware.ts
+++ b/packages/open-next/src/build/createMiddleware.ts
@@ -81,7 +81,7 @@ export async function createMiddleware(
       additionalExternals: config.edgeExternals,
       onlyBuildOnce: forceOnlyBuildOnce === true,
       name: "middleware",
-      additionalPlugins: [],
+      additionalPlugins: () => [],
     });
 
     installDependencies(outputPath, config.middleware?.install);
@@ -97,7 +97,7 @@ export async function createMiddleware(
       options,
       onlyBuildOnce: true,
       name: "middleware",
-      additionalPlugins: [],
+      additionalPlugins: () => [],
     });
   }
 }
--- a/packages/open-next/src/build/createMiddleware.ts
+++ b/packages/open-next/src/build/createMiddleware.ts
@@ -81,7 +81,6 @@ export async function createMiddleware(
       additionalExternals: config.edgeExternals,
       onlyBuildOnce: forceOnlyBuildOnce === true,
       name: "middleware",
-      additionalPlugins: () => [],
     });
 
     installDependencies(outputPath, config.middleware?.install);
@@ -97,7 +96,6 @@ export async function createMiddleware(
       options,
       onlyBuildOnce: true,
       name: "middleware",
-      additionalPlugins: () => [],
     });
   }
 }
--- a/packages/open-next/src/build/createServerBundle.ts
+++ b/packages/open-next/src/build/createServerBundle.ts
@@ -6,6 +6,7 @@ import type { FunctionOptions, SplittedFunctionOptions } from "types/open-next";
 import type { Plugin } from "esbuild";
 import logger from "../logger.js";
 import { minifyAll } from "../minimize-js.js";
+import { ContentUpdater } from "../plugins/content-updater.js";
 import { openNextReplacementPlugin } from "../plugins/replacement.js";
 import { openNextResolvePlugin } from "../plugins/resolve.js";
 import { getCrossPlatformPathRegex } from "../utils/regex.js";
@@ -22,7 +23,7 @@ interface CodeCustomization {
   additionalCodePatches: CodePatcher[];
   // These plugins are meant to apply during the esbuild bundling process.
   // This will only apply to OpenNext code.
-  additionalPlugins: Plugin[];
+  additionalPlugins: (contentUpdater: ContentUpdater) => Plugin[];
 }
 
 export async function createServerBundle(
@@ -207,7 +208,9 @@ async function generateBundle(
 
   const disableRouting = isBefore13413 || config.middleware?.external;
 
-  const additionalPlugins = codeCustomization?.additionalPlugins ?? [];
+  const updater = new ContentUpdater(options);
+
+  const additionalPlugins = codeCustomization?.additionalPlugins(updater) ?? [];
 
   const plugins = [
     openNextReplacementPlugin({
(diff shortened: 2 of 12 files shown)


packages/open-next/src/build/createServerBundle.ts
```diff
@@ -152,15 +167,30 @@ async function generateBundle(
   // Copy env files
   buildHelper.copyEnvFile(appBuildOutputPath, packagePath, outputPath);
 
-  // Copy all necessary traced files
-  await copyTracedFiles({
+  // Copy all necessary traced files{
```
vicb, line 170:
```suggestion
  // Copy all necessary traced files
```

packages/open-next/src/build/createServerBundle.ts
```diff
@@ -152,15 +167,30 @@ async function generateBundle(
...
+  await applyCodePatches(options, Array.from(tracedFiles), manifests, [
+    // TODO: create real code patchers here
+    {
+      name: "fakePatchChunks",
+      pathFilter: /chunks\/\d+\.js/,
+      patchCode: async ({ code, manifests }) => {
+        console.log(manifests);
+        return `console.log("patched chunk");\n${code}`;
+      },
+    },
```
vicb, line 190: Remove?

packages/open-next/src/build/patch/astCodePatcher.ts
```diff
@@ -0,0 +1,114 @@
...
+  const node = parse(lang, code).root();
+  const { edits } = applyRule(rule, node, { once });
+  return node.commitEdits(edits);
+}
+
+export function createPatchCode(
+  rule: string | RuleConfig,
+  lang = Lang.TypeScript,
+): PatchCodeFn {
+  return async ({ code }) => patchCode(code, rule, { lang });
```
vicb, line 113: Doesn't need to be async, can we pass the code as a string? (seems like this is not used in this PR).
```suggestion
  return (code) => patchCode(code, rule, { lang });
```

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,124 @@
+import * as fs from "node:fs/promises";
+import logger from "../../logger.js";
+import * as buildHelper from "../helper.js";
+
+// Either before or after should be provided, otherwise just use the field directly
+export interface VersionedField<T> {
+  // The version before which the field should be used
```
vicb, line 7: Would be nice to comment about if `before` & `after` are inclusive?

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,124 @@
...
+  after?: `${number}` | `${number}.${number}` | `${number}.${number}.${number}`;
+  field: T;
+}
+
+export type PatchCodeFn = (args: {
+  // The content of the file that needs to be patched
+  code: string;
+  // The final path of the file that needs to be patched
+  filePath: string;
+  // All js files that will be included in this specific server function
```
vicb, line 22: Add comment on the type about usage, i.e. why are we referring to a server function here?

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+export async function applyCodePatches(
+  buildOptions: buildHelper.BuildOptions,
+  tracedFiles: string[],
+  manifests: Record<string, any>,
+  codePatcher: CodePatcher[],
+) {
+  const nextVersion = buildOptions.nextVersion;
+  console.time("Applying code patches");
```
vicb, line 70: Is it expected to use the console over the logger?

packages/open-next/src/build/patch/astCodePatcher.ts
conico974: It is an object to accomodate for future use (and so that it can be used as is). I've took a look at the patch/plugins applied in cloudflare and we can move almost all of them to use this.
With the benefits of not having to load things like the manifest twice.

packages/open-next/src/build/patch/codePatcher.ts
conico974: We don't have a `logger.time`, but i can add it

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -0,0 +1,124 @@
...
+  tracedFiles: string[];
+  // All next.js manifest that are present at runtime - Key relative to `.next` folder
+  manifests: Record<string, any>;
+}) => Promise<string>;
+
+export interface CodePatcher {
+  name: string;
+  pathFilter: RegExp | VersionedField<RegExp>[];
+  contentFilter?: RegExp | VersionedField<RegExp>[];
+  patchCode: PatchCodeFn | VersionedField<PatchCodeFn>[];
```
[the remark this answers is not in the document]
conico974: Ok let's go for the simplified version then.
It will probably means that we'll have some duplicates in code but that's probably fine. I will have to make a few other change for this to work (as well as to update the other 2 linked PR)

conico974:
@vicb PTAL, i've updated it to use a simplified version for `CodePatcher`. Let me know what you think.
I've also updated the 2 other related PR

packages/open-next/src/build/createMiddleware.ts
```diff
@@ -81,6 +81,7 @@ export async function createMiddleware(
       additionalExternals: config.edgeExternals,
       onlyBuildOnce: forceOnlyBuildOnce === true,
       name: "middleware",
+      additionalPlugins: () => [],
```
vicb, line 84: could this be optional?

packages/open-next/src/build/createServerBundle.ts
```diff
@@ -153,14 +168,20 @@ async function generateBundle(
...
     buildOutputPath: appBuildOutputPath,
     packagePath,
     outputDir: outputPath,
     routes: fnOptions.routes ?? ["app/page.tsx"],
     bundledNextServer: isBundled,
   });
 
+  const additionalCodePatches = codeCustomization?.additionalCodePatches ?? [];
+
+  await applyCodePatches(options, Array.from(tracedFiles), manifests, [
```
vicb, line 181: should tracedFiles be an Array already?

packages/open-next/src/build/createMiddleware.ts
conico974: Yeah it could, i'll change that

packages/open-next/src/build/createServerBundle.ts
conico974: We could convert it directly in `createTracedFiles`

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -51,27 +47,76 @@ export interface CodePatcher {
...
 
+export function parseVersions(versions?: Versions): {
+  before?: string;
+  after?: string;
+} {
+  if (!versions) {
+    return {};
+  }
+  // We need to use regex to extract the versions
+  const versionRegex = /([<>]=?)([0-9]+\.[0-9]+\.[0-9]+)/g;
```
vicb, line 58: - `=` is not optional
- using `\d` for digits

```suggestion
  const versionRegex = /([<>]=)(\d+\.\d+\.\d+)/g;
```

packages/open-next/src/build/patch/codePatcher.ts
```diff
@@ -51,27 +47,76 @@ export interface CodePatcher {
...
+  if (matches.length > 2) {
+    throw new Error("Invalid version range, too many matches found");
+  }
+  let after: string | undefined;
+  let before: string | undefined;
+  for (const match of matches) {
+    const [_, operator, version] = match;
+    if (operator === "<=") {
+      before = version;
+    } else if (operator === ">=") {
```
vicb, line 72:
```suggestion
    } else {
```

vicb (approved):
Thanks, versions are much more readable like this!

PR #782, opened 2025-03-18
Fix: fetch cache set missing wait until

conico974:
Port of this patch to aws https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/build/patches/plugins/fetch-cache-wait-until.ts

This PR depends on #781. **Should not be merged before it**


packages/open-next/src/build/createServerBundle.ts
```diff
@@ -180,14 +181,7 @@ async function generateBundle(
 
   await applyCodePatches(options, Array.from(tracedFiles), manifests, [
     // TODO: create real code patchers here
```
vicb, line 183:
```suggestion
```

vicb (approved):

.changeset/fuzzy-eyes-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix dangling set on fetch cache
```
vicb, line 5: should "set" be "promise"?

vicb (approved):

.changeset/fuzzy-eyes-pump.md
conico974: Yep it should have been something like `fix dangling promise on set for the fetch cache`

vicb (approved):

PR #783, opened 2025-03-18
Fix fetch cache for isr

conico974:
Patch the fetch cache and `unstable_cache` so that they does not get revalidated (if not necessary) by our ISR revalidation request.

It replaces `patchAsyncStorage`.

Can be removed once this gets merged : https://github.com/vercel/next.js/pull/72082

This PR depends on #781. Should not be merged before it

### DIFF
--- a/packages/open-next/src/build/patch/patchFetchCacheISR.ts
+++ b/packages/open-next/src/build/patch/patchFetchCacheISR.ts
@@ -1,6 +1,7 @@
 import { getCrossPlatformPathRegex } from "utils/regex.js";
 import { createPatchCode } from "./astCodePatcher.js";
 import type { CodePatcher } from "./codePatcher";
+import { Lang } from "@ast-grep/napi";
 
 export const fetchRule = `
 rule:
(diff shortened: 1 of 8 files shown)


packages/tests-unit/tests/build/patch/patchFetchCacheISR.test.ts
```diff
@@ -0,0 +1,126 @@
...
+                        revalidate: options.revalidate,
+                        tags,
+                        fetchIdx,
+                        fetchUrl,
+                        softTags: implicitTags
+                    });
+}
+`;
+
+const patchFetchCacheCodeunMinified = `
```
vicb, line 42:
```suggestion
const patchFetchCacheCodeUnMinified = `
```

packages/open-next/src/build/patch/patchFetchCacheISR.ts
```diff
@@ -0,0 +1,113 @@
...
+        stopBy: end
+fix:
+  ($STORE_OR_CACHE.isOnDemandRevalidate && !globalThis.__openNextAls?.getStore()?.isISRRevalidation)
+`;
+
+export const patchFetchCacheForISR: CodePatcher = {
+  name: "patch-fetch-cache-for-isr",
+  patches: [
+    {
+      after: "14.0.0",
```
vicb, line 85: actually I'm not sure if I like before/after. I know one is inclusive and not the other but reviewing in GH is kind of hard `versions: ">=14.0.0 <15.2" would be much nicer IMO

vicb (approved):

packages/open-next/src/build/patch/patchFetchCacheISR.ts
conico974: I already thought about something like that, depending on how it's done it could actually introduce some errors.
I'll need to think about that, maybe with pretty strict template literals type we could enforce something enough that it won't cause trouble.
I'll test something tomorrow

vicb (approved):

vicb:
> It replaces `patchAsyncStorage`.

Should `patchAsyncStorage` be removed in this PR then?

conico974:
> Should patchAsyncStorage be removed in this PR then?

The unstable_cache is for next 14.2+, ideally i'd keep `patchAsyncStorage` for earlier versions.
I can remove it for Next 14.2+ with the esbuild plugin ? I guess that's fine for you guys

PR #785, opened 2025-03-19
fix e2e methods static get route

sommeeeer:
For some reason my own deployed `app-router` had a comma in the `cache-control` value, but the `app-router` that is deployed from here does not.


conico974 (approved):
Thanks for the quick fix

PR #786, opened 2025-03-19
export `findMonorepoRoot` function

james-elicx:
i would like to re-use the function in cloudflare to get the package manager

### DIFF
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -65,7 +65,7 @@ export function normalizeOptions(
   };
 }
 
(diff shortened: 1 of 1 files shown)


vicb:
> i would like to re-use the function in cloudflare to get the package manager

Can you get it from the build options returned by `normalizeOptions`?

james-elicx:
> > i would like to re-use the function in cloudflare to get the package manager
> 
> Can you get it from the build options returned by `normalizeOptions`?

yeah i changed what i was doing this morning and no longer need this

PR #787, opened 2025-03-20
refactor `compareSemver`

vicb:


conico974 (approved):
LGTM Thanks.
I guess we could update the generic patch as well to support different operator if we need to later (not that important for now i guess)

vicb:
> I guess we could update the generic patch as well to support different operator if we need to later (not that important for now i guess)

I wanted to do this change for a while but yeah updating the generic patch is what trigger that PR. It would be nice to be able to express i.e. "<15" for patch only required for 14.

Thanks for the review!

PR #791, opened 2025-03-20
fix: external rewrite for streaming

### LINKED ISSUE #789: Streaming breaks external rewrite
im currently working on some more e2e for streaming. during working on that i found out that enabling streaming in `pages-router` will break this: https://github.com/opennextjs/opennextjs-aws/blob/652550fef4a87de569dc85cb9f98fe4a689fea6f/packages/tests-e2e/tests/pagesRouter/rewrite.test.ts#L20-L25

it turns out that the headers is incorrect. the body is actually correct, but the headers from the response is not what it should be. same thing happens if i add the test to `app-router` aswell (which already has streaming enabled)

[here](https://github.com/sommeeeer/opennext-streaming-rewrite) is a reproduction of the problem. it seems to work with local build, but not once deployed to AWS. steps to reproduce:
```bash
# this will deploy to AWS with SST
pnpm install
pnpm sst:deploy

# i added a local opennext build that seems to work fine
pnpm opennextlocalbuild
# then visit http://localhost:3000/external-on-image
# however it seems that http://localhost:3000/opennextjs is broken locally. i guess its also related to headers
```
then go visit the `/external-on-image`, `/opennextjs` and `/rewrite`. they are all external rewrites. there is also a massive amount of headers on the response, so something with the headers is broken.

sommeeeer:
Closes #789

Turns out we were not passing the correct headers to the response when streaming is used.

### DIFF
--- a/.changeset/heavy-moles-share.md
+++ b/.changeset/heavy-moles-share.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix external rewrite for streaming
--- a/packages/open-next/src/core/requestHandler.ts
+++ b/packages/open-next/src/core/requestHandler.ts
@@ -138,7 +138,7 @@ export async function openNextHandler(
               origin: false,
               initialURL: internalEvent.url,
             },
-            headers,
+            routingResult.headers,
             options.streamCreator,
           );
           response.statusCode = routingResult.statusCode;


khuezy:
Nice catch! Ooof what are the repercussions of returning the wrong headers.

conico974 (approved):
LGTM Good catch

conico974:
> Nice catch! Ooof what are the repercussions of returning the wrong headers. I didn't see anyone specifically reporting weird behaviors.

It's only an issue for external rewrite if you use the internal middleware and you use streaming. Probably why not everyone was affected.
It's also only since we introduced the proxy override, earlier versions didn't have the issue

PR #794, opened 2025-03-21
add e2e for dynamicparams in app-router

sommeeeer:
E2E for `dynamicParams`  in `app-router`:

- dynamicParams set as `true` in `/isr/dynamic-params-true/[id]`
- dynamicParams set as `false` in `/isr/dynamic-params-false/[id]`

### DIFF
--- a/packages/tests-e2e/tests/appRouter/isr.test.ts
+++ b/packages/tests-e2e/tests/appRouter/isr.test.ts
@@ -114,8 +114,8 @@ test.describe("dynamicParams set to true", () => {
     expect(content).toEqual("This is post 11");
   });
 
-  test("should 404 for a path that is not found", async ({ page }) => {
-    const res = await page.goto("/isr/dynamic-params-false/11");
+  test("should 404 when you call notFound", async ({ page }) => {
+    const res = await page.goto("/isr/dynamic-params-true/20");
     expect(res?.status()).toEqual(404);
     expect(res?.headers()["cache-control"]).toBe(
       "private, no-cache, no-store, max-age=0, must-revalidate",
--- a/packages/tests-e2e/tests/appRouter/isr.test.ts
+++ b/packages/tests-e2e/tests/appRouter/isr.test.ts
@@ -94,46 +94,61 @@ test("Incremental Static Regeneration with data cache", async ({ page }) => {
   expect(originalFetchedDate).toEqual(finalFetchedDate);
 });
 
-test("dynamicParams set to true", async ({ page }) => {
-  const res = await page.goto("/isr/dynamic-params-true/1");
-  expect(res?.status()).toEqual(200);
-  expect(res?.headers()["x-nextjs-cache"]).toEqual("HIT");
-  const title = await page.getByTestId("title").textContent();
-  const content = await page.getByTestId("content").textContent();
-  expect(title).toEqual("Post 1");
-  expect(content).toEqual("This is post 1");
+test.describe("dynamicParams set to true", () => {
+  test("should be HIT on a path that was prebuilt", async ({ page }) => {
+    const res = await page.goto("/isr/dynamic-params-true/1");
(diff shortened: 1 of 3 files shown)


packages/tests-e2e/tests/appRouter/isr.test.ts
```diff
@@ -93,3 +93,47 @@ test("Incremental Static Regeneration with data cache", async ({ page }) => {
...
+  const res2 = await page.goto("/isr/dynamic-params-true/11");
+  expect(res2?.headers()["x-nextjs-cache"]).toEqual("MISS");
+  const title2 = await page.getByTestId("title").textContent();
+  const content2 = await page.getByTestId("content").textContent();
+  expect(title2).toEqual("Post 11");
+  expect(content2).toEqual("This is post 11");
+
+  // should 500 for a non-existing path
+  const res3 = await page.goto("/isr/dynamic-params-true/21");
+  expect(res3?.status()).toEqual(500);
```
conico974, line 116: Do we really want to test the 500 for this case ?
We should use `notFound()` when the id is expected but not correct (we could add a 500 case as well to simulate a broken case)

sommeeeer: i wanted to have a 500 case, and also check that the headers is what we want.

packages/tests-e2e/tests/appRouter/isr.test.ts
```diff
@@ -93,3 +93,62 @@ test("Incremental Static Regeneration with data cache", async ({ page }) => {
...
+    expect(res?.headers()["x-nextjs-cache"]).toEqual("HIT");
+    const title = await page.getByTestId("title").textContent();
+    const content = await page.getByTestId("content").textContent();
+    expect(title).toEqual("Post 1");
+    expect(content).toEqual("This is post 1");
+  });
+
+  test("should SSR on a path that was not prebuilt", async ({ page }) => {
+    const res = await page.goto("/isr/dynamic-params-true/11");
+    expect(res?.headers()["x-nextjs-cache"]).toEqual("MISS");
```
conico974, line 110: There is something that i don't like about this one. If for whatever reasons the test fails, any subsequent tries will fail as well. It's likely fine, but we should at least add a comment that this test cannot be retried.

packages/tests-e2e/tests/appRouter/isr.test.ts
```diff
@@ -93,3 +93,62 @@ test("Incremental Static Regeneration with data cache", async ({ page }) => {
...
+    const title = await page.getByTestId("title").textContent();
+    const content = await page.getByTestId("content").textContent();
+    expect(title).toEqual("Post 11");
+    expect(content).toEqual("This is post 11");
+  });
+
+  test("should 404 for a path that is not found", async ({ page }) => {
+    const res = await page.goto("/isr/dynamic-params-false/11");
```
conico974, line 118: This test should be under the other describe

packages/tests-e2e/tests/appRouter/isr.test.ts
sommeeeer: this is not actually true. in `next start` it is, but once deployed `x-nextjs-cache` never changes, because CF will cache it. i'll add a comment about it tho

packages/tests-e2e/tests/appRouter/isr.test.ts
sommeeeer: > This test should be under the other describe

i already had a 404 case there, but i wanted to test it both places to be consistent. will fix correct path

packages/tests-e2e/tests/appRouter/isr.test.ts
conico974: This is not a guarantee at all, as i said it's probably fine but still

conico974 (approved):
LGTM Thanks

PR #796, opened 2025-03-24
Add a `compileEdge` parameter to `compileOpenNextConfig`

vicb:
The cloudflare adapter will move away from using an internal middleware.

With this change we can still force the compilation of an edge configuration when the middleware is part of the server bundle.

(Longer term we might want to find a better solution to detect this)

### DIFF
--- a/.changeset/gorgeous-rocks-appear.md
+++ b/.changeset/gorgeous-rocks-appear.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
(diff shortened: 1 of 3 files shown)


conico974 (approved):
LGTM,
I think in the long term, we could look at the wrappers used to determine what we need to do here.

vicb:
> I think in the long term, we could look at the wrappers used to determine what we need to do here.

+1, the wrapper should have a way to indicate what they need.

Thanks for the review!

PR #799, opened 2025-03-26
fix: merge request headers in middleware in app-router example

sommeeeer:
When doing `pnpm build && pnpm start` in `app-router`. You will get an error when trying to run the e2e test locally for `/methods/post/formdata`:

`⨯ TypeError: Content-Type was not one of "multipart/form-data" or "application/x-www-form-urlencoded".`

The reason is this line: https://github.com/opennextjs/opennextjs-aws/blob/fbe09aed4280a5c0820efda74230a12c7cd5bf2c/examples/app-router/middleware.ts#L31

I guess we need to pass in the request headers already coming in. Reason it works once deployed is because OpenNext doesn't remove them like `next start`.

I also removed an unused variable in `/src/core/routing/middleware.ts`

### DIFF
--- a/examples/app-router/middleware.ts
+++ b/examples/app-router/middleware.ts
@@ -28,7 +28,7 @@ export function middleware(request: NextRequest) {
     const u = new URL("https://opennext.js.org/share.png");
     return NextResponse.rewrite(u);
   }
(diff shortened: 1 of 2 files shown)


conico974 (approved):
Thanks

PR #800, opened 2025-03-27
Improve Perf

conico974:
All these changes only apply to Next 15+

Adds patches to the Next.js codebase to remove unused imports at runtime, improving cold start times. Early tests indicate a performance gain of 120-200 ms without any additional code changes.

Disables the default preloading of routes at startup (which was done in the background). Now by default, only the necessary routes are lazily loaded as needed.
Introduces a new configuration setting to control when routes are preloaded. The available options are:
- `none` : No preloading, default option
-  `withWaitUntil`: Preload using platform provided waitUntil - Will only occur on the first request
- `onWarmerEvent` : Preload on warmer events - Need to be done by the wrapper itself, only `aws-lambda-streaming` does it for now.
- `onStart`: Preload at startup - Be aware this is a blocking operation


martinezdylan:
Would love to see if we could get this one in - hoping to use `onWarmerEvent` to keep routes preloaded on the warm instances. When we're using available warm option on NextJS (ref. https://sst.dev/docs/component/aws/nextjs/#warm) the instance is technically up but routes aren't loaded - this causes us to still experience cold-ish starts despite the warmer.

cc. @nguyendon for vis

conico974:
@martinezdylan `onWarmerEvent` won't work for sst without some change there because sst intercept the warmer event and does it's own thing.
It should work if you use `onStart`

martinezdylan:
I guess I'm confused, the PR description mentions this should technically be happening now by default? Wouldn't using onStart make no difference?

What we're experiencing suggests even with warm set to 20, we still are having to load / compile the routes when we first hit the endpoint.

conico974:
Because as said in the PR description, route preloading at the moment are done in the background, but warmer event return immediately, so the route are not preloaded, only NextServer itself.

The `onStart` from this PR is blocking, which means that it will not return before having done the preloading, but this would be for everything, not just the warmer event, real cold start would likely be slower (How much will depend on your app and deps)

Ideally you'd only use `onWarmerEvent` but this means that some change has to be done in sst itself

.changeset/warm-pans-argue.md
```diff
@@ -0,0 +1,8 @@
+---
+"@opennextjs/aws": patch
+---
+
+Some perf improvements : 
+- Eliminate unnecessary runtime imports.
```
vicb, line 6: add details about the imports that were removed?

packages/open-next/src/build/patch/patches/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./patchEnvVar.js";
```
vicb, line 1: nit: I think `export *` in a barrel is kind of an anti-pattern, could we change to named exports?

packages/open-next/src/build/patch/patches/patchEnvVar.ts
```diff
@@ -0,0 +1,46 @@
...
+  inside:
+    kind: if_statement
+    stopBy: end
+fix:
+  '${value}'
+`;
+
+export const patchEnvVars: CodePatcher = {
+  name: "patch-env-vars",
+  patches: [
```
vicb, line 17: maybe add a one line comment on the purpose of each patch?

packages/open-next/src/build/patch/patches/patchEnvVar.ts
```diff
@@ -0,0 +1,46 @@
+import { createPatchCode } from "../astCodePatcher.js";
+import type { CodePatcher } from "../codePatcher";
+
+export const envVarRuleCreator = (envVar: string, value: string) => `
+rule:
+  kind: member_expression
+  pattern: process.env.${envVar}
+  inside:
+    kind: if_statement
```
vicb, line 9: `if_statement` has both the condition and the block statement, I don't think that is what we are looking for here?

Maybe also add a comment on the functionμ?

packages/open-next/src/core/createMainHandler.ts
```diff
@@ -26,6 +26,8 @@ export async function createMainHandler() {
   globalThis.serverId = generateUniqueId();
   globalThis.openNextConfig = config;
 
+  await globalThis.__next_route_preloader("start");
```
vicb, line 29: Would be nice to add a comment here too

packages/open-next/src/core/util.ts
```diff
@@ -59,6 +59,51 @@ const nextServer = new NextServer.default({
...
+  if (stage === "waitUntil" && routePreloadingBehavior === "withWaitUntil") {
+    // We need to access the waitUntil
+    const waitUntil = globalThis.__openNextAls.getStore()?.waitUntil;
+    if (!waitUntil) {
+      error(
+        "You've tried to use the 'withWaitUntil' route preloading behavior, but the 'waitUntil' function is not available.",
+      );
+    }
+    debug("Preloading entries with waitUntil");
+    waitUntil?.(nextServer.unstable_preloadEntries());
```
vicb, line 93: Missing return in the `if` above? (or missing else clause)

packages/open-next/src/types/open-next.ts
```diff
@@ -313,6 +318,16 @@ export interface FunctionOptions extends DefaultFunctionOptions {
    * @deprecated This is not supported in 14.2+
    */
   experimentalBundledNextServer?: boolean;
+
+  /**
+   * The route preloading behavior. Only supported in Next 15+.
+   * - "none" - No preloading of the route at all
```
vicb, line 324: Move the comments on the type l121?

IIUC "none" is the default behavior of next (/OpenNext):
- maybe this should be explicitly documented
- is "none" accurate if it happens in the background ?

vicb (commented):
Great PR!

My comments resolve around adding more documentation.

packages/open-next/src/types/open-next.ts
```diff
@@ -118,6 +118,20 @@ export interface ResolvedRoute {
   type: RouteType;
 }
 
+/**
+ * The route preloading behavior. Only supported in Next 15+.
```
vicb, line 122: Suggest:

- Add the default behavior of Next is disabled
- Add some guidelines to pick a value

vicb (approved):
a couple nits but LGTM

packages/open-next/src/types/open-next.ts
conico974: There is no general guidelines really, it depends so much on the deployed app that everyone should do their own testing.  I'll add a comment about that

PR #802, opened 2025-03-27
fix(middleware): Support wasm in bundled middleware

vicb:
Related to https://github.com/opennextjs/opennextjs-cloudflare/issues/471

2 fixes in this PR:

- For bundled middleware, we do copy the source file (in `createServerBundle` but we also need to copy .wasm files
- The second fix is specific for the edge, the file name is used in Next middleware source, then if we re-use the same name to import the same file, ESBulid will append a 2 and Next can not reference this import.

### DIFF
--- a/packages/open-next/src/build/createServerBundle.ts
+++ b/packages/open-next/src/build/createServerBundle.ts
@@ -3,6 +3,7 @@ import path from "node:path";
 
 import type { FunctionOptions, SplittedFunctionOptions } from "types/open-next";
 
+import { loadMiddlewareManifest } from "config/util.js";
 import type { Plugin } from "esbuild";
 import logger from "../logger.js";
 import { minifyAll } from "../minimize-js.js";
@@ -135,12 +136,14 @@ async function generateBundle(
   //       `.next/standalone/package/path` (ie. `.next`, `server.js`).
   //       We need to output the handler file inside the package path.
   const packagePath = buildHelper.getPackagePath(options);
-  fs.mkdirSync(path.join(outputPath, packagePath), { recursive: true });
+  const outPackagePath = path.join(outputPath, packagePath);
+
+  fs.mkdirSync(outPackagePath, { recursive: true });
 
   const ext = fnOptions.runtime === "deno" ? "mjs" : "cjs";
   fs.copyFileSync(
     path.join(options.buildDir, `cache.${ext}`),
-    path.join(outputPath, packagePath, "cache.cjs"),
+    path.join(outPackagePath, "cache.cjs"),
   );
 
   if (fnOptions.runtime === "deno") {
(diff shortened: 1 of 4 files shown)


packages/open-next/src/build/createServerBundle.ts
```diff
@@ -159,15 +162,26 @@ async function generateBundle(
   if (!config.middleware?.external) {
     fs.copyFileSync(
       path.join(options.buildDir, "middleware.mjs"),
-      path.join(outputPath, packagePath, "middleware.mjs"),
+      path.join(outPackagePath, "middleware.mjs"),
+    );
+
+    const middlewareManifest = loadMiddlewareManifest(
```
conico974, line 168: We should create a generic function for this and use it both here and there https://github.com/opennextjs/opennextjs-aws/blob/fbe09aed4280a5c0820efda74230a12c7cd5bf2c/packages/open-next/src/build/edge/createEdgeBundle.ts#L212-L230
We also need to copy the assets

packages/open-next/src/plugins/edge.ts
```diff
@@ -139,15 +139,18 @@ if (!globalThis.URLPattern) {
 `
 }
 ${wasmFiles
-  .map((file) =>
+  .map((file, i) =>
     isInCloudfare
-      ? `import ${file.name} from './wasm/${file.name}.wasm';`
+      ? // Decorate the name to avoid name collisions
+        `import __OpenNextWasm${i} from './wasm/${file.name}.wasm';
+globalThis.${file.name} = __OpenNextWasm${i}`
```
conico974, line 146: I'm probably missing something here, but if there is collision, wouldn't this also cause some issue with the global ?

Only things that use this (without being a separate function) should be the middleware

packages/open-next/src/build/createServerBundle.ts
```diff
@@ -159,15 +162,26 @@ async function generateBundle(
   if (!config.middleware?.external) {
     fs.copyFileSync(
       path.join(options.buildDir, "middleware.mjs"),
-      path.join(outputPath, packagePath, "middleware.mjs"),
+      path.join(outPackagePath, "middleware.mjs"),
```
conico974, line 165: All this copy logic should also be in a function on this file so that everything middleware related is there https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/build/createMiddleware.ts.

Pretty sure we need to copy the wasm/asset files for the external middleware as well

packages/open-next/src/plugins/edge.ts
vicb: `.next/server/src/middleware.js` would have something like

```ts
683: (e) => {
  e.exports = wasm_ea010aa52414c7de383495546f15cb7d49b055fc;
}
```

If the code here is

```ts
  import wasm_ea010aa52414c7de383495546f15cb7d49b055fc from '...';
```

Then ESBuild will change that to `wasm_ea010aa52414c7de383495546f15cb7d49b055fc2` (appending a 2 at the end).

So we want the "collision" here.

conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review Nico!

PR #807, opened 2025-04-01
add new feature in config for copying files to a specified server function after the build is done

sommeeeer:
This PR introduces a new feature for `open-next.config.ts`. You can now specify additional files to be copied after the server function(s) has been bundled. This feature also supports split functions.

The `srcPath` can be a relative or absolute path and can accept glob patterns. If a glob matches multiple files, `dstPath` will be treated as a directory. If only a single file is matched, `dstPath` will be treated as a file.

Here is an example:
```ts
const config = {
  default: {
    copyFiles: [
      {
        // srcPath can be relative or absolute
        // can also be a glob pattern
        srcPath: '**/*.js',
        // dstPath is relative to the functions output
        // will become a directory if multiple files is matched in srcPath
        dstPath: '/myfolder/here',
      },
    ],
  },
  functions: {
    testfunction: {
      routes: ['app/api/test/route'], 
      patterns: ['api/test/route'],
      override: {
        wrapper: 'aws-lambda-streaming',
      },
      copyFiles: [
        {
          srcPath: '/home/user/Downloads/important.wasm',
          dstPath: 'wasm/my.wasm',
        },
      ],
    },
  },
} satisfies OpenNextConfig;
```

### DIFF
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -5,6 +5,7 @@ import url from "node:url";
 
 import type { BuildOptions as ESBuildOptions } from "esbuild";
 import { build as buildAsync, buildSync } from "esbuild";
+import { globSync } from "glob";
 import type {
   CopyFile,
   DefaultOverrideOptions,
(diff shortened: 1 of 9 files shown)


packages/open-next/src/build/helper.ts
```diff
@@ -440,3 +442,57 @@ export async function isEdgeRuntime(
...
+/**
+ * Copy files that are specified in the `copyFiles` property into the server functions output directory.
+ *
+ * @param copyFiles - Array of files to copy. Each file should have a `srcPath` and `dstPath` property.
+ * @param outputPath - Path to the output directory.
+ */
+export function copyCustomFiles(copyFiles: CopyFile[], outputPath: string) {
+  copyFiles.forEach(({ srcPath, dstPath }) => {
+    // Find all files matching the pattern
+    const matchedFiles = globSync(srcPath, {
```
vicb, line 455: would that work with "relative" files?
and relative to what?

packages/tests-unit/tests/build/helper.test.ts
```diff
@@ -65,3 +74,175 @@ describe("compareSemver", () => {
...
+          "another-important.js": "console.log('important!')",
+        },
+      },
+      "this/is/a/real/dir": {
+        "randomfile.txt": "some content",
+        "another-dirfdsf": {
+          "another-filedsfdsf.txt": "another content",
+        },
+        "empty-file.txtfdsf": "",
+        "imporant-files": {
```
vicb, line 97:
```suggestion
        "important-files": {
```

packages/open-next/src/build/helper.ts
```diff
@@ -440,3 +442,57 @@ export async function isEdgeRuntime(
...
+      nodir: true,
+      windowsPathsNoEscape: true,
+    });
+
+    if (matchedFiles.length === 0) {
+      logger.warn(`No files found for pattern: ${srcPath}`);
+      return;
+    }
+
+    if (matchedFiles.length === 1) {
```
vicb, line 465: If I want to copy `*.js` to `/some/path`, will be behavior be different if there is one vs multiple files? (if yes, that does not sound right)

packages/open-next/src/build/helper.ts
sommeeeer: > would that work with "relative" files?
> and relative to what?

yeah, and by default its relative to `process.cwd()`. perhaps i should change that in-case you are in a monorepo.

packages/open-next/src/build/helper.ts
sommeeeer: > If I want to copy `*.js` to `/some/path`, will be behavior be different if there is one vs multiple files? (if yes, that does not sound right)

yeah your correct about that one. after some more thought i should probably think about a better way to handle this. im open for suggestions.

vicb: One way to handle that could be to have exclusive `dstFile` (error on multiple match) and `dstDir` options

vicb: Also I am not sure what use case you are solving with this PR. Maybe having a hook accepting a function taking the buildOptions as a parameter could be good enough - that would mean the options are part of the API and must be documented and ~stable.

sommeeeer: > One way to handle that could be to have exclusive `dstFile` (error on multiple match) and `dstDir` options

thats a good idea. probably something I would go for if we choose to proceed with this PR. 


> Also I am not sure what use case you are solving with this PR

the idea was that sometimes you need or want a file that's not there. we've had a few cases where people were missing a file in the server function's output. i.e this discord [thread](https://discord.com/channels/1283128968140161065/1286094576788177059/1354125386241282179) 

however, if you guys think this wont be needed i can close it. Next does have `outputFileTracingIncludes` which probably might be enough for most cases anyways.


> Maybe having a hook accepting a function taking the buildOptions as a parameter could be good enough - that would mean the options are part of the API and must be documented and ~stable.

that is an excellent idea. im gonna think about this until tomorrow. with more thoroughly thought this PR might not be needed for any use-cases.

sommeeeer:
thanks for the feedback on this @vicb. after a night sleep I decided to close this, and I will instead write about `outputFileTracingIncludes` in our docs. will do some more testing on it with OpenNext beforehand.

PR #810, opened 2025-04-02
Fix middleware bypass

conico974:
A security vulnerability has been found similar to the recent CVE-2025-29927.
This PR fixes it

PR #811, opened 2025-04-02
[V2] backport vulnerability fix

conico974:
This is a backport of #810 to OpenNext V2

### DIFF
--- a/packages/open-next/src/adapters/routing/middleware.ts
+++ b/packages/open-next/src/adapters/routing/middleware.ts
@@ -1,6 +1,6 @@
 import path from "node:path";
 
-import { NEXT_DIR, NextConfig } from "../config/index.js";
+import { NEXT_DIR, NextConfig, PrerenderManifest } from "../config/index.js";
 import { InternalEvent, InternalResult } from "../event-mapper.js";
 import { IncomingMessage } from "../http/request.js";
 import { ServerlessResponse } from "../http/response.js";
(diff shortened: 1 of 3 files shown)


vicb (commented):
How has the change been validated?
Should there be a test?

conico974:
@vicb I'm doing some manual test right now, there wasn't any unit test in V2 and e2e is not possible anymore.
It's the exact same change as in V3, i'll update here once i've finished with my manual test.

khuezy (commented):
Would there be any way for a user to obtain the prerender manifest and thus get the preview id?

conico974:
No unless they have access to the build or that the prerender manifest has been leaked, but in those cases there is way more issues than just that

khuezy:
Just want to triple check that the manifest file doesn't somehow end up in the `public` folder, eg. Bots will scan for `.env` files to exploit sites.

packages/open-next/src/adapters/routing/middleware.ts
```diff
@@ -46,7 +46,12 @@ export async function handleMiddleware(
   const hasMatch = middleMatch.some((r) => r.test(rawPath));
   if (!hasMatch) return internalEvent;
   // We bypass the middleware if the request is internal
-  if (internalEvent.headers["x-isr"]) return internalEvent;
+  if (
+    internalEvent.headers["x-isr"] &&
+    internalEvent.headers["x-prerender-revalidate"] ===
```
khuezy, line 51: What if this header is `undefined` and the `previewModeId` is also undefined? This would still be a vulnerability right?
I'm not sure if previewModeId is guaranteed to exist.

conico974: `previewModeId` is always there.
Next relies on this internally as well

khuezy (approved):
LGTM, pending  your manual tests.

conico974:
Ok it works (i forgot how painful it was to test v2).
Gonna merge this, i hope that the prerelease action still works :crossed_fingers:

conico974:
And it does not work anymore, this might be a long night...

PR #814, opened 2025-04-04
fix for fully-static pages router page

conico974:
Fully static pages router page (without `getStaticProps`) can get an incorrect cache-control headers if the `x-middleware-prefetch` header is present


packages/open-next/src/core/routing/util.ts
```diff
@@ -243,7 +243,11 @@ export function fixCacheHeaderForHtmlPages(
   const localizedPath = localizePath(internalEvent);
   // WORKAROUND: `NextServer` does not set cache headers for HTML pages
   // https://opennext.js.org/aws/v2/advanced/workaround#workaround-nextserver-does-not-set-cache-headers-for-html-pages
-  if (HtmlPages.includes(localizedPath)) {
+  // We need to not cache if the request contains an `x-middleware-prefetch` header
```
vicb, line 246:
```suggestion
  // Requests containing an `x-middleware-prefetch` header must not be cached
```

vicb (approved):

PR #817, opened 2025-04-07
Fix for searchParams

conico974:
Fix multivalue query handling in searchParams in Node.
Should fix https://github.com/opennextjs/opennextjs-cloudflare/issues/533
Potentially fix this one as well https://github.com/opennextjs/opennextjs-cloudflare/issues/537

### DIFF
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -5,7 +5,7 @@ import { Readable } from "node:stream";
 import { BuildId, HtmlPages, NextConfig } from "config/index.js";
 import type { IncomingMessage } from "http/index.js";
 import { OpenNextNodeResponse } from "http/openNextResponse.js";
-import { parseHeaders } from "http/util.js";
+import { getQueryFromIterator, parseHeaders } from "http/util.js";
 import type {
   FunctionsConfigManifest,
   MiddlewareManifest,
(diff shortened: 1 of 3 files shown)


packages/open-next/src/core/routing/util.ts
```diff
@@ -122,23 +130,20 @@ export function convertRes(res: OpenNextNodeResponse): InternalResult {
...
 export function convertToQueryString(query: Record<string, string | string[]>) {
-  // URLSearchParams is a representation of the PARSED query.
-  // So we must decode the value before appending it to the URLSearchParams.
-  // https://stackoverflow.com/a/45516812
-  const urlQuery = new URLSearchParams();
+  const queryStrings: string[] = [];
   Object.entries(query).forEach(([key, value]) => {
     if (Array.isArray(value)) {
-      value.forEach((entry) => urlQuery.append(key, decodeURIComponent(entry)));
+      value.forEach((entry) => queryStrings.push(`${key}=${entry}`));
```
SamyPesse, line 140: Why not use `URLSearchParams` here without the `decodeURIComponent`?

Encoding the search parameters manually seems error-prone

conico974: I do remember now why i did that, `convertToQueryString` is only used for already properly encoded search parameters. 
The query in these cases are provided by the converter (which in the case of cloudflare already use `URLSearchParams`)

If we revert this without the `decodeURIComponent` it will break the `handleRewrite` from next config (Specifically this test https://github.com/opennextjs/opennextjs-aws/blob/e9b37fd44eb856eb8ae73168bf455ff85dd8b285/packages/tests-unit/tests/core/routing/matcher.test.ts#L295-L303)

It should be safe enough to use it here as long as we don't use it in places where query could be not properly encoded.

vicb (approved):
Thanks for the fix 👍

PR #819, opened 2025-04-08
add aws-lambda-compressed

### LINKED ISSUE #475: Re-enable nextjs compression or allow enabling it optionally
Although Cloudfront does brotli compression on its own, there are good reasons to have the nextjs lambda compress on its own.to avoid hitting the 6 MB Limit.
In my case when using Payload CMS, a 10 MB file shrinks to ~400 kb using the nextjs gzip compression.

This comment mentions that compression which appareantly caused issues in earlier version actually seems fine now:
https://github.com/sst/open-next/issues/35#issuecomment-1411131129

sommeeeer:
Closes #475 

For some apps you might get a `413` when your response body is too large. This is a limitation in Lambda. In you AWS console you will get this error message:
`[ERROR] [xxxxxxxxx] LAMBDA_RUNTIME Failed to post handler success response. Http response code: 413.`

The default limit in Lambdas is `6mb` for invocation payloads (synchronous request/response). You can read more about its limitations [here](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html)

This PR introduces a new wrapper called `aws-lambda-compressed` that will compress the response body by default if the client sends a request with a supported encoding in the `Accept-Encoding` header. Compression will be applied in the following priority order: `br` (Brotli) → `gzip` → `deflate`. If none of these is found, we just return the body as is. 

After some testing on various bodies of reasonable size (8mb+). I found out that the Lambda will timeout after 10s if we use the default `BROTLI_PARAM_QUALITY` which is `11`. Its also the highest. Therefore defaulting it to `6` is probably wise, as a compromise.

An environment variable `BROTLI_QUALITY` makes this configurable.

### DIFF
--- a/packages/open-next/src/overrides/wrappers/aws-lambda-compressed.ts
+++ b/packages/open-next/src/overrides/wrappers/aws-lambda-compressed.ts
@@ -1,0 +1,136 @@
+import { Readable, Writable } from "node:stream";
+import type { ReadableStream } from "node:stream/web";
+import zlib from "node:zlib";
+
+import type {
+  APIGatewayProxyEvent,
+  APIGatewayProxyEventV2,
+  APIGatewayProxyResult,
+  APIGatewayProxyResultV2,
+  CloudFrontRequestEvent,
+  CloudFrontRequestResult,
+} from "aws-lambda";
+import type { WrapperHandler } from "types/overrides";
+
+import type { InternalResult, StreamCreator } from "types/open-next";
+import { error } from "../../adapters/logger";
+import type {
+  WarmerEvent,
+  WarmerResponse,
+} from "../../adapters/warmer-function";
+
+type AwsLambdaEvent =
+  | APIGatewayProxyEventV2
+  | CloudFrontRequestEvent
+  | APIGatewayProxyEvent
+  | WarmerEvent;
+
+type AwsLambdaReturn =
+  | APIGatewayProxyResultV2
+  | APIGatewayProxyResult
+  | CloudFrontRequestResult
+  | WarmerResponse;
+
+function formatWarmerResponse(event: WarmerEvent) {
+  return new Promise<WarmerResponse>((resolve) => {
+    setTimeout(() => {
+      resolve({ serverId, type: "warmer" } satisfies WarmerResponse);
+    }, event.delay);
+  });
+}
+
+const handler: WrapperHandler =
+  async (handler, converter) =>
+  async (event: AwsLambdaEvent): Promise<AwsLambdaReturn> => {
+    // Handle warmer event
+    if ("type" in event) {
+      return formatWarmerResponse(event);
+    }
+
+    const internalEvent = await converter.convertFrom(event);
+    //TODO: create a simple reproduction and open an issue in the node repo
+    //This is a workaround, there is an issue in node that causes node to crash silently if the OpenNextNodeResponse stream is not consumed
+    //This does not happen everytime, it's probably caused by suspended component in ssr (either via <Suspense> or loading.tsx)
+    //Everyone that wish to create their own wrapper without a StreamCreator should implement this workaround
+    //This is not necessary if the underlying handler does not use OpenNextNodeResponse (At the moment, OpenNextNodeResponse is used by the node runtime servers and the image server)
+    const fakeStream: StreamCreator = {
+      writeHeaders: () => {
+        return new Writable({
+          write: (_chunk, _encoding, callback) => {
+            callback();
+          },
+        });
+      },
+    };
+
+    const acceptEncoding =
+      internalEvent.headers["accept-encoding"] ??
+      internalEvent.headers["Accept-Encoding"] ??
+      "";
+
+    let contentEncoding: string | null = null;
+    if (acceptEncoding?.includes("br")) {
+      contentEncoding = "br";
+    } else if (acceptEncoding?.includes("gzip")) {
+      contentEncoding = "gzip";
+    } else if (acceptEncoding?.includes("deflate")) {
+      contentEncoding = "deflate";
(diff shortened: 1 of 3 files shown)


.changeset/bright-bulldogs-laugh.md
```diff
@@ -0,0 +1,9 @@
+---
+"@opennextjs/aws": patch
+---
+
+Add aws-lambda-compressed wrapper
+
+Introduces a new wrapper called `aws-lambda-compressed`. Will compress the response body by default. Compression will be applied in the following priority order: br (Brotli) → gzip → deflate. If none of these is found, we just return the body as is.
```
conico974, line 7: We don't need that many lines in the changelog to be honest

packages/open-next/src/overrides/wrappers/aws-lambda-compressed.ts
```diff
@@ -0,0 +1,136 @@
...
+const handler: WrapperHandler =
+  async (handler, converter) =>
+  async (event: AwsLambdaEvent): Promise<AwsLambdaReturn> => {
+    // Handle warmer event
+    if ("type" in event) {
+      return formatWarmerResponse(event);
+    }
+
+    const internalEvent = await converter.convertFrom(event);
+    //TODO: create a simple reproduction and open an issue in the node repo
```
conico974, line 51: No need to reexplain everything here, just add a comment pointing to the explanation in aws-lambda

packages/open-next/src/overrides/wrappers/aws-lambda-compressed.ts
```diff
@@ -0,0 +1,136 @@
...
+      contentEncoding = "deflate";
+    }
+
+    const handlerResponse = await handler(internalEvent, {
+      streamCreator: fakeStream,
+    });
+
+    const response: InternalResult = {
+      ...handlerResponse,
+      body: compressBody(handlerResponse.body, contentEncoding),
```
conico974, line 86: We should check if the response hasn't already been compressed before applying compression (it could have been done by the user)

packages/open-next/src/overrides/wrappers/aws-lambda-compressed.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function compressBody(body: ReadableStream, encoding: string | null) {
+  // If no encoding is specified, return original body
+  if (!encoding) return body;
+  try {
+    const readable = Readable.fromWeb(body);
+
+    switch (encoding) {
+      case "br":
+        return Readable.toWeb(
```
conico974, line 111: No need to repeat `Readable.toWeb(readable.pipe(`, just set the zlib transform to a variable and use it at the end

packages/open-next/src/overrides/wrappers/aws-lambda-compressed.ts
```diff
@@ -0,0 +1,136 @@
...
+import type { WrapperHandler } from "types/overrides";
+
+import type { InternalResult, StreamCreator } from "types/open-next";
+import { error } from "../../adapters/logger";
+import type {
+  WarmerEvent,
+  WarmerResponse,
+} from "../../adapters/warmer-function";
+
+type AwsLambdaEvent =
```
conico974, line 22: We can share things like type and `formatWarmerResponse` with the aws-lambda wrapper

packages/open-next/src/overrides/wrappers/aws-lambda-compressed.ts
sommeeeer: Thanks for the review nico!

I made a commit with the other comments, but I was a bit unsure how to check if its already compressed. The plan is to return early if that is the case. Would it be enough to just check if the `content-encoding` header is present, or should I do some `.includes()` and look for all the directives listed here https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Encoding?

packages/open-next/src/types/overrides.ts
```diff
@@ -2,6 +2,15 @@ import type { Readable } from "node:stream";
 
 import type { Meta } from "types/cache";
 
+import type {
```
conico974, line 5: I don't like the idea of exporting aws only stuff from here, just export them from the `aws-lambda` wrapper or create a new file there like `aws-common` or `aws-utils`

packages/open-next/src/utils/overrides.ts
```diff
@@ -0,0 +1,9 @@
+import type { WarmerEvent, WarmerResponse } from "../adapters/warmer-function";
+
+export function formatWarmerResponse(event: WarmerEvent) {
```
conico974, line 3: Same as for the types, just export from the lambda wrapper or from that new file

packages/open-next/src/overrides/wrappers/aws-lambda-compressed.ts
```diff
@@ -0,0 +1,113 @@
...
+const handler: WrapperHandler =
+  async (handler, converter) =>
+  async (event: AwsLambdaEvent): Promise<AwsLambdaReturn> => {
+    // Handle warmer event
+    if ("type" in event) {
+      return formatWarmerResponse(event);
+    }
+
+    const internalEvent = await converter.convertFrom(event);
+    // This is a workaround, you can read more about it in the aws-lambda wrapper
```
conico974, line 23: Add a permalink

packages/open-next/src/overrides/wrappers/aws-lambda-compressed.ts
conico974: Just check if the header is there, it should be enough

conico974 (approved):
LGTM Thanks

PR #821, opened 2025-04-10
Add an option for createStaticAssets to respect the basePath of Next.js

vicb:
The plan is to use this in cloudflare so that the path it handled by the Workers Assets automatically when a basePath is configured

### DIFF
--- a/packages/open-next/src/build/createAssets.ts
+++ b/packages/open-next/src/build/createAssets.ts
@@ -1,22 +1,41 @@
 import fs from "node:fs";
 import path from "node:path";
 
+import { loadConfig } from "config/util.js";
 import logger from "../logger.js";
 import type { TagCacheMetaFile } from "../types/cache.js";
 import { isBinaryContentType } from "../utils/binary.js";
 import * as buildHelper from "./helper.js";
 
-export function createStaticAssets(options: buildHelper.BuildOptions) {
+/**
+ * Copy the static assets to the output folder
+ *
+ * WARNING: `useBasePath` should be set to `false` when the output file is used.
+ *
+ * @param options OpenNext build options
+ * @param useBasePath whether to copy files into the to Next.js configured basePath
+ */
+export function createStaticAssets(
+  options: buildHelper.BuildOptions,
+  { useBasePath = false } = {},
+) {
   logger.info("Bundling static assets...");
 
   const { appBuildOutputPath, appPublicPath, outputDir, appPath } = options;
 
+  const NextConfig = loadConfig(path.join(appBuildOutputPath, ".next"));
+  const basePath = useBasePath ? (NextConfig.basePath ?? "") : "";
+
   // Create output folder
-  const outputPath = path.join(outputDir, "assets");
+  const outputPath = path.join(outputDir, "assets", basePath);
   fs.mkdirSync(outputPath, { recursive: true });
 
   /**
-   * Next.js outputs assets into multiple files. Copy into the same directory.
+   * Next.js outputs assets into multiple files.
+   *
+   * Copy into the same directory:
+   * - `.open-next/assets` when `useBasePath` is `false`
+   * - `.open-next/assets/basePath` when `useBasePath` is `true`
    *
    * Copy over:
    * - .next/BUILD_ID => BUILD_ID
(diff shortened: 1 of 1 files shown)


vicb:
> If we do that we'll need to modify this file as well otherwise it will break existing aws deployment https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/build/generateOutput.ts I'm wondering if the good long term solution would be to actually use that output file instead ?

Yes, it's more a short term hack for cloudflare...
From my testing it look like only the `_next` dir is prefixed and others are not touched.
Has the output file been tested with a basePath?

conico974 (commented):
Yeah it is used in sst and other places, but i've missed the `useBasePath` which should make it do nothing there.
FYI every places that use [this function](https://github.com/opennextjs/opennextjs-aws/blob/73281c958d43d865fdded0d5d86dd82747365fee/packages/open-next/src/build/generateOutput.ts#L191) need to use the basePath (this includes public file)

conico974:
I'm not 100% sure either, my first idea would be that cloudflare (or anyone else) should use the output file to actually finalize the deployment from the created assets from OpenNext.

For example these 2 lines: https://github.com/opennextjs/opennextjs-aws/blob/73281c958d43d865fdded0d5d86dd82747365fee/packages/open-next/src/build/generateOutput.ts#L201-L202  would not be correct in aws if we use the base path in `createStaticAssets` because files would be at `.open-next/assets/basePath` and would thus create an `_assets/basePath/basePath` folder.

vicb:
> would not be correct in aws if we use the base path in createStaticAssets because files would be at .open-next/assets/basePath and would thus create an _assets/basePath/basePath folder.

The thing I don't get with this code is that static assets do not seem to be prefixed by the basePath? 
I see that only `_next/...` is

conico974:
On aws it's on the behavior for cloudfront, not in the bucket themselves

packages/open-next/src/build/createAssets.ts
```diff
@@ -30,9 +40,13 @@ export function createStaticAssets(options: buildHelper.BuildOptions) {
...
+  const NextConfig = loadConfig(path.join(appBuildOutputPath, ".next"));
+  const basePath = useBasePath ? (NextConfig.basePath ?? "") : "";
+
   fs.cpSync(
     path.join(appBuildOutputPath, ".next/static"),
-    path.join(outputPath, "_next", "static"),
+    path.join(outputPath, basePath, "_next", "static"),
     { recursive: true },
   );
   if (fs.existsSync(appPublicPath)) {
```
conico974, line 52: Everything in here should be under `outputPath/basePath`

conico974 (commented):
@vicb What do you want to do with this one ? I'm fine with adding it like that for now as long as we add some comment about it (it won't have any effect on aws since we won't use `useBasePath`)

packages/open-next/src/build/createAssets.ts
vicb: What do you mean by "everything in here"?

Looking at the [blog example](https://github.com/opennextjs/opennextjs-cloudflare/blob/2ce64546165b5bbcc1718649bf91b599aa01594c/examples/vercel-blog-starter/src/app/_components/avatar.tsx#L9), path are not re-written. The avatar will not be fetch for the basePath. Only `<Image/>` and <Link/>` will use the base path.

conico974: Favicon and assets in public should also start with `basePath`. That's the behavior in `next start`

vicb: That's not the behavior I observe on the blog example with `next start`?

conico974: If i add the `/blog` basePath that's what i get 
image

vicb: Oops, you'right.

What I did with the blog example is that I probably used `npm preview` and make that work.
As the path are nor rewritten, the assets should not be prefixed if you want to access them.
But it must be an overlook in the blog example.

In conclusion, you are right, everything must be prefixed.

Would that work if the copy is unconditionally done to the basePath subDir and the basePath logic is removed from the output file? I think this could be a way to support Cloudflare & AWS?

conico974: We need to add it to the behavior as well anyway (to avoid intercepting things it should not). 
I think we go for your solution for now, just add a comment about it, and try to figure out a way to create an output that could be used by everyone.

I had this idea awhile back to add an `outputOverride` to be able to compute an output that could be used in each platform easily. This inside something like an `open-next.build.ts` could be something that could help us in a lot of ways.

vicb:
I updated the PR to add a comment, does it looks good like this?

conico974 (approved):
LGTM Thanks

vicb:
Thanks Nico!

PR #823, opened 2025-04-12
Fix middleware fetch

conico974:
Fix 2 small issues with the middleware.
- Directly returning a fetch from the middleware crashed because of immutable headers
- Returning a `Response` with an empty body which was not a redirect was not working

### DIFF
--- a/packages/open-next/src/core/routing/middleware.ts
+++ b/packages/open-next/src/core/routing/middleware.ts
@@ -111,13 +111,15 @@ export async function handleMiddleware(
   const reqHeaders: Record<string, string> = {};
   const resHeaders: Record<string, string | string[]> = {};
 
-  responseHeaders.delete("x-middleware-override-headers");
-  /*  Next will set the header `x-middleware-set-cookie` when you `set-cookie` in the middleware.
-   *  We can delete it here since it will be set in `set-cookie` aswell. Next removes this header in the response themselves.
-   * `x-middleware-next` is set when you invoke `NextResponse.next()`. We can delete it here aswell.
-   */
-  responseHeaders.delete("x-middleware-set-cookie");
-  responseHeaders.delete("x-middleware-next");
+  // These are internal headers used by Next.js, we don't want to expose them to the client
+  const filteredHeaders = [
+    "x-middleware-override-headers",
+    "x-middleware-set-cookie",
+    "x-middleware-next",
+    "x-middleware-rewrite",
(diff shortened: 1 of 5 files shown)


packages/open-next/src/core/routing/middleware.ts
```diff
@@ -111,20 +111,23 @@ export async function handleMiddleware(
...
+    "content-encoding",
+  ];
 
   const xMiddlewareKey = "x-middleware-request-";
   responseHeaders.forEach((value, key) => {
     if (key.startsWith(xMiddlewareKey)) {
       const k = key.substring(xMiddlewareKey.length);
       reqHeaders[k] = value;
     } else {
+      if (filteredHeaders.includes(key.toLowerCase())) return;
```
vicb, line 130: Do we want to lowercase once at the start of the fn body?
That might also fix a bug with `key.startsWith(xMiddlewareKey)`

packages/tests-e2e/tests/appPagesRouter/middleware.test.ts
```diff
@@ -0,0 +1,20 @@
+import { expect, test } from "@playwright/test";
+
+test("should return correctly on HEAD request with an empty body", async ({
+  request,
+}) => {
+  const response = await request.head("/head");
+  const body = await response.text();
+  expect(response.status()).toBe(200);
```
vicb, line 8: bit
```suggestion
    expect(response.status()).toBe(200);
    const body = await response.text();
```

packages/tests-e2e/tests/appPagesRouter/middleware.test.ts
```diff
@@ -0,0 +1,20 @@
...
+  expect(body).toBe("");
+  expect(response.headers()["x-from-middleware"]).toBe("true");
+});
+
+test("should return correctly for directly returning a fetch response", async ({
+  request,
+}) => {
+  const response = await request.get("/fetch");
+  const body = await response.json();
+  expect(response.status()).toBe(200);
```
vicb, line 18: same nit
```suggestion
  expect(response.status()).toBe(200);
  const body = await response.json();
```

vicb (approved):

packages/open-next/src/core/routing/middleware.ts
conico974: That's not necessary, those are added by Next and they should always be lowercase.

PR #825, opened 2025-04-14
Fix remove prefetch header for Next 14.1+

conico974:
For Next 14.1, we'll rely on a patch to disable background revalidation instead of adding a `purpose:prefetch` header to every request.
Should fix https://github.com/opennextjs/opennextjs-cloudflare/issues/567

### DIFF
--- a/packages/tests-unit/tests/build/patch/patches/patchBackgroundRevalidation.test.ts
+++ b/packages/tests-unit/tests/build/patch/patches/patchBackgroundRevalidation.test.ts
@@ -1,0 +1,43 @@
+import { patchCode } from "@opennextjs/aws/build/patch/astCodePatcher.js";
+import { rule } from "@opennextjs/aws/build/patch/patches/patchBackgroundRevalidation.js";
+import { describe, it } from "vitest";
+
+const codeToPatch = `if (cachedResponse && !isOnDemandRevalidate) {
+                    var _cachedResponse_value;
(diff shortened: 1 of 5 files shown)


packages/tests-unit/tests/build/patch/patches/patchBackgroundRevalidation.test.ts
```diff
@@ -0,0 +1,43 @@
...
+                    resolved = true;
+                    if (!cachedResponse.isStale || context.isPrefetch) {
+                        // The cached value is still valid, so we don't need
+                        // to update it yet.
+                        return null;
+                    }
+                }`;
+
+describe("patchBackgroundRevalidation", () => {
+  it("Should patch code for Next 14+", () => {
```
vicb, line 23: Nit:  The version should be >= 14.1 but I don't think it needs to be specified here?
```suggestion
  it("Should patch code", () => {
```

vicb (approved):
LGTM, thanks!

> Should fix https://github.com/opennextjs/opennextjs-cloudflare/issues/567

It needs to be ported to the [cloudflare repo](https://github.com/opennextjs/opennextjs-cloudflare/blob/2ce64546165b5bbcc1718649bf91b599aa01594c/packages/cloudflare/src/cli/build/open-next/createServerBundle.ts) - looks like we are missing some patches from the last aws version bump as well.

PR #827, opened 2025-04-14
refactor: export patchBackgroundRevalidation from patches/index.js

vicb:
sorry @conico974 I missed that during the review


conico974 (approved):
Oops missed it as well.
I guess for now we'll just import directly

PR #828, opened 2025-04-14
fix(tagCache): do not call writeTags with an empty list

vicb:
See https://github.com/opennextjs/opennextjs-cloudflare/issues/576

Called from `node_modules/next/dist/compiled/next-server/app-route.runtime.dev.js`

```js
_workStore_incrementalCache.revalidateTag(workStore.revalidatedTags||[])
```


conico974 (approved):
Thanks

vicb:
Thanks for the review!

Issue #826, opened 2025-04-14
Incorrect StatusCode causing monitoring to flag


iDVB:
We're seeing a ton of error alerts in this function. At closer inspection we see that its simply when a url is referenced without a required querystring parameter. Eg. "url" or "w". Currently the resulting statusCode is "500" which is why our Monitoring is flagging it. 

Shouldn't this be a **4xx series** error code? Likely **400 Bad Request** ?

>The HTTP 400 Bad Request response status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (for example, malformed request syntax, invalid request message framing, or deceptive request routing).

# Examples
- StatusCode: 500 `"url" parameter is required`
- StatusCode: 500 `"w" parameter (width) is required`

# NextJs correctly marks these as "400" errors (not 500)

![Image](https://github.com/user-attachments/assets/d40bac94-7a2b-4297-8689-87ddd130dfce)

# While OpenNext, marks them as 500

![Image](https://github.com/user-attachments/assets/3434bc4d-f4f3-4d30-87cc-ad4cbe530d1b)

khuezy:
Ya that makes sense to do. Can you open a PR to address around https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/adapters/image-optimization-adapter.ts#L184

I'm not sure if we need to just return the status code from the response, or need to do some logic to decide what error code to return.

iDVB:
Ya, saw that and was thinking to leave that the same... but take the statusCode from the original event object.

khuezy:
I thought there was some work to allow you to run ON locally but I haven't done it myself. @conico974 would know but it might be quicker just to deploy a test app until he responds.

conico974:
@iDVB https://opennext.js.org/aws/contribute/local_run. Be aware there is some issues with monorepo for dev overrides, just don't test on a monorepo.
You want to modify the status code both here https://github.com/opennextjs/opennextjs-aws/blob/199a26f1d3fbb2acb494e77f599e558a7264b988/packages/open-next/src/adapters/image-optimization-adapter.ts#L205 and here https://github.com/opennextjs/opennextjs-aws/blob/199a26f1d3fbb2acb494e77f599e558a7264b988/packages/open-next/src/adapters/image-optimization-adapter.ts#L195
BTW i think we'll need to infer the status code from the error itself (and return 500 by default)

iDVB:
[...]

PR #830, opened 2025-04-14
fix: should return 400 in image optimization when Next passes errorMessage

sommeeeer:
For #826 

In versions 13, 14, and 15, `next start` returns a 400 Bad Request if `validateImageParams` returns an `errorMessage`. We should align with this behavior.


khuezy (approved):
LGTM thx!

packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -184,6 +190,7 @@ function buildSuccessResponse(
 function buildFailureResponse(
   e: any,
```
conico974, line 191: We might as well replace this with a string representing the error message we want to display

conico974 (approved):
LGTM Thanks

PR #831, opened 2025-04-15
fix: only report actually copied traced files

vicb:


conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review

PR #832, opened 2025-04-16
E2E: Add a new app to test experimental features

conico974:
Create a new e2e app to allow to test some of the experimental features of Next (i.e. ppr, dynamicIO...) that are only available on canary releases


examples/experimental/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "scripts": {
+    "openbuild": "node ../../packages/open-next/dist/index.js build",
+    "dev": "next dev --turbopack --port 3004",
+    "build": "next build",
+    "start": "next start --port 3004",
+    "lint": "next lint",
+    "clean": "rm -rf .turbo node_modules .next .open-next"
+  },
+  "dependencies": {
+    "next": "15.3.1-canary.8",
```
khuezy, line 14: Will this always be hard coded? I think you can install `next@canary` to pull the latest canary.

conico974: Yes, this is on purpose actually.
I think we don't want change in canary to break our e2e (especially since sometimes it will get reverted in the next canary)
We'll need to update it from time to time to keep it kind of in sync with next. Ideally at least once before every release

examples/experimental/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "experimental",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "openbuild": "node ../../packages/open-next/dist/index.js build",
```
vicb, line 6: question: this looks like a dev script, right?

examples/experimental/src/components/dynamic.tsx
```diff
@@ -0,0 +1,18 @@
+import { headers } from "next/headers";
+
+export async function DynamicComponent() {
+  const _headers = await headers();
+  // Simulate a delay to mimic server-side calls
+  const date = await new Promise((resolve) =>
```
vicb, line 6: nit: simplify using `import { setTimeout } from 'node:timers/promises';`?

packages/tests-e2e/tests/experimental/nodeMiddleware.test.ts
```diff
@@ -0,0 +1,29 @@
+import { expect, test } from "@playwright/test";
+
+test("Node middleware should add headers", async ({ request }) => {
```
vicb, line 3: nit: enclose in a `test.describe("node middlewre")` ?

packages/tests-e2e/tests/experimental/ppr.test.ts
```diff
@@ -0,0 +1,22 @@
+import { expect, test } from "@playwright/test";
+
+test("PPR should show loading first", async ({ page }) => {
```
vicb, line 3: nit: enclose in a test.describe("ppr")

vicb (approved):
LGTM, a few minor comments

examples/experimental/package.json
conico974: This is the OpenNext build script, i just kept it this way for consistency with the other e2e app

examples/experimental/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "dev": "next dev --turbopack --port 3004",
+    "build": "next build",
+    "start": "next start --port 3004",
+    "lint": "next lint",
+    "clean": "rm -rf .turbo node_modules .next .open-next"
+  },
+  "dependencies": {
```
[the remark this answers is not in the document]
conico974: We are actually not using it directly right now

PR #835, opened 2025-04-18
trace node packages

vicb:
In the cloudflare adapter, we want to copy files used via the "workerd" build conditions.

To be able to do this this PR updates `copyTracedFiles` to return a map of traced node packages.
It maps the original path of the package to the path in `.open-next`.

As a follow up, we might integrate that better with the patch system.
Today, `applyCodePatches` assumes that a patch will update file content.
We could modify `applyCodePatches` to take a list of `Patch`. 
The `Patch` class would be instantiated with `options, tracedFiles, manifests, nodePackages, nextVersion` and have a public `apply` method.
We could then have derived `PatchFileContent`, `PatchCopyFile`, ... classes.

### DIFF
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -89,40 +89,46 @@ export async function copyTracedFiles({
   const standaloneServerDir = path.join(standaloneNextDir, "server");
   const outputNextDir = path.join(outputDir, packagePath, ".next");
 
-  // On next 14+, we might not have to include those files
-  // For next 13, we need to include them otherwise we get runtime error
-  const requiredServerFiles = JSON.parse(
-    readFileSync(
-      path.join(
-        dotNextDir,
-        bundledNextServer
-          ? "next-minimal-server.js.nft.json"
-          : "next-server.js.nft.json",
-      ),
-      "utf8",
-    ),
-  );
-
   // Files to copy
   // Map from files in the `.next/standalone` to files in the `.open-next` folder
   const filesToCopy = new Map<string, string>();
 
   // Node packages
   // Map from folders in the project to folders in the `.open-next` folder
+  // The map might also include the mono-repo path.
   const nodePackages = new Map<string, string>();
 
-  // Files necessary by the server
-  if (!skipServerFiles) {
-    requiredServerFiles.files.forEach((tracedPath: string) => {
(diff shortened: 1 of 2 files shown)


packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -156,27 +168,37 @@ See the docs for more information on how to bundle edge runtime functions.
...
 --------------------------------------------------------------------------------`);
     }
-    const dir = path.dirname(fullFilePath);
-    extractFiles(
-      requiredFiles.files,
-      path.join(standaloneNextDir, dir),
-    ).forEach((f) => {
-      filesToCopy.set(f, f.replace(standaloneDir, outputDir));
+
+    requiredFiles.files.forEach((tracedPath: string) => {
```
conico974, line 175: Only difference between this one, and the one above is the `serverDir`, we could create a function for that

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -106,12 +103,25 @@ export async function copyTracedFiles({
     ),
   );
 
+  // Files to copy
+  // Map from files in the `.next/standalone` to files in the `.open-next` folder
   const filesToCopy = new Map<string, string>();
 
+  // Node packages
+  // Map from folders in the project to folders in the `.open-next` folder
+  const nodePackages = new Map<string, string>();
```
conico974, line 112: This may also includes monorepo package.json, not only the ones from node_modules. We could add a comment about it here.
It probably does not matter that much though

conico974:
> As a follow up, we might integrate that better with the patch system.
Today, applyCodePatches assumes that a patch will update file content.
We could modify applyCodePatches to take a list of Patch.
The Patch class would be instantiated with options, tracedFiles, manifests, nodePackages, nextVersion and have a public apply method.
We could then have derived PatchFileContent, PatchCopyFile, ... classes.

That's a good idea yeah.
One other improvements that would be great to have would be for ast-grep to avoid parsing the same files multiple times, (but i think it involves changes in ast-grep itself)

conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review Nico!
When do you plan the next release?

conico974:
@vicb Later today, just let me know if you want to add something in it

Issue #842, opened 2025-04-28
[Test] Add e2e for next/head


conico974:
Add e2e tests for next/head and similar (i.e. page router context)

PR #845, opened 2025-04-30
add e2e for next/head and next/amp in pages router

sommeeeer:
For #842

### DIFF
--- a/packages/tests-e2e/tests/pagesRouter/amp.test.ts
+++ b/packages/tests-e2e/tests/pagesRouter/amp.test.ts
@@ -5,5 +5,7 @@ test.describe("next/amp", () => {
     await page.goto("/amp");
     const timeago = await page.getByTestId("amp-timeago").textContent();
     expect(timeago).toBe("just now");
+    const htmlEl = page.locator("html");
+    await expect(htmlEl).toHaveAttribute("amp");
   });
 });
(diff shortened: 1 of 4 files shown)


packages/tests-e2e/tests/pagesRouter/amp.test.ts
```diff
@@ -0,0 +1,11 @@
+import { expect, test } from "@playwright/test";
+
+test.describe("next/amp", () => {
+  test("should load and display the timeago component", async ({ page }) => {
+    await page.goto("/amp");
+    const timeago = await page.getByTestId("amp-timeago").textContent();
+    expect(timeago).toBe("just now");
```
conico974, line 7: How precise is it ?
Could it display a `1s ago` ?

sommeeeer: > How precise is it ?

In theory it should be always showing `just now` here as the value is calculated at render time, and that wont take many ms. Its using `format()` from  [`timeago.js`](https://www.npmjs.com/package/timeago.js/v/4.0.0-beta.3) under the hood, and it will show `just now` until it reaches 10s, then it will start showing `11 seconds ago`.

sommeeeer: I can add a comment about it in the test.

conico974 (approved):

conico974:
LGTM Thanks

PR #849, opened 2025-05-02
fix use cache flaky test

conico974:
Because both the revalidateTag and the ssr test could run at the same time, it could cause the ssr test to fail because `revalidateTag` would have been called by the other test

### DIFF
--- a/examples/experimental/src/app/use-cache/ssr/page.tsx
+++ b/examples/experimental/src/app/use-cache/ssr/page.tsx
@@ -1,4 +1,8 @@
-import { FullyCachedComponent, ISRComponent } from "@/components/cached";
+import {
+  FullyCachedComponent,
+  FullyCachedComponentWithTag,
+  ISRComponent,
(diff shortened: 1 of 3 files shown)


khuezy (approved):
LGTM

Slightly off topic: `use cache` defaults to 15 minutes but I don't think we'll ever hit the scenario where the github action builds the app to the e2e test is > 15minutes, which may cause some flakiness in the tests. But if that ever does happen in the future, that could explain why.

conico974:
Thanks

> Slightly off topic: use cache defaults to 15 minutes but I don't think we'll ever hit the scenario where the github action builds the app to the e2e test is > 15minutes, which may cause some flakiness in the tests. But if that ever does happen in the future, that could explain why.

It's even less likely that what you think. `use cache` right now does not produce build time cache, so for it to be an issue, it would need to be 15 minutes between the first and last request to the same route. And worst case scenario, we just add `unstable_cachelife`

Issue #850, opened 2025-05-03
Cookies set in middleware are unavailable on initial render


dnewbound0:
Currently, the `x-middleware-set-cookie` header is being removed from the middleware response: https://github.com/opennextjs/opennextjs-aws/blob/c212ed0213ff7b57817e17fc21ab1c0b7c77a41b/packages/open-next/src/core/routing/middleware.ts#L116-L123
When reading `cookies()` the value of this header is merged into the existing cookies object: https://github.com/vercel/next.js/blob/f62a135d9bde6e92b9f8f50d236b5de02de74400/packages/next/src/server/async-storage/request-store.ts#L83
https://github.com/vercel/next.js/blob/f62a135d9bde6e92b9f8f50d236b5de02de74400/packages/next/src/server/async-storage/request-store.ts#L211
So removing this header breaks reading a cookie that was set by the middleware for that request, as shown here

`middleware.ts`:
```ts
export function middleware(request: NextRequest) {
  const response = NextResponse.next()
  response.cookies.set("foo", "bar")

  return response
}
```
`app/page.tsx`:
```ts
import { cookies } from "next/headers";

export default async function Page() {
  // On initial render, this will be undefined!
  const foo = cookies().get("foo")?.value

  return <div>{foo}</div>
}
```

sommeeeer:
Great catch.

PR #851, opened 2025-05-03
fix: Ensure cookies set in middleware are available on initial render

sommeeeer:
For #850

We should remove `x-middleware-set-cookie` from `filteredHeaders`. Instead we remove it on the `overwrittenResponseHeaders` which will be passed to the response. 

`reqProps` in `requestHandler.ts` will use the original `headers` object, and pass it into `processRequest`. This makes sure we will see the header on initial render when using `cookies().get()` from Next.

### DIFF
--- a/packages/open-next/src/core/requestHandler.ts
+++ b/packages/open-next/src/core/requestHandler.ts
@@ -92,6 +92,8 @@ export async function openNextHandler(
         delete headers[rawKey];
       }
 
+      delete overwrittenResponseHeaders["x-middleware-set-cookie"];
+
       if (
         "isExternalRewrite" in routingResult &&
         routingResult.isExternalRewrite === true
--- a/packages/open-next/src/core/requestHandler.ts
+++ b/packages/open-next/src/core/requestHandler.ts
@@ -87,12 +87,14 @@ export async function openNextHandler(
           continue;
         }
         const key = rawKey.slice(MIDDLEWARE_HEADER_PREFIX_LEN);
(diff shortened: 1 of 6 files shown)


packages/open-next/src/core/requestHandler.ts
```diff
@@ -92,6 +92,8 @@ export async function openNextHandler(
         delete headers[rawKey];
       }
 
+      delete overwrittenResponseHeaders["x-middleware-set-cookie"];
```
khuezy, line 95: Should we add a comment here to note the reason why this header needs to be deleted here?

dnewbound0:
apologies if this is a dumb q, but is there any reason why we can't move all the`filteredHeaders` logic to `requestHandler.ts`?

packages/open-next/src/core/requestHandler.ts
sommeeeer: Yeah will do that. 

I might refactor this to be a part of the `for` loop above, and just skip adding the header to `overwrittenResponseHeaders`. Like this:
```ts
if (key !== "x-middleware-set-cookie") {
  overwrittenResponseHeaders[key] = value;
}
```

What should I go for?

khuezy: Either would be fine. `delete` opts out of the v8 optimization, so maybe setting it to `undefined` would be better?

khuezy (approved):
LGTM thanks!

conico974 (approved):
LGTM Thanks

Issue #541, opened 2024-10-10
[Feature] Implement a local mode


conico974:
Allow to run OpenNext locally for testing/debugging purposes.

It should be relatively easy, as everything is already in the doc

conico974:
Thanks, but this issue is more a reminder of what i need to do as a PR.
Right now it's a hassle to run OpenNext (not Next) locally, you have to manually create a bunch of wrapper converter and other stuff.
Implementing a `@opennextjs/aws local` or `@opennextjs/aws dev` command would be great and helpful for people contributing

conico974:
Basically it would do this https://opennext.js.org/aws/contribute/local_run, but in one command

PR #853, opened 2025-05-04
add command to build and run OpenNext locally with dev overrides

sommeeeer:
For #541

I'll open this in draft mode for now. It wont work in a monorepo as of now. Im open for suggestions.

### DIFF
--- a/packages/open-next/src/index.ts
+++ b/packages/open-next/src/index.ts
@@ -9,9 +9,9 @@ import { printHeader } from "./build/utils.js";
 
 const command = process.argv[2];
 const stage = process.argv[3];
-const validCommand = 
-  command === "build" || 
-  command === "preview" || 
+const validCommand =
+  command === "build" ||
+  command === "preview" ||
   (command === "generate" && stage === "local");
 
 if (!validCommand) {


conico974:
@sommeeeer Honestly i'm not sure what to do with this. The only point right now of such a command is for people contributing to OpenNext, and you'll likely end up with your own config (at least to test things out).
`dev` is not a good name either, it will be misleading.
It's also not like `preview` in cloudflare (which use the exact same things as deployed), here every overrides are dev ones.

Best things to do would probably just be to make it work for monorepo, and if we really want to have a `dev` command (with a different name) it should just call build with another config path (i.e. `open-next.config.dev.ts` for example) and run the server right away

sommeeeer:
> `dev` is not a good name either, it will be misleading.

Yeah the name should not be dev. `local` could perhaps be a better fit? 

> Best things to do would probably just be to make it work for monorepo, and if we really want to have a `dev` command (with a different name) it should just call build with another config path (i.e. `open-next.config.dev.ts` for example) and run the server right away

Should I continue on this one or perhaps close it? This was mainly for people working on OpenNext as it can be a bit cumbersone to copy and make config files each time you need to test something quick locally. Like features in the routingHandler or similar.

conico974:
Why not just a command to generate the base "dev" config instead. I feel like this will give us enough

packages/open-next/src/index.ts
```diff
@@ -1,15 +1,31 @@
...
+  printHelp();
 
 const args = parseArgs();
 if (Object.keys(args).includes("--help")) printHelp();
 
-await build(args["--config-path"], args["--node-externals"]);
-
+if (command === "build") {
+  await build(args["--config-path"], args["--node-externals"]);
+} else if (command === "local:config") {
```
conico974, line 23: It should probably be something like `generate local`, this way later we could add other ones like `generate node` or `generate streaming`

packages/open-next/src/index.ts
```diff
@@ -1,15 +1,31 @@
...
+  writeLocalConfig();
+} else if (command === "local:run") {
```
conico974, line 25: Not a big fan of `local:run`, maybe something similar to cloudflare like `preview`

packages/open-next/src/build/constant.ts
```diff
@@ -1,3 +1,29 @@
...
+    },
+  },
+  imageOptimization: {
+    override: {
+      wrapper: "dummy",
+      converter: "dummy",
+    },
+    loader: "fs-dev",
+    install: {
+      arch: "x64",
```
conico974, line 25: We should add comment here for people that use other arch/system

PR #857, opened 2025-05-07
chore: make dev overrides work in monorepo

sommeeeer:


packages/open-next/src/build/createServerBundle.ts
```diff
@@ -302,7 +302,9 @@ async function generateBundle(
       outfile: path.join(outputPath, packagePath, `index.${outfileExt}`),
       banner: {
         js: [
-          `globalThis.monorepoPackagePath = "${packagePath}";`,
```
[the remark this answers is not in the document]
conico974: By us maybe, but you could use it in your custom overrides, we should apply it every time.
You could add it to the globals though

packages/open-next/src/overrides/incrementalCache/fs-dev.ts
```diff
@@ -4,7 +4,14 @@ import fs from "node:fs/promises";
 import path from "node:path";
 
 const buildId = process.env.NEXT_BUILD_ID;
-const basePath = path.resolve(process.cwd(), `../../cache/${buildId}`);
+const basePath = path.join(
+  globalThis.monorepoPackagePath
```
conico974, line 8: You could make it as a function in `utils/normalize-path`

packages/open-next/src/overrides/incrementalCache/fs-dev.ts
```diff
@@ -4,7 +4,14 @@ import fs from "node:fs/promises";
...
+    .split("/")
+    .filter(Boolean)
+    .map(() => "../")
```
conico974, line 11:
```suggestion
    .map(() => "..")
```

packages/open-next/src/overrides/incrementalCache/fs-dev.ts
```diff
@@ -4,7 +4,14 @@ import fs from "node:fs/promises";
...
+    .join(""),
```
conico974, line 12:
```suggestion
    .join("/"),
```

conico974 (approved):
LGTM Thanks

PR #862, opened 2025-05-13
fix(middleware): copy wasm files for middleware

naikaayush:
prisma's wasm was not being copied into the middleware directory, which was throwing errors when prisma was being used in the middleware with the authjs database adaptor.

added the copyMiddlewareResources which is already being used in createServerBundle

### DIFF
--- a/.changeset/popular-candles-watch.md
+++ b/.changeset/popular-candles-watch.md
@@ -1,5 +1,5 @@
 ---
-"@opennextjs/aws": major
+"@opennextjs/aws": patch
 ---
 
-fix(middleware): copy wasm files for bundled middleware
+fix(middleware): copy wasm files for the external middleware
--- a/.changeset/popular-candles-watch.md
+++ b/.changeset/popular-candles-watch.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": major
+---
+
+fix(middleware): copy wasm files for bundled middleware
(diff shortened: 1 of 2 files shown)


.changeset/popular-candles-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": major
+---
+
+fix(middleware): copy wasm files for bundled middleware
```
conico974, line 5: It's for the `external` middleware that there is an issue right ?

.changeset/popular-candles-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": major
```
conico974, line 2:
```suggestion
"@opennextjs/aws": patch
```

packages/open-next/src/build/createMiddleware.ts
```diff
@@ -51,8 +51,10 @@ export async function createMiddleware(
     }
   }
 
+  const outputPath = path.join(outputDir, "middleware");
+  copyMiddlewareResources(options, edgeMiddlewareInfo, outputPath);
```
conico974, line 55: We are already copying the files in `createServerBundle` when the middleware is bundled

https://github.com/opennextjs/opennextjs-aws/blob/ae1afbb6053c34b3523749fb71d5068e7b24b6f5/packages/open-next/src/build/createServerBundle.ts#L175-L190

Both lines should go under `if(config.middleware?.external)`

conico974 (approved):
Thanks for the PR.
I went ahead and made the requested updates since I hadn't heard back from you

naikaayushnz:
@conico974 sorry about that, I missed it, thanks for the changes

PR #863, opened 2025-05-16
Redirect requests with repeated slashes

dnewbound0:
Let's say you had the following middleware function:
```ts
export async function middleware(request: NextRequest) {
  return NextResponse.redirect(new URL("/foo", request.url))
}
```
A request to `https://yourdomain.com//evil.com` would result in `request.url ===  "https://evil.com"`, leading to a malicious redirect. Here we redirect these bad requests containing repeated slashes to a sanitised path.

### DIFF
--- a/packages/open-next/src/core/routing/matcher.ts
+++ b/packages/open-next/src/core/routing/matcher.ts
@@ -263,6 +263,10 @@ export function handleRewrites<T extends RewriteDefinition>(
   };
 }
 
+// Normalizes repeated slashes in the path e.g. hello//world -> hello/world
+// or backslashes to forward slashes. This prevents requests such as //domain
+// from invoking the middleware with `request.url === "domain"`.
+// See: https://github.com/vercel/next.js/blob/3ecf087f10fdfba4426daa02b459387bc9c3c54f/packages/next/src/server/base-server.ts#L1016-L1020
 function handleRepeatedSlashRedirect(
   event: InternalEvent,
 ): false | InternalResult {
--- a/packages/open-next/src/core/routing/matcher.ts
+++ b/packages/open-next/src/core/routing/matcher.ts
@@ -10,7 +10,7 @@ import type {
 } from "types/next-types";
 import type { InternalEvent, InternalResult } from "types/open-next";
 import { emptyReadableStream, toReadableStream } from "utils/stream";
-import { normalizePath } from "utils/normalize-path"
+import { normalizeRepeatedSlashes } from "utils/normalize-path"
 
 import { debug } from "../../adapters/logger";
 import { handleLocaleRedirect, localizePath } from "./i18n";
@@ -267,15 +267,12 @@ function handleRepeatedSlashRedirect(
   event: InternalEvent,
 ): false | InternalResult {
   // Redirect `https://example.com//foo` to `https://example.com/foo`.
-  const url = new URL(event.url);
-  if (url.pathname.match(/(\\|\/\/)/)) {
+  if (event.rawPath.match(/(\\|\/\/)/)) {
     return {
       type: event.type,
       statusCode: 308,
       headers: {
-        Location: `${url.protocol}//${url.host}${normalizePath(url.pathname)}${
-          url.search
-        }`,
+        Location: normalizeRepeatedSlashes(new URL(event.url)),
       },
       body: emptyReadableStream(),
       isBase64Encoded: false,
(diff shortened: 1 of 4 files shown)


conico974:
@dnewbound0 I'm sorry but i don't understand what you're trying to solve here.
If you had this in your middleware, it would redirect to `https://yourdomain.com/foo`, that's the case for me on my test.
And modifying `handleRedirects` have absolutely no effect since middleware redirect won't go there at all

sommeeeer:
@dnewbound0 if you had that middleware function you would create an infinite redirect loop

dnewbound0:
> @dnewbound0 I'm sorry but i don't understand what you're trying to solve here. If you had this in your middleware, it would redirect to `https://yourdomain.com/foo`, that's the case for me on my test. And modifying `handleRedirects` have absolutely no effect since middleware redirect won't go there at all

(i've replied to this in the OpenNext discord already but i figured i'd post my response here as well for posterity)

i can recreate this using OpenNext  v3.6.1, SST v2.48.5, and Next.js v14.2.28. making a request to `http://yourdomain.com//evil.com` gives an internal event where the raw path is `//evil.com`. as far as i understand it that means https://github.com/opennextjs/opennextjs-aws/blob/daaf38f45e6aa5596fb4dfdb912aee254fcb94e3/packages/open-next/src/core/routing/middleware.ts#L70 evaluates to `new URL("//evil.com", "http://yourdomain.com//evil.com")` so you're going to a get a middleware invocation where `request.url === "http://evil.com"`.

the PR isn't trying to change any functionality about redirecting within the middleware function, but to avoid calling the middleware entirely with that bad path. so it would go:
1. request to `http://yourdomain.com//evil.com`
2. before the middleware is invoked, redirect to `http://yourdomain.com/evil.com`
3. middleware is invoked like normal

as far as i can tell, this is how the default NextJS server handles this case?

apologies if i've missed something obvious here, this is my first PR and i'm just trying to resolve something that i noticed in our deployment :sweat_smile:

packages/open-next/src/utils/normalize-path.ts
```diff
@@ -1,7 +1,7 @@
 import path from "node:path";
 
 export function normalizePath(path: string) {
-  return path.replace(/\\/g, "/");
+  return path.replace(/\\/g, "/").replace(/\/\/+/g, "/");
```
conico974, line 4: Don't mess with this function, we should use a separate function, just like next does https://github.com/vercel/next.js/blob/3ecf087f10fdfba4426daa02b459387bc9c3c54f/packages/next/src/shared/lib/utils.ts#L348-L360

packages/open-next/src/core/routing/matcher.ts
```diff
@@ -262,6 +263,28 @@ export function handleRewrites<T extends RewriteDefinition>(
   };
 }
 
+function handleRepeatedSlashRedirect(
+  event: InternalEvent,
+): false | InternalResult {
+  // Redirect `https://example.com//foo` to `https://example.com/foo`.
+  const url = new URL(event.url);
+  if (url.pathname.match(/(\\|\/\/)/)) {
```
conico974, line 271: Why not use `event.pathname` here ?

conico974 (commented):
That's basically what Next does
https://github.com/vercel/next.js/blob/3ecf087f10fdfba4426daa02b459387bc9c3c54f/packages/next/src/server/base-server.ts#L1016-L1020

packages/open-next/src/core/routing/matcher.ts
dnewbound0: do you mean `event.rawPath`? i figured they were equivalent but you're right, that does seem cleaner.

packages/open-next/src/core/routing/matcher.ts
```diff
@@ -262,6 +263,25 @@ export function handleRewrites<T extends RewriteDefinition>(
   };
 }
 
+function handleRepeatedSlashRedirect(
```
conico974, line 266: Could you add a comment explaining why we need that. With a permalink to Next https://github.com/vercel/next.js/blob/3ecf087f10fdfba4426daa02b459387bc9c3c54f/packages/next/src/server/base-server.ts#L1016-L1020

packages/open-next/src/utils/normalize-path.ts
```diff
@@ -4,6 +4,13 @@ export function normalizePath(path: string) {
   return path.replace(/\\/g, "/");
 }
 
+export function normalizeRepeatedSlashes(url: URL) {
```
conico974, line 7: Same thing as above, add a link to the function in Next https://github.com/vercel/next.js/blob/3ecf087f10fdfba4426daa02b459387bc9c3c54f/packages/next/src/shared/lib/utils.ts#L348-L360

conico974 (commented):
Just 2 little nit, after that we should be good

conico974 (approved):
Thanks for the PR

PR #864, opened 2025-05-16
Fix NODE_ENV in edge config compilation

conico974:
Set `process.env.NODE_ENV` to production during edge config compilation to avoid defaulting to development.


khuezy (approved):

conico974:
Thanks for the review

PR #867, opened 2025-05-18
fix(converters):  we remove leading forward slashes from url

sommeeeer:
We should remove leading slashes from `req.url` before passing it in to `new URL()` to prevent URL parsing attack. (protocol-relative URL)

### DIFF
--- a/packages/open-next/src/overrides/converters/node.ts
+++ b/packages/open-next/src/overrides/converters/node.ts
@@ -25,7 +25,10 @@ const converter: Converter = {
         ])
         .filter(([key]) => key),
     );
-    const url = new URL(req.url!, `http://${extractHostFromHeaders(headers)}`);
+    const url = new URL(
+      req.url!.replace(/^\/\/+/, "/"),
+      `http://${extractHostFromHeaders(headers)}`,
+    );
     const query = getQueryFromSearchParams(url.searchParams);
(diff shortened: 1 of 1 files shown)


packages/open-next/src/overrides/converters/node.ts
```diff
@@ -25,7 +25,10 @@ const converter: Converter = {
         ])
         .filter(([key]) => key),
     );
-    const url = new URL(req.url!, `http://${extractHostFromHeaders(headers)}`);
+    const url = new URL(
+      req.url!.replace(/^\/\/+/, "/"),
```
conico974, line 29: Use this function instead https://github.com/opennextjs/opennextjs-aws/blob/d434708d07ac297b5317f4ace990836195bd8b3f/packages/open-next/src/utils/normalize-path.ts#L8-L13

sommeeeer: `req.url` is a string, and it could potentially have multiple slashes in the query string aswell.

sommeeeer: `const url = new URL(`http://${extractHostFromHeaders(headers)}${req.url}`);`. Like this perhaps?

conico974 (approved):
LGTM Thanks

PR #868, opened 2025-05-19
Fix decode path params in cache interceptor

conico974:
The SSG cache key used by the incremental cache needs to be decoded.
Otherwise the cache interceptor won't be able to find the cache, and will fallback to `NextServer`


packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -129,6 +129,28 @@ async function generateResult(
...
+ *
+ * SSG cache key needs to be decoded, but some characters needs to be properly escaped
+ * https://github.com/vercel/next.js/blob/34039551d2e5f611c0abde31a197d9985918adaf/packages/next/src/server/lib/router-utils/decode-path-params.ts#L11-L26
+ */
+function decodePathParams(pathname: string): string {
+  return pathname
+    .split("/")
+    .map((segment) => {
+      try {
+        // https://github.com/vercel/next.js/blob/34039551d2e5f611c0abde31a197d9985918adaf/packages/next/src/shared/lib/router/utils/escape-path-delimiters.ts#L2-L10
```
vicb, line 142: nit: what about copy the original `escapePathDelimiters` from next and call from here.

It would be easier to check that the code is in sync / update when needed.

vicb (approved):
LGTM, one nit.

Issue #846, opened 2025-05-01
esbuild Vulnerability and Compatibility Issue


Paulos-se:
I'm encountering a critical security and compatibility issue with OpenNext and esbuild:

1. Security Vulnerability: OpenNext requires esbuild 0.19.2, which has a known security vulnerability ([GHSA-67mh-4wv8-2f99](https://github.com/advisories/GHSA-
67mh-4wv8-2f99)) allowing any website to send requests to the development server and read responses.

2. Compatibility Problem: When attempting to follow the suggestion to "update your version of esbuild" to a secure version (0.25.x), OpenNext 3.6.0 fails with 
the error:
  
   Error: [opennext-edge] "onResolve" filter is not a valid Go regular expression: "(?g)\\.(mjs|wasm)$"
   

3. Regression Issue: OpenNext 3.5.3 works with esbuild 0.25.x override (though still depends on 0.19.2), but 3.6.0 breaks with the same override.

This creates a no-win situation where we must choose between:
• Using a version with a known security vulnerability
• Breaking our build process

Request: Please update the regex pattern in OpenNext to be compatible with newer esbuild versions, allowing users to secure their build environments while continuing to use OpenNext.

khuezy:
See: https://github.com/opennextjs/opennextjs-aws/pull/746

This doesn't affect opennext, since the vulnerability only applies to `esbuild serve/watch`. OpenNext only uses esbuild to build the production bundle. If you are using `esbuild serve/watch` somewhere in your stack, using `pnpm` should allow you to run 2 different versions for each package.

If this vulnerability does impact OpenNext, please provide the details so we can address.

Paulos-se:
While the vulnerability may not directly affect OpenNext's usage of esbuild, the compatibility issue with newer esbuild versions (0.25.x) still exists. The invalid regex pattern (?g)\\.(mjs|wasm)$ prevents users from upgrading esbuild globally in their projects. Could you update this pattern to be compatible with newer esbuild versions?"

khuezy:
`The invalid regex pattern (?g)\.(mjs|wasm)$`

Are you referring to this line?
https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/plugins/edge.ts#L62

The regex looks valid in the source code, how are you getting the `(?g)` in front?

Paulos-se:
[...]

PR #869, opened 2025-05-19
chore: bump esbuild to latest version

sommeeeer:
For #846

I ran into 2 problems after bumping `esbuild`. The first one was a warning during building with OpenNext: 
image

The second problem was that `onResolve` filter error as indicated in the original issue.

To solve them I reordered the `exports` in `package.json`, and I also removed the `/g` at the end from the RegEx filter. This seemed to solve those issues.

I then built and deployed all Next versions from 12-15. I also ran the E2E for Next 14-15. It seems to work fine.

### DIFF
--- a/.changeset/polite-teachers-doubt.md
+++ b/.changeset/polite-teachers-doubt.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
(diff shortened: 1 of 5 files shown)


conico974 (approved):
LGTM
Thank you

Issue #854, opened 2025-05-06
[BUG] Malformed UTF-8 Characters in Query Parameters starting from v3.5.6


lamminsalo:
### Bug Report: Malformed UTF-8 Characters in Query Parameters in v3.5.6

**Description**  
After v3.5.6, UTF-8 characters (e.g., Nordic characters like `äöå`) in query parameters are being malformed into � in deployed environments. This issue only occurs when navigating within the application to a cached page with UTF-8 encoded query parameters. Reverting to v3.5.5 resolves the issue.  

The issue is specific to **deployed environments** and is not reproducible locally, making debugging and reproduction more challenging.

---

**Expected Behavior**  
When navigating to a page with UTF-8 encoded URL parameters, the request should return:  
- **HTTP 304**: "Not Modified" with the query parameter `q=äää`.

Example:  
```bash
GET /search?q=äää
Response: 304 Not Modified
Query: q=äää
```

---

**Actual Behavior**  
Instead of returning the cached page, navigating to a cached page results in a redirect along with malformed query parameters:  
- **HTTP 308**: Permanent redirect with query parameters encoded incorrectly.  

Malformed Example:  
```bash
GET /search?q=äää
Response: 308 Permanent Redirect
To: /_next/data/[hash]/search.json?q=%C3%A4%C3%A4%C3%A4
Actual Query: q=��� (Malformed)
```

---

**Steps to Reproduce**  
1. Deploy the reproduction example linked below.  
2. Navigate to `/search?q=äöå€` using the bottom button (direct navigation from URL works)
3. Observe the malformed query parameters after a redirect.

---

**Minimal Reproduction**  
- Code: https://github.com/lamminsalo/Opennext-Query-Parameter-Minimal-Reproduction
- Live Environment: https://d16ppgwd8iy16k.cloudfront.net/

---

**Discord Discussion**  
For additional context, see the ongoing thread on Discord:  
https://discord.com/channels/1283128968140161065/1286094576788177059/1367919144694972569

---

**Reproduction Environment**  
- **Version:** opennext v3.6.0 

Manually tested that v3.5.6 is the breaking update.

---

_EDIT: Added a more minimal reproduction repo and deployed it. Links and steps updated._

sommeeeer:
[...]

PR #871, opened 2025-05-21
fix: Ensure Location header is properly encoded in redirects

sommeeeer:
For #854

We should ensure that the `Location` header in redirect responses is properly encoded before returning the response to the client. This is for redirects happening from next config.

### DIFF
--- a/examples/app-router/app/config-redirect/dest/page.tsx
+++ b/examples/app-router/app/config-redirect/dest/page.tsx
@@ -1,0 +1,13 @@
+export default async function Page({
+  searchParams,
+}: {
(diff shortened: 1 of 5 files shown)


lamminsalo:
Confirmed it now works as expected:
https://github.com/lamminsalo/Opennext-Query-Parameter-Minimal-Reproduction/commit/82a7f85ef9aa3a8db5cc6b54522d4c162f803e12

https://dv60pajvcna4h.cloudfront.net/

Great job @sommeeeer !

conico974 (approved):
LGTM Thanks

PR #877, opened 2025-05-23
bump turbo to v2

sommeeeer:


sommeeeer:
E2E fails after this bump when I run it locally. Not sure whats causing it yet, but it works fine on turbo v1. Some fails are passing, but many are just timing out with this error: 
image

PR #879, opened 2025-05-25
chore: update contribute readme

sommeeeer:


README.md
```diff
@@ -96,6 +96,8 @@ To run `OpenNext` locally:
    path/to/opennextjs-aws/packages/open-next/dist/index.js build
    ```
 
+There is also a way to run OpenNext locally. You can read a guide about it [here](https://opennext.js.org/aws/contribute/local_run). Be aware that this will not work the same way as a deployed version. Its mostly used to development/debugging purposes.
```
conico974, line 99: > Be aware that this will not work the same way as a deployed version

In what way ? If you say that you have to explain it. But I'm not sure we want 10 lines explaining this here

conico974 (approved):
LGTM Thanks

PR #880, opened 2025-05-25
fix cache control for static page with base path

### LINKED ISSUE #872: Cache control headers not set for static page with base path
My organisation has two projects: one uses `basePath` in `next.config.js` and the other does not. I noticed that for the project using `basePath`, the default cache control header is not showing up for static pages. It shows up for the other project.

I narrowed the header setting down to this check here in `fixCacheHeaderForHtmlPages` [(ref)](https://github.com/opennextjs/opennextjs-aws/blob/ad701c261f0c003eb3cae9cebd688c8408932e84/packages/open-next/src/core/routing/util.ts#L248):

```
  if (
    HtmlPages.includes(localizedPath) &&
    !internalEvent.headers["x-middleware-prefetch"]
  ) {
    headers[CommonHeaders.CACHE_CONTROL] =
      "public, max-age=0, s-maxage=31536000, must-revalidate";
  }
```

If I patch the line to include the base path, then it works
```
HtmlPages.find((path) => `${NextConfig.basePath ?? ""}${path}` === localizedPath)
```

However, I'm not sure if this is the intended method or if there's a canonical way to treat base paths. If I could be pointed in the right direction, I'd be happy to submit a PR for this.

conico974:
Fixes #872

### DIFF
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -245,7 +245,9 @@ export function fixCacheHeaderForHtmlPages(
   // https://opennext.js.org/aws/v2/advanced/workaround#workaround-nextserver-does-not-set-cache-headers-for-html-pages
   // Requests containing an `x-middleware-prefetch` header must not be cached
   if (
-    HtmlPages.includes(localizedPath) &&
+    HtmlPages.find(
+      (path) => `${NextConfig.basePath ?? ""}${path}` === localizedPath,
+    ) &&
     !internalEvent.headers["x-middleware-prefetch"]
   ) {
     headers[CommonHeaders.CACHE_CONTROL] =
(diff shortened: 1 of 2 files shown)


packages/open-next/src/core/routing/util.ts
```diff
@@ -245,7 +245,9 @@ export function fixCacheHeaderForHtmlPages(
   // https://opennext.js.org/aws/v2/advanced/workaround#workaround-nextserver-does-not-set-cache-headers-for-html-pages
   // Requests containing an `x-middleware-prefetch` header must not be cached
   if (
-    HtmlPages.includes(localizedPath) &&
+    HtmlPages.find(
+      (path) => `${NextConfig.basePath ?? ""}${path}` === localizedPath,
```
vicb, line 249: nit: instead of having to string concat time, you could compare to `localizedPathFromBasePath` that would be computed once before the loop

conico974: Ho this made me realize, this might not be correct when both i18n and basePath are set. 
I'll convert this one to draft and make a release in the meantime

Issue #822, opened 2025-04-10
Do not copy unnecessary cache file


conico974:
We should not copy and delete file there https://github.com/opennextjs/opennextjs-aws/blob/73281c958d43d865fdded0d5d86dd82747365fee/packages/open-next/src/build/createAssets.ts#L75-L89

Ideally, we should also add a `revalidate` value (from the prerender manifest) here

PR #883, opened 2025-05-28
fix: move cache assets instead of copying them

makepanic:
fixes parts of #822

I'm not sure if this is the right approach though, as it mutates the `.next` directory by moving files from there directly to the .open-next dir.

From a application with tons of static cache files i get the following difference:

* current version:
  * .next:	2.9 GiB (3,136,622,592 bytes)
  * .open-next:	1.8 GiB (1,918,308,352 bytes)
* with this change
  * .next:	1.6 GiB (1,768,300,544 bytes)
  * .open-next:	1.8 GiB (1,910,587,392 bytes)


conico974:
Thanks for the PR, but i don't think that's what we want to do. 
We rely on a bunch of things from the `.next` folder itself and moving things from there will lead to something that break.
And this will also break some people's deployment, not something we'd want either.

What we should do here instead of copying the file, is to just use the files from `.next` to compute the cache files instead of copying them. This won't reduce the final size that much, but it will reduce the size used during the deployment itself.

conico974 (commented):
Thanks for the PR, but i don't think that's what we want to do.
We rely on a bunch of things from the .next folder itself and moving things from there will lead to something that break.
And this will also break some people's deployment, not something we'd want either.

What we should do here instead of copying the file, is to just use the files from .next to compute the cache files instead of copying them. This won't reduce the final size that much, but it will reduce the size used during the deployment itself.

PR #886, opened 2025-05-30
feat(image-adapter): improve error handling and status codes

### LINKED ISSUE #885: Image Optimization Adapter Returns Improper HTTP Status Codes
### Issue
The OpenNext image optimizer currently returns a 500 error when attempting to access images that are not reachable, regardless of the actual error type. For example, when an image returns a 403 Access Denied error from the source, the image optimizer returns a generic 500 Internal Server Error instead of preserving the original 403 status code.

This behavior makes debugging difficult as all image access errors appear as server errors in monitoring platforms, when in reality they are client-side issues (requesting inaccessible images).

### Solution
The PR below modifies the image optimization adapter to properly handle and preserve HTTP error status codes:

When a remote image server returns a specific HTTP error code (403, 404, etc.), that same status code is preserved and returned to the client
For network errors (ENOTFOUND, ECONNREFUSED), a 404 Not Found status is returned
For other errors, appropriate status codes are determined based on error message content

### Changes
Enhanced error detection in the downloadHandler function to properly handle HTTP status codes from remote servers
Improved error classification in the main handler's catch block to detect different types of errors
Added type-safe error handling with proper TypeScript guards
Improved error messages to be more descriptive

### Benefits
More accurate error reporting in monitoring platforms
Easier debugging of image access issues
Better user experience with appropriate error codes
Follows HTTP standards by preserving original error status codes

### Testing
Tested with images that return various HTTP status codes (403, 404) and verified that the image optimizer correctly preserves these status codes in its response.

iDVB:
Fixes #885

### DIFF
--- a/.changeset/hot-chicken-ring.md
+++ b/.changeset/hot-chicken-ring.md
@@ -1,5 +1,0 @@
----
-"@opennextjs/aws": patch
----
-
-fix cookie decoding for Node and Cloudflare
--- a/.changeset/new-wombats-crash.md
+++ b/.changeset/new-wombats-crash.md
@@ -1,5 +1,0 @@
----
-"@opennextjs/aws": patch
----
-
-perf: drop `babel` to reduce the server bundle size
--- a/examples/app-pages-router/CHANGELOG.md
+++ b/examples/app-pages-router/CHANGELOG.md
@@ -1,5 +1,12 @@
 # app-pages-router
 
+## 0.1.22
+
+### Patch Changes
+
+- Updated dependencies [[`caa57b3ef11f5d2409bf7d64665f8a917bb15486`](https://github.com/opennextjs/opennextjs-aws/commit/caa57b3ef11f5d2409bf7d64665f8a917bb15486), [`de687b29fe9205245dcf23585813cf8faafce5c2`](https://github.com/opennextjs/opennextjs-aws/commit/de687b29fe9205245dcf23585813cf8faafce5c2)]:
+  - @opennextjs/aws@3.6.5
+
 ## 0.1.21
 
 ### Patch Changes
--- a/examples/app-pages-router/package.json
+++ b/examples/app-pages-router/package.json
@@ -1,6 +1,6 @@
 {
   "name": "app-pages-router",
-  "version": "0.1.21",
+  "version": "0.1.22",
   "private": true,
   "scripts": {
     "openbuild": "node ../../packages/open-next/dist/index.js build --build-command \"npx turbo build\"",
--- a/examples/app-router/CHANGELOG.md
+++ b/examples/app-router/CHANGELOG.md
@@ -1,5 +1,12 @@
 # app-router
 
+## 0.1.22
+
+### Patch Changes
+
+- Updated dependencies [[`caa57b3ef11f5d2409bf7d64665f8a917bb15486`](https://github.com/opennextjs/opennextjs-aws/commit/caa57b3ef11f5d2409bf7d64665f8a917bb15486), [`de687b29fe9205245dcf23585813cf8faafce5c2`](https://github.com/opennextjs/opennextjs-aws/commit/de687b29fe9205245dcf23585813cf8faafce5c2)]:
+  - @opennextjs/aws@3.6.5
+
 ## 0.1.21
 
 ### Patch Changes
--- a/examples/app-router/package.json
+++ b/examples/app-router/package.json
@@ -1,6 +1,6 @@
 {
   "name": "app-router",
-  "version": "0.1.21",
+  "version": "0.1.22",
   "private": true,
   "scripts": {
     "openbuild": "node ../../packages/open-next/dist/index.js build --streaming --build-command \"npx turbo build\"",
--- a/packages/open-next/CHANGELOG.md
+++ b/packages/open-next/CHANGELOG.md
@@ -1,5 +1,13 @@
 # open-next
 
+## 3.6.5
+
+### Patch Changes
+
+- [#889](https://github.com/opennextjs/opennextjs-aws/pull/889) [`caa57b3ef11f5d2409bf7d64665f8a917bb15486`](https://github.com/opennextjs/opennextjs-aws/commit/caa57b3ef11f5d2409bf7d64665f8a917bb15486) Thanks [@vicb](https://github.com/vicb)! - fix cookie decoding for Node and Cloudflare
+
+- [#891](https://github.com/opennextjs/opennextjs-aws/pull/891) [`de687b29fe9205245dcf23585813cf8faafce5c2`](https://github.com/opennextjs/opennextjs-aws/commit/de687b29fe9205245dcf23585813cf8faafce5c2) Thanks [@vicb](https://github.com/vicb)! - perf: drop `babel` to reduce the server bundle size
+
 ## 3.6.4
 
 ### Patch Changes
--- a/packages/open-next/package.json
+++ b/packages/open-next/package.json
@@ -3,7 +3,7 @@
     "access": "public"
   },
   "name": "@opennextjs/aws",
-  "version": "3.6.4",
+  "version": "3.6.5",
   "bin": {
     "open-next": "./dist/index.js"
   },


packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -115,9 +115,45 @@ export async function defaultHandler(
...
+    // Check if error has HTTP status information
+    if (e && typeof e === "object") {
+      if ("statusCode" in e && typeof e.statusCode === "number") {
+        statusCode = e.statusCode;
+        errorMessage = `HTTP Error: ${statusCode}`;
+      } else if ("code" in e) {
+        const code = e.code as string;
+        if (code === "ENOTFOUND" || code === "ECONNREFUSED") {
+          statusCode = 404;
+          errorMessage = `Image not found: ${e.message}`;
```
conico974, line 132: We don't want to send the message here. It could contain private/sensitive information

packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -115,9 +115,45 @@ export async function defaultHandler(
...
+        }
+      }
+
+      if (e.message && typeof e.message === "string") {
+        // Try to extract status codes from error messages
+        if (e.message.includes("403") || e.message.includes("Access Denied")) {
+          statusCode = 403;
+          errorMessage = `Access denied: ${e.message}`;
```
conico974, line 140: Same as above

packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -115,9 +115,45 @@ export async function defaultHandler(
...
+        } else if (
+          e.message.includes("404") ||
+          e.message.includes("Not Found")
+        ) {
+          statusCode = 404;
+          errorMessage = `Image not found: ${e.message}`;
```
conico974, line 146: Same as above

packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -115,9 +115,45 @@ export async function defaultHandler(
...
+        } else {
+          errorMessage = e.message;
```
conico974, line 148: Definitely not

conico974 (changes requested):
I have one big issue with this one, it's that in general we don't want to return the error message. Or if we want to do that, it should be under some env variable like `OPEN_NEXT_FORWARD_ERROR_MESSAGE` that would default to false or in debug mode only.

Even forwarding the status code could be bad, we need to look into what `next start` is doing here. We should do the same, and if they don't forward the status code we could put it under an env variable as well

packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -281,6 +375,48 @@ async function downloadHandler(
...
+      // Just pass through the original error to preserve Next.js's error handling
+      // but wrap it in IgnorableError to prevent it from being logged as an error
+      const clientError = new IgnorableError(
+        e.message || "Client error downloading image",
+      );
+
+      // Preserve the original status code or set an appropriate one
+      if (e && typeof e === "object") {
+        if ("statusCode" in e && typeof e.statusCode === "number") {
+          (clientError as any).statusCode = e.statusCode;
```
conico974, line 405: You can modify the `IgnorableError` (and all the other one there) class to include an optional status code

packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -281,6 +375,48 @@ async function downloadHandler(
       }
     }
   } catch (e: any) {
+    // Check if this is a client error (like 404, 403, etc.)
+    const isClientError =
```
conico974, line 379: We probably want a function instead of repeating this twice and it could directly return either an `IgnorableError` (if it's a client error) or a `FatalError` otherwise (which would return a 500 Internal Server Error) and be appropriately logged by the error logger

packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -114,10 +115,56 @@ export async function defaultHandler(
...
+        e.code === "ECONNREFUSED" ||
+        (e.message &&
+          (e.message.includes("403") ||
+            e.message.includes("404") ||
+            e.message.includes("Access Denied") ||
+            e.message.includes("Not Found"))));
+
+    // Only log actual server errors as errors, log client errors as debug
+    if (isClientError) {
+      debug("Client error in image optimization", e);
```
conico974, line 137: Don't use the `debug` logger, just use the `error` with an `IgnorableError`.
The error logger can already properly handle the log level https://github.com/opennextjs/opennextjs-aws/blob/ab492ca53d4494aa81234c7d6c0cdaf42cd05ba6/packages/open-next/src/adapters/logger.ts#L43-L74

packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -255,7 +302,54 @@ async function downloadHandler(
...
+
+      request.on("error", (err: Error & { code?: string }) => {
+        // For network errors, these are typically client errors (bad URL, etc.)
+        // so log as debug instead of error to avoid false alarms
+        const isClientError =
+          err.code === "ENOTFOUND" || err.code === "ECONNREFUSED";
+
+        if (isClientError) {
+          // Log the full error for debugging but don't expose it to the client
+          debug("Client error fetching image", {
```
conico974, line 340: Use the `error` logger here as well

sommeeeer:
https://github.com/opennextjs/opennextjs-aws/blob/6d36c0a59dcf4b45426a4356df190662bf88e5a2/packages/open-next/src/adapters/image-optimization-adapter.ts#L86-L89 - Could you put a `RecoverableError` in this one aswell?

packages/open-next/src/adapters/image-optimization-adapter.ts
iDVB: Addressed in the latest changes.

packages/open-next/src/adapters/image-optimization-adapter.ts
iDVB: Addressed in the latest changes.

packages/open-next/src/adapters/image-optimization-adapter.ts
iDVB: Addressed in the latest changes.

packages/open-next/src/adapters/image-optimization-adapter.ts
iDVB: Addressed in the latest changes.

iDVB:
1. **Error Classes in error.ts**
   - Added statusCode property to BaseOpenNextError interface
   - Added statusCode field to all error classes
   - Updated constructors to accept an optional statusCode parameter with appropriate defaults:
     - IgnorableError: default 400 (client error)
     - RecoverableError: default 400 (client error, but logged as warning)
     - FatalError: default 500 (server error)


2. Update IgnorableError class to include status code
Implemented as suggested
3. Create a function to classify errors instead of repeating logic
Added classifyError() function that centralizes error classification
This function returns the appropriate error type with correct status code
4. Use error logger with proper error types instead of debug logger
Now using error() with appropriate error types that control the log level
Removed debug() calls for client errors
5. For validation errors, use RecoverableError
Image parameter validation errors now use RecoverableError

packages/open-next/src/adapters/image-optimization-adapter.ts
```diff
@@ -126,6 +136,63 @@ export async function defaultHandler(
...
+          e.message.includes("Access Denied") ||
+          e.message.includes("Not Found"))));
+
+  // Determine appropriate status code based on error type
+  if (e && typeof e === "object") {
+    if ("statusCode" in e && typeof e.statusCode === "number") {
+      statusCode = e.statusCode;
+    } else if ("code" in e) {
+      const code = e.code as string;
+      if (code === "ENOTFOUND" || code === "ECONNREFUSED") {
```
khuezy: Isn't `ECONNREFUSED` usually a 500 level error?

iDVB: @khuezy I think you might be missing the original PR context?
All these errors (most) are fundamentally due to the server acting on the client's provided URL.
If a client provides a broken url (eg. invalid url, missing image, access denied, connection refused etc) then thats a "client error" 4xx not a fault of the server 5xx. 

This is indeed how this works with native NextJS image optimizer. `next start`. 
You can test this by going to 
`/_next/image?url=https://someUrl.com`
where `https://someUrl.com` is a url that returns ECONNREFUSED, ENOTFOUNT, ACCESS DENIED etc. 
Open Next server will 500
NextJS (`next start`) will return 4xx with the standard message `"url" parameter is valid but upstream response is invalid`

iDVB: I think an important question might be to what extent does the native NextJS Image Optimizer disregard the error codes and message it gets back from this adapter and simply respond to the client in that generic way.

Then perhaps all we need to do in the adapter is ensure we're not throwing or console.error on anything ... while still returning the accurate error messages? 

This would change how this code works now.... as its trying to modify the error codes and messages to lower 5xx to 4xx so as not to crash / throw the server due to a client bad request (url)

khuezy: gotcha. Very weird that nextjs decided to swallow the 500s into 400. Ooof

iDVB:
I've simplified the classifyError function to focus on the core principle: distinguishing between client errors and server errors for proper logging, while preserving the original error information.

The new implementation:

**Preserves original error details** - We keep the original error message and status code when available
**Minimal classification** - We only classify errors as client (4xx) or server (5xx)
**Passes through to Next.js** - We don't try to normalize error messages or create complex mappings

**This approach has several benefits:**
- Simplicity - The code is more straightforward and easier to maintain
- Less error-prone - Fewer code paths means fewer places for bugs to hide
- Better alignment with Next.js - We let Next.js handle the client response formatting
- The key insight is that our primary goal is to ensure proper error logging and monitoring (avoiding alerts for client errors), while allowing Next.js to handle the actual client-facing error responses. We're not trying to override Next.js's error handling behavior - just making sure our monitoring systems don't get noisy with expected client errors.

packages/open-next/src/adapters/image-optimization-adapter.ts
iDVB: Really? This is what this PR is proposing it should be. 

Otherwise you give access to end users, malicious or otherwise, to cause a 5xx on your servers but simply giving the endpoint a faulty URL or "Bad Request". 

It's no fault of the image optimizer that the end-user passed it a URL to a ZIP file vs an image, or a 404 link, or an access denied one etc etc.

khuezy: You'd have that problem no matter what right? You're suppose to whitelist the endpoints, eg even for server actions and images - `remotePatterns`

iDVB: Next.js's approach of normalizing upstream 5xx errors to 4xx responses is actually quite sensible for this use case. Here's why:

**Monitoring clarity:** By converting upstream 5xx errors to 4xx, Next.js ensures that your monitoring systems only alert on problems you can actually fix, not issues with third-party services.

**Semantic correctness:** From Next.js's perspective, if a user provides a URL that leads to a broken resource, that's fundamentally a client input problem - the client provided a "bad" URL, even if it's "bad" because the upstream server failed.

**Practical operations:** Development and operations teams need clean separation between "our systems are broken" vs. "external dependencies are broken." The conversion to 4xx helps maintain this boundary.

khuezy: `Monitoring clarity` - I think you can argue the opposite too. If Nextjs converts the specific error to a generic, it's hard for you to determine what the problem is.

`Semantic correctness` - Most of the time, the "end user" is not the consumer of the image optimization request - the app is. This is why the `remotePatterns` exists. Your app service is not a service for people to arbitrarily use to optimize their images.

```
Practical operations: Development and operations teams need clean separation between "our systems are broken" vs. "external dependencies are broken." The conversion to 4xx helps maintain this boundary.
```
I'm confused by this statement, it sounds contradicting? If the error is swallowed, how would you know if it's "our system" or "their system" that's broken?

iDVB:
## Core Changes

- **Consolidated error handling**: Refactored into a single [handleImageError](cci:1://file:///Users/dvanbrunt/Projects/opennextjs-aws/packages/open-next/src/adapters/image-optimization-adapter.ts:4:2-73:3) helper function for clarity and reuse
- **Enhanced error detection**: Improved S3 ListBucket permission error detection by checking multiple error paths in AWS SDK v3
- **Preserved status codes**: Maintained appropriate HTTP status codes (403 for access denied) while avoiding server error status codes
- **Secure messaging**: Generic error messages for clients while preserving detailed server-side logging
- **Better monitoring**: Used IgnorableError to classify client errors and prevent false monitoring alerts

## Internal vs External Image Handling

Next.js treats internal and external image errors differently, requiring separate approaches:

- **For internal images (S3)**: 
  - Avoided setting Content-Type to trigger Next.js's "URL valid but internal response invalid" error message
  - Added debugging headers while maintaining secure client responses

- **For external images**:
  - Preserved existing behavior with appropriate text/plain responses
  - Maintained proper error propagation that Next.js expects for external resources

This approach ensures consistent error behavior across image types while providing appropriate client feedback and maintaining security.

PR #887, opened 2025-05-30
fix(converters): Ensure x-forwarded-proto is set to http when running locally

sommeeeer:
Discussed in [#686](https://github.com/opennextjs/opennextjs-cloudflare/issues/686) over at `opennextjs-cloudflare`

`next-auth` and possibly more libraries rely on the `x-forwarded-proto` header to be `http` to work out of the box when run locally. In `next start` or `next dev` they set this to `http` when ran locally. 

To make `opennextjs-cloudflarejs preview` and our node converter when running OpenNext locally behave the same we would want to set this header to `http`. This would make the user not have to set an [`AUTH_URL`](https://authjs.dev/getting-started/deployment#auth_url) in those modes.

### DIFF
--- a/.changeset/fair-pillows-yawn.md
+++ b/.changeset/fair-pillows-yawn.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix(expres-dev): Ensure x-forwarded-proto and \_\_NEXT_PRIVATE_ORIGIN is set
--- a/packages/open-next/src/overrides/wrappers/express-dev.ts
+++ b/packages/open-next/src/overrides/wrappers/express-dev.ts
@@ -34,6 +34,12 @@ const wrapper: WrapperHandler = async (handler, converter) => {
   });
 
   app.all("*paths", async (req, res) => {
+    if (req.protocol === "http" && req.hostname === "localhost") {
+      // This is used internally by Next.js during redirects in server actions. We need to set it to the origin of the request.
+      process.env.__NEXT_PRIVATE_ORIGIN = `${req.protocol}://${req.hostname}`;
+      // This is to make `next-auth` and other libraries that rely on this header to work locally out of the box.
+      req.headers["x-forwarded-proto"] = req.protocol;
+    }
     const internalEvent = await converter.convertFrom(req);
     const streamCreator: StreamCreator = {
       writeHeaders: (prelude) => {


packages/open-next/src/overrides/converters/node.ts
```diff
@@ -30,6 +30,14 @@ const converter: Converter = {
       `${req.protocol ? req.protocol : "http"}://${extractHostFromHeaders(headers)}${req.url}`,
     );
     const query = getQueryFromSearchParams(url.searchParams);
+
+    if (req.protocol === "http" && url.hostname === "localhost") {
```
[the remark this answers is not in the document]
conico974: I don't think we want to do that actually, on docker you may get these values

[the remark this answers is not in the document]
sommeeeer: > Probably not, you are likely inside a cluster behind a CDN or a load balancer.

We do check for the protocol aswell though. 

Then again, I will refactor this so it happens in `express-dev` to avoid any problems with docker.

packages/open-next/src/overrides/converters/edge.ts
```diff
@@ -35,6 +35,12 @@ const converter: Converter<InternalEvent, InternalResult | MiddlewareResult> = {
       ),
     );
 
+    // This is to make libraries (e.g. next-auth) that rely on this header to work out of the box in `opennextjs-cloudflare preview`.
+    // See https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/utils/env.ts#L95-L96
+    if (url.protocol.startsWith("http") && url.hostname === "localhost") {
```
vicb: Maybe something like:

```ts
if (url.hostname === "localhost") {
  headers["x-forwarded-proto"] = url.protocol;
}
```

Because `wrangler dev --local-protocol=https`

sommeeeer: Yeah alright. I was trying to decide if this should go in the converter or wrapper? WDYT?

vicb: QQ how is this header passed to https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/utils/env.ts#L95-L96

sommeeeer: https://github.com/nextauthjs/next-auth/blob/4e26fd01a5286f88d84d79fce16f346592ef8b8f/packages/next-auth/src/lib/actions.ts#L26-L34

They use `{ headers} ` from `next/headers` inside an action.

sommeeeer: This is again another case where the protocol would default to `https` if we don't have a `x-forwarded-proto` header set.

sommeeeer:
I have to dig a bit further on this one. Turns out you might need an `AUTH_URL` after all.

packages/open-next/src/overrides/converters/edge.ts
```diff
@@ -35,6 +35,11 @@ const converter: Converter<InternalEvent, InternalResult | MiddlewareResult> = {
       ),
     );
 
+    // This is to make libraries (i.e next-auth) that rely on this header to work out of the box in `opennextjs-cloudflare preview`.
+    if (url.protocol.slice(0, -1) === "http" && url.hostname === "localhost") {
```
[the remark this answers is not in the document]
sommeeeer: I think this one would not be reliable. If its `https:`, that condition would pass.

sommeeeer:
Im using [this](https://github.com/sommeeeer/next-auth-test) repo to test out. First build the repo with `pnpm next build && pnpm next start` and you can see it works fine. 
Username: `john@opennext.com`
Password: `password`

However when in `wrangler dev` (or running express-dev/node) for some reason the `callbackUrl` [here](https://github.com/nextauthjs/next-auth/blob/4e26fd01a5286f88d84d79fce16f346592ef8b8f/packages/core/src/lib/pages/signin.tsx#L177-L201) will have `https` even when I set this in the edge converter: 
```ts
// this is for the redirects to work with http
process.env.__NEXT_PRIVATE_ORIGIN = url.origin;
// this makes the signin button redirect to http
headers["x-forwarded-proto"] = "http";
```

Remember to change the overrides path in `package.json`. After you have added those things above in the edge converter in your `opennextjs-aws` repo and built it, go and build the `next-auth-test` project with `pnpm install && pnpm preview -- --port 3000`. If you click on the `Sign in with Credentials` button, you will get sent to a page with this form:
image

Notice the `https` and not `http` like in `next start`. Thats make this form action not work out of the box. You can set an `AUTH_URL` and it will work, but im trying to get this to work like with `next start`. Still digging to figure out whats causing this.

sommeeeer:
Alright, this one was a bit harder to debug than I initially thought. 

Turns out this is related to `request.url` being wrong in API routes when running OpenNext locally. The `url` is `https....` even though you are running locally with `http`.

`next-auth` uses this `url` from the `request` object in its API handlers on `/app/api/auth/[...nextauth]/route.ts`. 

There is some workarounds (`AUTH_URL`, or creating your custom signin page), but im not sure whats causing this yet, or if its even possible to fix. I checked and the `req.url` is correct here: https://github.com/opennextjs/opennextjs-aws/blob/68bb9b0e6b45b3a6c67081529c0b1074520d14be/packages/open-next/src/core/requestHandler.ts#L259

To see this yourself you can just add an API route in your app and run OpenNext locally either via `opennextjs-cloudflare preview` or `node .open-next/server-functions/default/index.mjs`:
```ts
import { NextRequest } from "next/server";

export async function GET(request: NextRequest) {
  return Response.json({
    // when running OpenNext locally this url with https when it should not
    url: request.url,
  })
}
```

sommeeeer:
I removed the `edge` part as this should be handled by `wrangler` and not us. However, the changes in `express-dev` will make redirects from server actions work and also make libraries that rely on the `x-forwarded-proto` header work out of the box when running OpenNext locally.

conico974 (approved):

Issue #888, opened 2025-05-28
[BUG] Cookies matcher doesn't work properly (cookies wrongly parsed it seems)


lauri865:
### Describe the bug

We have a missing cookie rewrite which doesn't work. After some debugging, it seems like the issue is in cookies being wrongly parsed.

Cookies passed to the matcher:
`cookies { _ga: 'xxx; codeVariant' }`

Cookies from the headers:
`cookie: 'xxx; codeVariant=TS; codeStyling=none; _ga_XJ83JQEK7J=xxx; crisp-client%2Fsession%2Fd4cd5bac-6848-4f4d-9a7f-082f5b324b79=xxx; _octo=xxx; sidebar:state=false; cpu_bucket=xxx; preferred_color_mode=xxx; tz=xxx; test=xxx'`

This implementation is wrong: https://github.com/opennextjs/opennextjs-aws/blob/ab492ca53d4494aa81234c7d6c0cdaf42cd05ba6/packages/open-next/src/http/util.ts#L39

Cookies are semicolon-separated, not comma-separated.

### Steps to reproduce

1. Add a rewrite to next.config:
`async rewrites() {
    return {
      beforeFiles: [
        {
          source: "/test",
          destination: "https://github.com",
          missing: [
            {
              type: "cookie",
              key: "test",
            },
          ],
        },
      ],
    };
  },`

2. Add a cookie `test` with any value.

3. The rewrite will always pass.

### Expected behavior

Rewrite should 

### @opennextjs/cloudflare version

1.0.4

### Wrangler version

4.17.0

### next info output

```shell
N/A
```

vicb:
Moving the issue to the aws repo where the issue should be fixed.

I think we have multiple issues here: [`parseCookies`](https://github.com/opennextjs/opennextjs-aws/blob/ab492ca53d4494aa81234c7d6c0cdaf42cd05ba6/packages/open-next/src/http/util.ts#L39) parses Response cookies, i.e. the `Set-Cookie` header. We should probably renamed that function.

Request cookies are in the form `key=value; ...`.

So I believe we should not use `parseCookies` in:

- packages/open-next/src/overrides/converters/node.ts
- convertFrom in packages/open-next/src/overrides/converters/edge.ts

We should double check all the usage of `parseCookies`

khuezy:
[...]

PR #889, opened 2025-06-02
fix: cookie decoding for Node and Cloudflare

vicb:
fixes https://github.com/opennextjs/opennextjs-aws/issues/888

### DIFF
--- a/packages/open-next/src/http/util.ts
+++ b/packages/open-next/src/http/util.ts
@@ -41,9 +41,14 @@ export function parseSetCookieHeader(
     return [];
   }
 
-  return typeof cookies === "string"
-    ? cookies.split(/(?<!Expires=\w+),/i).map((c) => c.trim())
-    : cookies;
+  if (typeof cookies === "string") {
+    // Split the cookie string on ",".
+    // Note that "," can also appear in the Expires value (i.e. `Expires=Thu, 01 June`)
+    // so we have to skip it with a negative lookbehind.
+    return cookies.split(/(?<!Expires=\w+),/i).map((c) => c.trim());
+  }
+
+  return cookies;
 }
 
 /**


packages/open-next/src/http/util.ts
```diff
@@ -28,7 +28,13 @@ export const convertHeader = (header: http.OutgoingHttpHeader) => {
   return String(header);
 };
 
-export function parseCookies(
+/**
+ * Parses a (coma-separated) list of Set-Cookie headers
```
sommeeeer: comma

packages/open-next/src/http/util.ts
```diff
@@ -28,7 +28,13 @@ export const convertHeader = (header: http.OutgoingHttpHeader) => {
...
+ *
+ * @param cookies A coma-separated list of Set-Cookie headers or a list of Set-Cookie header
+ * @returns A list of Set-Cookie header
```
vicb:
```suggestion
 * Parses a (comma-separated) list of Set-Cookie headers
 *
 * @param cookies A comma-separated list of Set-Cookie headers or a list of Set-Cookie header
 * @returns A list of Set-Cookie header
```

packages/open-next/src/http/util.ts
vicb: Thanks!

packages/open-next/src/http/util.ts
```diff
@@ -28,7 +28,13 @@ export const convertHeader = (header: http.OutgoingHttpHeader) => {
   return String(header);
 };
 
-export function parseCookies(
+/**
+ * Parses a (comma-separated) list of Set-Cookie headers
+ *
+ * @param cookies A comma-separated list of Set-Cookie headers or a list of Set-Cookie header
```
vicb:
```suggestion
 * @param cookies A comma-separated list of Set-Cookie headers or a list of Set-Cookie headers
```

sommeeeer:
LGTM

Thanks for doing this. We can consider adding an unit test for the edge converter at a later time.

vicb:
> Thanks for doing this. We can consider adding an unit test for the edge converter at a later time.

I'll look into adding an e2e...
_(there were unit tests for the Node impl)_

packages/open-next/src/http/util.ts
```diff
@@ -28,7 +28,13 @@ export const convertHeader = (header: http.OutgoingHttpHeader) => {
...
 };
 
-export function parseCookies(
+/**
+ * Parses a (comma-separated) list of Set-Cookie headers
+ *
+ * @param cookies A comma-separated list of Set-Cookie headers or a list of Set-Cookie headers
+ * @returns A list of Set-Cookie header
+ */
+export function parseSetCookieHeader(
```
khuezy, line 37: 👍 In hindsight, there should have been a comment on the "Expires-" regex, it can be cryptic if you don't have the context.

TL;DR: it splits each cookie on the "," but not when the comma belongs to the `Expires=Thu, 01 June ....` value.

vicb: I wanted to add that but forgot, thanks for the reminder :)

vicb:
Thanks for the reviews!

PR #892, opened 2025-06-05
refactor: patch NextServer#getMiddlewareManifest regardless of Next version

vicb:


sommeeeer:
I think you can drop that `minimalRule` aswell? Since `runEdgeFunction` and `runMiddleware` is emptied in the `dropBabel` patch? You could most likely do the same with the `imageOptimizer` there to be consistent.

vicb:
> I think you can drop that `minimalRule` aswell? Since `runEdgeFunction` and `runMiddleware` is emptied in the `dropBabel` patch? You could most likely do the same with the `imageOptimizer` there to be consistent.

Good point.

I'll create a distinct PR for that for tracking purposes

sommeeeer:
What about `hasMiddleware` and `loadNodeMiddleware`?

vicb:
> What about `hasMiddleware` and `loadNodeMiddleware`?

It doesn't look like they pulls deps?
So while we could probably empty their body, there wouldn't be much gain?

packages/open-next/src/build/patch/patches/patchNextServer.ts
```diff
@@ -78,16 +78,15 @@ export const patchNextServer: CodePatcher = {
...
       field: {
-        pathFilter: /next-server\.(js)$/,
+        pathFilter: /next-server\.js$/,
         contentFilter: /this\.nextConfig\.experimental\.preloadEntriesOnStart/,
         patchCode: createPatchCode(disablePreloadingRule),
       },
     },
     // Don't match edge functions in `NextServer`
     {
-      versions: ">=15.0.0",
```
conico974, line 88: We should probably set it to above 14, because next 12 and some version of 13 actually use the bundled middleware/edge function

vicb: I'll tackle that in another follow up PR.
I also want to do that for the preloading (>=14) as it will simplify our life with the cloudflare adapter.

vicb: OOps sorry, read it too quickly. I'll update.

vicb:
Thanks Nico!

PR #896, opened 2025-06-09
Disable background route preloading for Next >= 14

vicb:
~Will test with the cloudflare adapter before removing draft state~

Added a test and made sure it works with the cloudflare adapter


vicb:
Thanks for the review!

PR #899, opened 2025-06-10
refactor: Use diff in patch tests rather than raw output

vicb:
IMO it makes the tests much more readable

### DIFF
--- a/packages/open-next/package.json
+++ b/packages/open-next/package.json
@@ -59,7 +59,6 @@
     "@types/aws-lambda": "^8.10.109",
     "@types/express": "5.0.0",
     "@types/node": "catalog:",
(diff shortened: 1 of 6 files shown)


packages/open-next/package.json
```diff
@@ -59,6 +59,7 @@
     "@types/aws-lambda": "^8.10.109",
     "@types/express": "5.0.0",
     "@types/node": "catalog:",
+    "diff": "^8.0.2",
```
conico974: Shouldn't this be in the `test-unit` package instead ?

vicb: Good catch (mixed with the cloudflare adapter where the tests are alongside the source).
Thanks!

Issue #903, opened 2025-06-12
next.js optimizeCss feature doesn't work since v3


artabr:
Dear Team,

please consider looking into this specific issue.

The inlining of Critical CSS with the optimizeCss feature doesn't work when using Next.js 15.3.3 (Page Router) and @opennextjs/aws 3.6.5.

Although this might not affect a big amount of users, I found this feature very beneficial in terms of SEO and overall web performance, that's why I would like to bring your attention to this.

> Below, I attached examples to illustrate the problem.
> Disclaimer. Although, it uses Netlify adapter the issue in @opennextjs/aws is the same.
> In production environment we experience this with the @opennextjs/aws adapter - I just created minimal reproduction on Netlify for the sake of speed. But the problem stays the same with any adapter - no css files in the bundle, no inlined styles. Hope you will understand and forgive me for this little inconsistency with the reproduction 🙏.

**❌ Actual Behavior:**
In Next.js 15.3.3 with @netlify/plugin-nextjs 5.11.2:

Next.js using optimizeCss feature with Page Router:

```
const nextConfig = {
  experimental: {
    optimizeCss: true,
  },
};
```

Critical CSS is not inlined into the HTML.
Pages lack critical CSS styling, which can negatively impact SEO and performance metrics.

![Image](https://github.com/user-attachments/assets/aff5916d-94d2-4c13-90ee-42bbb89296a1)

**✅ Expected Behavior:**
When the optimizeCss feature is enabled Next.js should inline critical CSS files into the HTML.

![Image](https://github.com/user-attachments/assets/eea1f3c0-d040-4669-baaa-44ee4d8c1b02)

**Steps to Reproduce**
A reproduction example has been set up on Netlify:

**Broken Example:**

URL: https://v15-3-3--optimize-css.netlify.app/
GitHub: https://github.com/artabr/optimize-css-repro/tree/v15-3-3
Setup: Next.js 15.3.3 with the original Netlify adapter.
Outcome: No critical CSS inlined.

**Patched Example (Working Version):**

URL: https://v15-3-3-plugin-patched--optimize-css.netlify.app/
GitHub: https://github.com/artabr/optimize-css-repro/tree/v15-3-3-plugin-patched
Setup: Patched Netlify adapter to explicitly include .next/static/css in the server-function bundle.
Outcome: Critical CSS is correctly inlined.

**Root Cause Analysis**
The issue appears to be related to how @netlify/plugin-nextjs bundles assets:

[...]

PR #904, opened 2025-06-13
feat: add css files to server bundle for optimizeCss feature to work

artabr:
## Description

The inlining of Critical CSS with the optimizeCss feature doesn't work when using Next.js 15.3.3 (Page Router) and @opennextjs/aws 3.6.5.

The optimizeCss feature relies on the CSS files being present in the server-function bundle. During page rendering (e.g., during ISR or revalidation), a server function using the Critters library attempts to locate the CSS files used on the page and inline them in the HTML. Without these files in the server-function bundle, the inlining process fails.

These changes make the bundling process include `.next/static/css` files in the server-function bundle.

Detailed information is in the attached issue below.

## GitHub issue

https://github.com/opennextjs/opennextjs-aws/issues/903

### DIFF
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -2,6 +2,7 @@ import url from "node:url";
 
 import {
   copyFileSync,
+  cpSync,
   existsSync,
   mkdirSync,
   readFileSync,
@@ -359,6 +360,14 @@ File ${serverPath} does not exist
       .filter((file) => file.endsWith(".html"))
       .forEach((file) => copyStaticFile(`server/${file}`));
   }
+
+  // Copy .next/static/css from standalone to output dir
+  // needed for optimizeCss feature to work
+  cpSync(
+    path.join(standaloneNextDir, "static", "css"),
+    path.join(outputNextDir, "static", "css"),
+    { recursive: true },
+  );
 
   logger.debug("copyTracedFiles:", Date.now() - tsStart, "ms");
 
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -361,19 +361,25 @@ File ${serverPath} does not exist
       .forEach((file) => copyStaticFile(`server/${file}`));
   }
 
+  const manifests = getManifests(standaloneNextDir);
+
+  const { optimizeCss } = manifests.config.experimental;
+
   // Copy .next/static/css from standalone to output dir
   // needed for optimizeCss feature to work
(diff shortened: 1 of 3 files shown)


packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -360,6 +361,14 @@ File ${serverPath} does not exist
       .forEach((file) => copyStaticFile(`server/${file}`));
   }
 
+  // Copy .next/static/css from standalone to output dir
+  // needed for optimizeCss feature to work
+  cpSync(
```
conico974: We should only copy this if `optimizeCss` is true. We don't want to add unnecessary files to the bundle

artabr: Got you, I will make changes then

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -361,19 +361,25 @@ File ${serverPath} does not exist
       .forEach((file) => copyStaticFile(`server/${file}`));
   }
 
+  const manifests = getManifests(standaloneNextDir);
```
sommeeeer: We do `loadConfig` just above, so you can just use that instead. Not sure what people would prefer?

artabr: `loadConfig` is also called inside `getManifests` - I just wanted to reuse and not to do additional calls. I could move `manifests` even above  `if (hasPageDir) {` and reuse manifests.config inside that block as well, what do you think about it?

sommeeeer: Ah hold on, I was wrong. The `loadConfig` I was looking at was in a conditional block. I think how you are doing it is fine! Thanks for the PR.

sommeeeer: That sounds like a good idea. That way we only call it once, even if we need it inside `if (hasPageDir)....`. Not that it matters much, but both `loadPrerenderManifest` and `loadConfig` can then be replaced inside that conditional.

artabr:
Hi, team, I added the changes we discussed here, and tested them on our infrastructure, works as expected. Would you be able to take a look once again at this PR?

We'll be looking forward to your comments

sommeeeer:
Could you add a changeset? `npx changeset` from root.

artabr:
@sommeeeer sure, changeset added

khuezy (approved):
LGTM Thanks!

conico974 (approved):
LGTM Thanks

artabr:
Thank you team for all your reviews and comments!

PR #906, opened 2025-06-17
fix(edge-converter): Ensure null body status codes dont return a body

sommeeeer:
For https://github.com/opennextjs/opennextjs-cloudflare/issues/684

We need to ensure that the response doesn't return a body if its a status code that does not allow it. (101, 204, 205, 304).

[Here](https://github.com/sommeeeer/opennext-rewrite-null-body) is a reproduction I made for that issue. To run the reproduction just `pnpm install && pnpm preview`. Then in another terminal you `curl -i http://localhost:8787/src2`.

We could fix this inside the `overrides/proxyExternalRequest/fetch.ts` aswell. We would just return the body as is, and not fallback to `emptyReadableStream()`. Not sure what would be preferred? I dont think `emptyReadableStream()` has any effect in this converter.

### DIFF
--- a/packages/open-next/src/overrides/converters/edge.ts
+++ b/packages/open-next/src/overrides/converters/edge.ts
@@ -15,6 +15,7 @@ declare global {
   var __dangerous_ON_edge_converter_returns_request: boolean | undefined;
 }
 
+// https://fetch.spec.whatwg.org/#statuses
 const NULL_BODY_STATUSES = new Set([101, 103, 204, 205, 304]);
 
 const converter: Converter<InternalEvent, InternalResult | MiddlewareResult> = {


packages/open-next/src/overrides/converters/edge.ts
```diff
@@ -15,6 +15,8 @@ declare global {
   var __dangerous_ON_edge_converter_returns_request: boolean | undefined;
 }
 
+const NULL_BODY_STATUSES = new Set([101, 204, 205, 304]);
```
conico974: https://fetch.spec.whatwg.org/#statuses
```suggestion
const NULL_BODY_STATUSES = new Set([101, 103, 204, 205, 304]);
```

conico974 (commented):
Just one little change and it should be good

PR #907, opened 2025-06-18
Fix incorrect status code being sent from interceptor

### LINKED ISSUE #901: Wrong error code in cache interceptor
Cache interceptor for other value than a 200 return the wrong status code

conico974:
Fix #901


khuezy (approved):
LGTM

PR #911, opened 2025-06-21
Optimize tag writing process

conico974:
Enhance the tag caching mechanism by ensuring tags are written are written only once. Before that same tags were written both for the composable cache and the incremental cache.

Improve the composable cache read by reading from the pending set promise first.

### DIFF
--- a/packages/open-next/src/utils/cache.ts
+++ b/packages/open-next/src/utils/cache.ts
@@ -1,4 +1,8 @@
-import type { CacheValue, WithLastModified } from "types/overrides";
+import type {
+  CacheValue,
+  OriginalTagCacheWriteInput,
+  WithLastModified,
+} from "types/overrides";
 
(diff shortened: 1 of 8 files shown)


packages/open-next/src/utils/cache.ts
```diff
@@ -39,3 +43,39 @@ export function getTagsFromValue(value?: CacheValue<"cache">) {
...
+    tag: tag.tag,
+    path: tag.path,
+  });
+}
+
+export async function writeTags(
+  tags: (string | OriginalTagCacheWriteInput)[],
+): Promise<void> {
+  const store = globalThis.__openNextAls.getStore();
+  console.log("Writing tags", tags, store);
```
sommeeeer: debug?

vicb: use the logger?

vicb (approved):

PR #912, opened 2025-06-23
chore(image): add content-security-policy and content-disposition

sommeeeer:
Add `Content-Security-Policy` and `Content-Disposition` header to `buildSuccessResponse` in image optimization adapter. 

It will now support `images.contentSecurityPolicy` and `images.contentDispositionType` from Next config aswell.


conico974 (changes requested):
We don't want to import things from Next for that

PR #914, opened 2025-06-23
Add request ID handling

### LINKED ISSUE #913: Add an option to set the request id
We should add an option to set the request id in the header even if we are not in debug mode.
This requestId should also stay the same if you use an external middleware, both in the server and the middleware

conico974:
Introduce an environment variable `OPEN_NEXT_REQUEST_ID_HEADER` to append the request ID to headers.

The requestId is also now shared between the middleware and the server in case of an external middleware

Close #913

### DIFF
--- a/packages/open-next/src/adapters/edge-adapter.ts
+++ b/packages/open-next/src/adapters/edge-adapter.ts
@@ -9,6 +9,7 @@ import type { OpenNextHandlerOptions } from "types/overrides";
 import { NextConfig } from "../adapters/config";
 import { createGenericHandler } from "../core/createGenericHandler";
 import { convertBodyToReadableStream } from "../core/routing/util";
+import { INTERNAL_EVENT_REQUEST_ID } from "../core/routingHandler";
 
 globalThis.__openNextAls = new AsyncLocalStorage();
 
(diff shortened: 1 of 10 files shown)


packages/open-next/src/adapters/edge-adapter.ts
```diff
@@ -18,9 +19,13 @@ const defaultHandler = async (
 ): Promise<InternalResult> => {
   globalThis.isEdgeRuntime = true;
 
+  const requestId = globalThis.openNextConfig.middleware?.external
```
khuezy, line 22: sorry I'm missing some context here. How come it's using the internal req id for external middleware and a random string if it's internal?

sommeeeer: If you have an external middleware (i.e set in your OpenNext config) the `requestId` gets computed there first and passed as a internal header to your request handler. If you dont have an external middleware it will default to a random string.

khuezy: thx. Is there some kind of existing aws context request id we can use instead of generating a new random one?

conico974: We could, but we cannot know where we are running. It could be on lambda, ecs or cloudflare.

We could add an override to provide the request id.
I'm planning to improve the default logger with an override at some point, it make sense that we do it there.

conico974:
Thanks for the review

PR #916, opened 2025-06-24
fix: add Vary header to cache interceptor

sommeeeer:
Closes https://github.com/opennextjs/opennextjs-cloudflare/issues/750

There is some issue with Firefox and possibly some CDNs that they will incorrectly be reusing RSC responses during bfcache restores. Adding a `Vary` header seems to solve the problem.

### DIFF
--- a/packages/open-next/src/core/routing/cacheInterceptor.ts
+++ b/packages/open-next/src/core/routing/cacheInterceptor.ts
@@ -23,7 +23,7 @@ const CACHE_ONE_MONTH = 60 * 60 * 24 * 30;
  * Also see this PR: https://github.com/vercel/next.js/pull/79426
  */
 const VARY_HEADER =
-  "RSC, Next-Router-State-Tree, Next-Router-Prefetch, Next-Router-Segment-Prefetch";
+  "RSC, Next-Router-State-Tree, Next-Router-Prefetch, Next-Router-Segment-Prefetch, Next-Url";
 
 async function computeCacheControl(
   path: string,


packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -13,6 +13,9 @@ import { generateMessageGroupId } from "./queue";
 const CACHE_ONE_YEAR = 60 * 60 * 24 * 365;
 const CACHE_ONE_MONTH = 60 * 60 * 24 * 30;
 
+const VARY_HEADER =
```
vicb, line 25: Can you explain (in a comment) why this value?

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -13,6 +13,17 @@ import { generateMessageGroupId } from "./queue";
 const CACHE_ONE_YEAR = 60 * 60 * 24 * 365;
 const CACHE_ONE_MONTH = 60 * 60 * 24 * 30;
 
+/*
+ * We use this header to prevent Firefox (and possibly some CDNs) from incorrectly reusing the RSC responses during caching.
+ * This can especially happen when theres a redirect in the middleware as the `_rsc` query parameter is not visible there.
```
khuezy:
```suggestion
 * This can especially happen when there's a redirect in the middleware as the `_rsc` query parameter is not visible there.
```

sommeeeer:
Thanks for the reviews @vicb and @khuezy

khuezy (approved):
LGTM. I thought I saw something about nextjs putting the vary header back in the newer release? Do we need to patch it on a specific version?

sommeeeer:
> I thought I saw something about nextjs putting the vary header back in the newer release? Do we need to patch it on a specific version?

Yeah they did, but they added it back in version `15.3.3`. I guess theres no point making a patch for NextServer, as we can probably just tell the users to upgrade to that version or higher.

Edit: This is also a really hard edge case to hit only in Firefox, and it does require you to have a redirect in the middleware to a route with RSC. Firefox caches the `_rsc` response instead of the actual page then.

vicb:
> > I thought I saw something about nextjs putting the vary header back in the newer release? Do we need to patch it on a specific version?
> 
> Yeah they did, but they added it back in version `15.3.3`. I guess theres no point making a patch for NextServer, as we can probably just tell the users to upgrade to that version or higher.
> 
> Edit: This is also a really hard edge case to hit only in Firefox, and it does require you to have a redirect in the middleware to a route with RSC. Firefox caches the `_rsc` response instead of the actual page then.

The reason why I did not formally approve is that I have no clue if the value for the vary header will fix all cases.
If Next uses the same value, maybe you could add a link to the PR/code in the comment?

khuezy:
We could expand the playwright tests to include the other browsers too.

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -13,6 +13,18 @@ import { generateMessageGroupId } from "./queue";
...
+ * We use this header to prevent Firefox (and possibly some CDNs) from incorrectly reusing the RSC responses during caching.
+ * This can especially happen when there's a redirect in the middleware as the `_rsc` query parameter is not visible there.
+ * So it will get dropped during the redirect, which results in the RSC response being cached instead of the actual HTML on the path `/`.
+ * This value can be found in the routes manifest, under `rsc.varyHeader`.
+ * They recompute it here in Next:
+ * https://github.com/vercel/next.js/blob/c5bf5bb4c8b01b1befbbfa7ad97a97476ee9d0d7/packages/next/src/server/base-server.ts#L2011
+ * Also see this PR: https://github.com/vercel/next.js/pull/79426
+ */
+const VARY_HEADER =
+  "RSC, Next-Router-State-Tree, Next-Router-Prefetch, Next-Router-Segment-Prefetch";
```
conico974: We should add the Next-url header

conico974 (approved):

PR #917, opened 2025-06-26
Add an asset resolver override

vicb:


packages/open-next/src/core/routing/matcher.ts
```diff
@@ -170,6 +170,11 @@ export function getNextConfigHeaders(
   return requestHeaders;
 }
 
+/**
+ * TODO: This method currently only check for the first match.
+ *       It should check for all matches for `beforeFiles` and `afterFiles` rewrite
```
conico974, line 175: Is it for all or only for `beforeFiles` ? The docs talks about `beforeFiles` only for that.
We need to look at how this is done in Next, but yeah definitely needs to be fixed

packages/open-next/src/core/routingHandler.ts
```diff
@@ -105,70 +117,90 @@ export default async function routingHandler(
...
+        eventOrResult,
         RoutesManifest.rewrites.afterFiles,
       );
-      internalEvent = afterRewrites.internalEvent;
-      isExternalRewrite = afterRewrites.isExternalRewrite;
+      eventOrResult = afterRewrite.internalEvent;
+      isExternalRewrite = afterRewrite.isExternalRewrite;
+      // Check for matching public files after `afterFiles` rewrites
+      if (!isExternalRewrite) {
+        const assetResult =
```
conico974: We should not look for assets here, it should only be done before that https://nextjs.org/docs/app/api-reference/file-conventions/middleware#execution-order

conico974: Same thing for all the one after that

conico974 (commented):
I think we don't have a choice, but to check every single static assets in `maybeGetAssetResult`

packages/open-next/src/core/routing/matcher.ts
vicb: It is a few lines appart but

> Good to know: rewrites in beforeFiles do not check the filesystem/dynamic routes immediately after matching a source, they continue until all beforeFiles have been checked.

and

> afterFiles rewrites are checked/applied, if one of these rewrites is matched we check dynamic routes/static files after each match

packages/open-next/src/core/routingHandler.ts
vicb: https://nextjs.org/docs/app/api-reference/config/next-config-js/rewrites says something different:

image

#5 seems to indicate that we need to check here - at least if `__rewrite` is defined (i.e. there is a match)

> Same thing for all the one after that

That sounds right.

But we probably need to check the implementation as the 2 doc pages say something different

conico974: Ho ok, i know where the confusion is coming from. When they talk about static files here they refer to static SSG/ISR route (we should verify that, but if i recall correctly, that's the case).
What we call assets are only checked at `4.`

vicb:
@conico974 I think I have addressed all the feedback, could you please take another look?

conico974 (approved):
LGTM Thanks!

vicb:
Thanks for the review and feedback Nico!

PR #920, opened 2025-07-02
Implement persistent data cache option and update cache key logic

conico974:
Introduce a new option to maintain the data cache across deployments and adjust cache key generation to reflect this change. Fix issues related to tag caching and ensure proper handling of cache keys in various scenarios.

BREAKING CHANGE: Incremental cache keys are now an object of type `CacheKey` instead of a string. The new type includes properties like `baseKey`, `buildId`, and `cacheType`. Build_id is automatically provided according to the cache type and the `dangerous.persistentDataCache` option. Up to the Incremental Cache implementation to use it as they see fit.

### DIFF
--- a/packages/tests-unit/tests/adapters/cache.test.ts
+++ b/packages/tests-unit/tests/adapters/cache.test.ts
@@ -562,13 +562,22 @@ describe("CacheHandler", () => {
       globalThis.tagCache.getByTag.mockResolvedValueOnce(["/path"]);
       await cache.revalidateTag("tag");
 
-      expect(globalThis.tagCache.getByTag).toHaveBeenCalledWith("tag");
+      expect(globalThis.tagCache.getByTag).toHaveBeenCalledWith({
+        baseKey: "tag",
+        buildId: "undefined-build-id",
+      });
 
       expect(tagCache.writeTags).toHaveBeenCalledTimes(1);
       expect(tagCache.writeTags).toHaveBeenCalledWith([
         {
-          path: "/path",
-          tag: "tag",
+          path: {
+            baseKey: "/path",
+            buildId: "undefined-build-id",
+          },
+          tag: {
+            baseKey: "tag",
+            buildId: "undefined-build-id",
+          },
         },
       ]);
     });
@@ -581,8 +590,14 @@ describe("CacheHandler", () => {
       expect(tagCache.writeTags).toHaveBeenCalledTimes(1);
       expect(tagCache.writeTags).toHaveBeenCalledWith([
         {
-          path: "/path",
-          tag: "_N_T_/path",
+          path: {
+            baseKey: "/path",
+            buildId: "undefined-build-id",
+          },
+          tag: {
+            baseKey: "_N_T_/path",
+            buildId: "undefined-build-id",
+          },
         },
       ]);
 
@@ -596,8 +611,14 @@ describe("CacheHandler", () => {
       expect(tagCache.writeTags).toHaveBeenCalledTimes(1);
       expect(tagCache.writeTags).toHaveBeenCalledWith([
         {
-          path: "123456",
-          tag: "tag",
+          path: {
+            baseKey: "123456",
+            buildId: "undefined-build-id",
+          },
+          tag: {
+            baseKey: "tag",
+            buildId: "undefined-build-id",
+          },
         },
       ]);
 


vicb:
Is the fix related to the other changes? (Maybe clarify in the PR description) 

There are conflicts. 

I'll take a look later today

conico974:
My bad i rebased on the wrong branch... I'll fix it

packages/open-next/src/types/open-next.ts
```diff
@@ -78,6 +78,16 @@ export interface DangerousOptions {
   headersAndCookiesPriority?: (
     event: InternalEvent,
   ) => "middleware" | "handler";
+
+  /**
+   * Persist data cache between deployments.
+   * Next.js claims that the data cache is persistent (not true for `use cache` and it depends on how you build/deploy otherwise).
+   * By default, every entry will be prepended with the BUILD_ID, when enabled it will not.
+   * This means that the data cache will be persistent between deployments.
+   * This is useful in a lot of cases, but be aware that it could cause issues, especially with `use cache` or `unstable_cache` (Some external change may not be reflected in the key, leading to stale data)
```
vicb, line 87: IIUC:
- `false` -> current behavior, the data cache is not shared across builds
- `true` -> the data cache is always shared across build
- Next.js -> the data is shared outside of `use cache` / `unstable_cache`

2 questions:
- What would it take to implement the same behavior has Next.js - or is there any reason not to pursue this at all?
- Should the config value be an enum for extensibility?

packages/open-next/src/adapters/cache.ts
```diff
@@ -49,22 +54,24 @@ export default class Cache {
 
     const softTags = typeof options === "object" ? options.softTags : [];
     const tags = typeof options === "object" ? options.tags : [];
-    return isFetchCache(options)
-      ? this.getFetchCache(key, softTags, tags)
-      : this.getIncrementalCache(key);
+    const isDataCache = isFetchCache(options);
+    return isDataCache
```
vicb: nit: I liked the former version better (data/fetch/data sounds weird)

```suggestion
    return isFetchCache(options)
```

packages/open-next/src/utils/cache.ts
```diff
@@ -80,3 +80,19 @@ export async function writeTags(
...
+  // We always prepend the build ID to the cache key for ISR/SSG cache entry
+  // For data cache, we only prepend the build ID if the persistentDataCache is not enabled
+  const shouldPrependBuildId =
+    globalThis.openNextConfig.dangerous?.persistentDataCache !== true ||
+    !isDataCache;
+  if (shouldPrependBuildId) {
+    // If we don't have a build ID, we just return the key as is
+    if (!process.env.NEXT_BUILD_ID) {
+      return key;
+    }
```
vicb: nit:

```suggestion
  if (shouldPrependBuildId && process.env.NEXT_BUILD_ID) {    
```

.changeset/slow-walls-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": minor
+---
+
+Add an option to keep the data cache persistent between deployments
```
vicb: We should document how the cache overrides should be updated to support that - at least that they need to be.

Maybe we should add some docs on the interfaces?

packages/open-next/src/utils/cache.ts
```diff
@@ -80,3 +80,19 @@ export async function writeTags(
   // Here we know that we have the correct type
   await globalThis.tagCache.writeTags(tagsToWrite as any);
 }
+
+export function createCacheKey(key: string, isDataCache: boolean): string {
```
vicb: I would very much prefer:

```ts
createCacheKey(baseKey, { isDataCache: true });
// or
createDataCacheKey(baseKey);
```

over

```ts
const key = createCacheKey(baseKey, true);
```

packages/open-next/src/adapters/cache.ts
```diff
@@ -31,7 +36,7 @@ function isFetchCache(
 // We need to use globalThis client here as this class can be defined at load time in next 12 but client is not available at load time
 export default class Cache {
```
vicb, line 39: Maybe we should document baseKey vs key on the interface?

packages/open-next/src/adapters/cache.ts
```diff
@@ -49,22 +54,24 @@ export default class Cache {
...
+    debug("get fetch cache", { baseKey, softTags, tags });
     try {
+      const key = createCacheKey(baseKey, true);
       const cachedEntry = await globalThis.incrementalCache.get(key, "fetch");
 
       if (cachedEntry?.value === undefined) return null;
 
       const _tags = [...(tags ?? []), ...(softTags ?? [])];
       const _lastModified = cachedEntry.lastModified ?? Date.now();
       const _hasBeenRevalidated = await hasBeenRevalidated(
```
vicb: Maybe we should add a comment explaining why the base key is used (inline + on the method)?

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -4,10 +4,33 @@ import { fromReadableStream, toReadableStream } from "utils/stream";
...
+  try {
+    const shouldPrependBuildId =
+      globalThis.openNextConfig.dangerous?.persistentDataCache !== true;
+    if (shouldPrependBuildId) {
+      return key;
+    }
+    const [_buildId, ...rest] = JSON.parse(key);
+    return JSON.stringify([...rest]);
+  } catch (e) {
+    debug("Error while parsing composable cache key", e);
```
vicb: Should this be warning/error?

packages/open-next/src/overrides/incrementalCache/s3-lite.ts
```diff
@@ -34,11 +34,10 @@ const awsFetch = async (key: string, options: RequestInit) => {
 };
 
 function buildS3Key(key: string, extension: Extension) {
-  const { CACHE_BUCKET_KEY_PREFIX, NEXT_BUILD_ID } = process.env;
+  const { CACHE_BUCKET_KEY_PREFIX } = process.env;
   return path.posix.join(
     CACHE_BUCKET_KEY_PREFIX ?? "",
```
vicb:
```suggestion
  return path.posix.join(
    process.env.CACHE_BUCKET_KEY_PREFIX ?? "",
```

packages/open-next/src/types/open-next.ts
conico974: Next.js (and by that i mean `next start`) is reusing the data cache for `unstable_cache` and the fetch cache. But it's a local filesystem, so you have to keep it between build **and** find a way to merge it with the runtime one. 
For example if you deploy with docker, you'll likely erase all data that was generated at runtime.

`use cache` is purely in memory right now (completely useless for us, and that's not the case on Vercel)

One big things that Next completely disregards (and the main reason why it was implemented this way), is that the key generated for the data cache (for `unstable_cache` and `use cache`) may not change when there is external changes (i.e. a dependencies that was updated, a global, an env variable...) and this could cause some pretty big issues if not handled properly by the user.

I think a boolean is enough, it would get too complicated to add other behavior IMO

vicb: Looking at the changes in this PR make me think that implementing the Next.js behavior should be possible?

vicb (commented):
Nice!

I added a few minor comments but changes look good.

Probably we should doc that overrides should not use the BUILD_ID to enable the feature.

I question:

For composable cache, we use 

```ts
globalThis.tagCache.hasBeenRevalidated(
          result.value.tags,
          result.lastModified,
```

vs `hasBeenRevalidated` from `utils` - any particular reason. Would be nice to add comments.

packages/open-next/src/types/open-next.ts
conico974: Yeah, but the problem is that the Next.js behavior is not really intended.
Based on what they say (and how it works in Vercel), all data cache are supposed to be persistent between build/deployment.
The `use cache` behavior is likely just because they haven't implemented it properly yet (it's still considered experimental)

conico974:
I'll update the PR with the changes later today

vicb:
Thought: In the Couldflare adapter the cache key is `prefix/build-id/hash(key)`

- build-id is used to partition the cache
- we need to `hash(...)` to fit within the allowed length and chars

That will become harder to do with this PR.

There are multiple ways to solve this but while thinking about this I was wondering if the change could not be implemented by only updating the overrides? i.e. we could pass a key that is independent of the BUILD_ID and then up to the override to add the build Id or not.

conico974:
> There are multiple ways to solve this but while thinking about this I was wondering if the change could not be implemented by only updating the overrides? i.e. we could pass a key that is independent of the BUILD_ID and then up to the override to add the build Id or not.

That's exactly what was done before (except for the composable cache). 
I feel like this is something that we need to handle here, not the override itself. It feels weird to have to do the same thing on every single override.

One option would be to have a special type as a key like 
```ts
interface CacheKey {
  baseKey: string,
  buildId?:string
}
```
You can then decide what you do with it.

vicb:
> One option would be to have a special type as a key like

+1

It could also help alleviate the confusion between `key: string` and `baseKey: string`

.changeset/slow-walls-allow.md
```diff
@@ -0,0 +1,7 @@
+---
+"@opennextjs/aws": minor
+---
+
+Add an option to keep the data cache persistent between deployments.
+
+BREAKING CHANGE: Incremental cache keys are now an object of type `CacheKey` instead of a string. The new type includes properties like `baseKey`, `buildId`, and `cacheType`. Build_id is automatically provided according to the cache type and the `dangerous.persistentDataCache` option. Up to the Incremental Cache implementation to use it as they see fit.
```
vicb: Explicitly say that the overrides should be updated?

(Adding this text to the PR description)

packages/open-next/src/adapters/cache.ts
vicb: Re-opening.
Could we expand on why this is not using the obj key?

packages/open-next/src/adapters/cache.ts
```diff
@@ -191,118 +204,104 @@ export default class Cache {
...
-              );
-            }
-            break;
-          }
-          case "APP_PAGE": {
-            const { html, rscData, headers, status } = data;
-            await globalThis.incrementalCache.set(
-              key,
-              {
+              await globalThis.incrementalCache.set(key, {
```
vicb, line 255: Look like this is a formatting only diff?

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -1,22 +1,48 @@
...
+import { debug, warn } from "./logger";
 
 const pendingWritePromiseMap = new Map<string, Promise<ComposableCacheEntry>>();
+/**
+ * Get the cache key for a composable entry.
+ * Composable cache keys are a special cases as they are a stringified version of a tuple composed of a representation of the BUILD_ID and the actual key.
+ * @param key The composable cache key
+ * @returns The composable cache key.
+ */
+function getComposableCacheKey(key: string): CacheKey<"composable"> {
```
vicb: Could this be merged into createCacheKey?

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -1,22 +1,48 @@
...
+      baseKey: key,
+    };
+  }
+}
 
 export default {
-  async get(cacheKey: string) {
+  async get(key: string) {
     try {
+      const cacheKey = getComposableCacheKey(key);
```
vicb: Would it make sense to move that l45 and use `key` on l42 and 43?

edit:
Oh actually I think that's what we want as they might differ.
Should we add a comment warning that `key` should not be used?

edit2:
But I guess that I have the same question as the tag cache: why wouldn't we use the build?

packages/open-next/src/overrides/incrementalCache/fs-dev.ts
```diff
@@ -4,33 +4,32 @@ import path from "node:path";
...
 
 const getCacheKey = (key: string) => {
   return path.join(basePath, `${key}.cache`);
 };
 
 const cache: IncrementalCache = {
   name: "fs-dev",
-  get: async (key: string) => {
-    const fileData = await fs.readFile(getCacheKey(key), "utf-8");
+  get: async ({ baseKey }) => {
```
vicb:
```suggestion
  get: async (cacheKey: CacheKey) => {
    // This cache is always shared across build (the build id is not used)
    const { baseKey } = cacheKey; 
```

packages/open-next/src/overrides/incrementalCache/multi-tier-ddb-s3.ts
```diff
@@ -40,9 +41,8 @@ const awsFetch = (body: RequestInit["body"], type: "get" | "set" = "get") => {
   );
 };
 
-const buildDynamoKey = (key: string) => {
-  const { NEXT_BUILD_ID } = process.env;
-  return `__meta_${NEXT_BUILD_ID}_${key}`;
+const buildDynamoKey = (key: CacheKey<CacheEntryType>) => {
+  return `__meta_${key.buildId ? `${key.buildId}_` : ""}${key.baseKey}`;
```
vicb: nit: could this be simplified?
not the exact same output as the current code but it should not matter?

```suggestion
  return `__meta_${key.buildId ?? ""}_${key.baseKey}`;
```

packages/open-next/src/utils/cache.ts
```diff
@@ -80,3 +82,34 @@ export async function writeTags(
...
+      cacheType: type,
+      buildId,
+      baseKey: key,
+    };
+  }
+  return {
+    cacheType: type,
+    buildId: undefined,
+    baseKey: key,
+  } as CacheKey<CacheType>;
```
vicb, line 133: is `as CacheKey<CacheType>` needed here?

packages/open-next/src/utils/cache.ts
```diff
@@ -80,3 +82,34 @@ export async function writeTags(
   // Here we know that we have the correct type
   await globalThis.tagCache.writeTags(tagsToWrite as any);
 }
+
+export function createCacheKey<CacheType extends CacheEntryType>({
+  key,
+  type,
+}: { key: string; type: CacheType }): CacheKey<CacheType> {
+  // We always prepend the build ID to the cache key for ISR/SSG cache entry
+  // For data cache, we only prepend the build ID if the persistentDataCache is not enabled
```
vicb: "prepend" do not sound right any more.

would be great to update the comments and move to their corresponding case.

packages/open-next/src/adapters/composable-cache.ts
conico974: Yeah probably

packages/open-next/src/adapters/composable-cache.ts
conico974: Here it doesn't really matter, it's in memory anyway.
For the tag cache, technically we don't need the build_id as it is time based.
There is one case actually when it is useful to have the build_id, it's for more complex deployment (like blue/green).
I think what make sense is just to pass the `CacheKey` to the tag cache as well and let the implementation decides.
For the default one, I think we should just follow the `persistentDataCache` option
WDYT ?

vicb: SGTM

.changeset/slow-walls-allow.md
```diff
@@ -0,0 +1,8 @@
+---
+"@opennextjs/aws": minor
+---
+
+Add an option to keep the data cache persistent between deployments.
```
vicb, line 5:
```suggestion
Add `dangerous.persistentDataCache` option to persist the data cache across deployments.
```

.changeset/slow-walls-allow.md
```diff
@@ -0,0 +1,8 @@
...
+
+BREAKING CHANGE: Incremental cache keys are now an object of type `CacheKey` instead of a string. The new type includes properties like `baseKey`, `buildId`, and `cacheType`. Build_id is automatically provided according to the cache type and the `dangerous.persistentDataCache` option. Up to the Incremental Cache implementation to use it as they see fit.
```
vicb, line 7: `TagKey` should be added here too

PR #922, opened 2025-07-10
fix: middleware type

vicb:
`assetResolver` can be used regardless if the middleware is external

### DIFF
--- a/packages/open-next/src/types/open-next.ts
+++ b/packages/open-next/src/types/open-next.ts
@@ -419,7 +419,7 @@ export interface OpenNextConfig {
    * Override the default middleware
    * When `external` is true, the middleware need to be deployed separately.
    * It supports both edge and node runtime.
-   * @default undefined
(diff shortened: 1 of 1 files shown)


packages/open-next/src/types/open-next.ts
```diff
@@ -400,15 +404,25 @@ export interface OpenNextConfig {
...
+
+export interface OpenNextConfig {
+  default: FunctionOptions;
+  functions?: Record<string, SplittedFunctionOptions>;
+
+  /**
+   * Override the default middleware
+   * When `external` is true, the middleware need to be deployed separately.
+   * It supports both edge and node runtime.
+   * @default undefined
```
conico974:
```suggestion
   * @default undefined - Which is equivalent to `external: false`
```

conico974 (approved):
LGTM

vicb:
Thanks for the review Nico!

PR #926, opened 2025-07-15
Fix edge functions bundle path and add additionalInject option support

psyrenpark:
## Summary
This PR includes two main improvements to the OpenNext edge functions handling:

1. **Fix edge functions bundle path**: Corrects the bundle path from `.open-next/functions` to `.open-next/server-functions` to match the actual build output location.

2. **Add additionalInject option support**: Introduces a new `additionalInject` option for injecting custom code directly into the esbuild banner section.

## Changes
- **Fix**: Updated edge functions bundle path in `generateOutput.ts` to use the correct server-functions directory
- **Feature**: Added `additionalInject` option support in `createEdgeBundle.ts` 
- **Documentation**: Added comprehensive JSDoc documentation for the new `additionalInject` option with usage guidelines and caution notes

## Breaking Changes
None. These changes are backward compatible.

## Notes
- The `additionalInject` option should be used with caution as it directly injects code into the esbuild banner section
- The bundle path fix ensures edge functions are correctly located during deployment

## Testing
- [ ] Verified edge functions bundle path matches build output
- [ ] Tested additionalInject option functionality
- [ ] Confirmed backward compatibility

### DIFF
--- a/packages/open-next/src/types/open-next.ts
+++ b/packages/open-next/src/types/open-next.ts
@@ -361,6 +361,15 @@ export interface SplittedFunctionOptions extends FunctionOptions {
    * @default []
    */
   patterns: string[];
+
+
+  /**
+   * Additional inject code for the function.
+   * This is used to inject code into the function.
+   * Used when code needs to be directly injected into the banner section during esbuild. Should be used with caution.
+   * @default undefined
+   */
+  additionalInject?: string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/open-next/src/types/open-next.ts
```diff
@@ -361,6 +361,15 @@ export interface SplittedFunctionOptions extends FunctionOptions {
...
   patterns: string[];
+
+
+  /**
+   * Additional inject code for the function.
+   * This is used to inject code into the function.
+   * Used when code needs to be directly injected into the banner section during esbuild. Should be used with caution.
+   * @default undefined
+   */
+  additionalInject?: string;
```
conico974: I don't think the `additionalInject` option here make that much sense. it would only work for edge bundle.
An option here should work everywhere, not just in edge runtime

conico974 (approved):
I took the liberty to update the PR.
Regarding the `additionalInject` that I removed, if you still need this i'd rather have another PR. And i don't think it should be done this way

PR #927, opened 2025-07-15
fix(e2e): next/head elements in pages router are getting duplicated

sommeeeer:
In Next `15.4.x` elements added with `next/head` and possibly some more like `<meta name="viewport"....` gets duplicated in the rendered HTML. 

We skip this tests for now until [this](https://github.com/vercel/next.js/issues/81655) issue in Next is solved.


conico974 (approved):
Thanks @sommeeeer

PR #928, opened 2025-07-16
Fix for Next 15.4+

### LINKED ISSUE #848: [BUG] `next.config` rewrites result in request URL being rewrite destination
### Describe the bug

When I have a `rewrite` rule in my `next.config.ts` with a destination to an API route, the request URL received by the API route is that of the _destination_. This is different behavior from the same app deployed on Vercel (or running the app locally), where the request URL received by the API route corresponds to the _originally requested URL_ from the client.

Video demonstration:

https://github.com/user-attachments/assets/8a2c7d64-3c41-4c95-bcee-c4e3b2efd83e

### Steps to reproduce

Reproduction repo (this was used for the video above, and can be directly deployed to both Cloudflare and Vercel): https://github.com/arvinpoddar/cloudflare-rewrites-repro

Reproduction steps:

1. In your `next.config.ts`, create a rewrite rule that rewrites `/about` to `/api/hello`. 
2. Create a `/api/hello/route.ts`, and `console.log` the incoming request URL.
3. Navigate to `/about` in your browser
4. When locally running the app, or when the app is deployed to Vercel, the logged request's URL path will be `/about`, which is the originally requested path.
5. When the same app is deployed to Cloudflare, the logged request's URL path will be `/api/hello`, which is the destination of the rewrite.

### Expected behavior

I would expect the request URL to be that of the original client request, rather than the destination of the rewrite. This would be consistent with the app's behavior locally and on Vercel.

### @opennextjs/cloudflare version

~1.0.0-beta.0 || ^1.0.0

### Wrangler version

^4.13.2

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6020
  Available memory (MB): 16384
  Available CPU cores: 10
Binaries:
  Node: 21.6.1
  npm: 10.2.4
  Yarn: 1.22.22
  pnpm: 10.7.0
Relevant Packages:
  next: 15.3.1 // Latest available version is detected (15.3.1).
  eslint-config-next: 15.3.1
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
```

### Additional context

_No response_

conico974:
Migrate to a canary version for now (we need to wait for 15.4.2)
Enable local execution of some examples (first step to run e2e test without deploying)
And fix the page router functionality. This one needs additional testing and be restricted to versions above 15.4.2.
All e2e works on this PR.

It should also close #848 (again needs additional testing)

### DIFF
--- a/packages/open-next/src/build/createServerBundle.ts
+++ b/packages/open-next/src/build/createServerBundle.ts
@@ -243,9 +243,9 @@ async function generateBundle(
     "15.2.0",
   );
 
-  const isBefore154 = buildHelper.compareSemver(
+  const isAfter154 = buildHelper.compareSemver(
     options.nextVersion,
-    "<",
+    ">=",
     "15.4.0",
   );
 
@@ -266,7 +266,7 @@ async function generateBundle(
         ...(disableRouting ? ["withRouting"] : []),
         ...(isAfter142 ? ["patchAsyncStorage"] : []),
         ...(isAfter141 ? ["appendPrefetch"] : []),
-        ...(isBefore154 ? ["setInitialURL"] : []),
+        ...(isAfter154 ? [] : ["setInitialURL"]),
       ],
     }),
     openNextReplacementPlugin({
--- a/packages/open-next/src/build/createServerBundle.ts
+++ b/packages/open-next/src/build/createServerBundle.ts
@@ -243,6 +243,12 @@ async function generateBundle(
     "15.2.0",
   );
 
+  const isBefore154 = buildHelper.compareSemver(
+    options.nextVersion,
+    "<",
+    "15.4.0",
+  );
+
   const disableRouting = isBefore13413 || config.middleware?.external;
 
   const updater = new ContentUpdater(options);
@@ -260,6 +266,7 @@ async function generateBundle(
         ...(disableRouting ? ["withRouting"] : []),
         ...(isAfter142 ? ["patchAsyncStorage"] : []),
         ...(isAfter141 ? ["appendPrefetch"] : []),
+        ...(isBefore154 ? ["setInitialURL"] : []),
       ],
     }),
     openNextReplacementPlugin({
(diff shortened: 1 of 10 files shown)


sommeeeer:
Good job on this Nico! 

I tested that the rewrites will have the correct `request.url` now. That is of the original request url, and not the destination. 

Is there any possibility that we might fix that the `request.url` is wrong in rewrites on earlier versions aswell?

packages/open-next/src/core/requestHandler.ts
```diff
@@ -266,6 +267,12 @@ async function processRequest(
     //#endOverride
 
     // Next Server
+    // TODO: only enable this on Next 15.4+
+    const reqUrl = new URL(routingResult.initialURL);
```
sommeeeer: We compute `initialURL` above on L242

packages/open-next/src/build/createServerBundle.ts
```diff
@@ -260,6 +266,7 @@ async function generateBundle(
         ...(disableRouting ? ["withRouting"] : []),
         ...(isAfter142 ? ["patchAsyncStorage"] : []),
         ...(isAfter141 ? ["appendPrefetch"] : []),
+        ...(isBefore154 ? ["setInitialURL"] : []),
```
khuezy: In the TODO comment below:
`TODO: only enable this on Next 15.4+`

Shouldn't this be `isAfter154`? Or swap the 2 arrays?

conico974: It is correct, it's a delete, and we want to delete that on before 15.4.
But yeah it might be better the other way around (by swapping the arrays)

khuezy: Ah I see it now after expanding the code above.

conico974:
Thanks for the review

Issue #933, opened 2025-07-26
On Node 22.17 the installation of sharp in image optimization bundling will create invalid symbolic links


sommeeeer:
This was reported first on Discord in this [thread](https://discord.com/channels/983865673656705025/1387809824732876810). During `sst deploy` you would get an error like this:
```Error: ENOENT: no such file or directory, open '/home/user/reproductions/sst-opennext-semver/.sst/dist/asset.ac657782e4de0bc902dc2aa50204299a3d4f962bb14db71e08584c7ec5c4da48/node_modules/.bin/semver'```

This symbolic link was pointing to a file that does not exist so it would crash during `sst deploy` as it would not find those files. 
```
semver -> /tmp/open-next-install-image-optimization-functioneUpKfE/node_modules/semver/bin/semver.js
```

It seems to only affect Node version `22.17.0` and `22.17.1`. All other Node versions seems to work fine. I debugged this down to these [lines](https://github.com/opennextjs/opennextjs-aws/blob/c48ac9565863706486f796a75aaecdd27aebd551/packages/open-next/src/build/installDeps.ts#L48-L54). Setting `dereference: true` should remove all references to symbolic links and copy the target to the destination instead.

In theory we could patch this one on our side, or wait until Node fixes it upstream.

Related issue: https://github.com/nodejs/node/issues/59168

sommeeeer:
Fixed in #934

berenddeboer:
Seeing exactly the same issue with sst v2 on node 24.11.1. Very odd. Can only fix this by patching `installDeps.js` too.

PR #934, opened 2025-07-26
fix(image-optimization): Workaround for broken symlink dereferencing in Node 22.17.0 and 22.17.1

sommeeeer:
For #933

`npm install` will create symbolic links by default for binaries in `node_modules/.bin`. I tried using [`bin-links`](https://docs.npmjs.com/cli/v11/commands/npm-install#bin-links) in `npm install` but I did not get it to work. Instead we do it manually for the affected Node versions.

### DIFF
--- a/packages/open-next/src/build/installDeps.ts
+++ b/packages/open-next/src/build/installDeps.ts
@@ -48,9 +48,36 @@ export function installDependencies(
     fs.cpSync(
       path.join(tempInstallDir, "node_modules"),
       path.join(outputDir, "node_modules"),
-
       { recursive: true, force: true, dereference: true },
     );
+
+    // This is a workaround for Node `22.17.0` and `22.17.1`
+    // Can remove these lines when this is fixed upstream: https://github.com/nodejs/node/issues/59168
+    const nodeVersion = process.version;
+    if (nodeVersion === "v22.17.0" || nodeVersion === "v22.17.1") {
+      const tempBinDir = path.join(tempInstallDir, "node_modules", ".bin");
+      const outputBinDir = path.join(outputDir, "node_modules", ".bin");
+
+      for (const fileName of fs.readdirSync(tempBinDir)) {
+        const symlinkPath = path.join(tempBinDir, fileName);
+        const stat = fs.lstatSync(symlinkPath);
+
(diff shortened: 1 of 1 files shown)


packages/open-next/src/build/installDeps.ts
```diff
@@ -48,10 +48,37 @@ export function installDependencies(
...
       path.join(tempInstallDir, "node_modules"),
       path.join(outputDir, "node_modules"),
-
       { recursive: true, force: true, dereference: true },
     );
 
+    // This is a workaround for Node `22.17.0` and `22.17.1`
+    // Can remove these lines when this is fixed upstream: https://github.com/nodejs/node/issues/59168
+    const nodeVersion = process.version;
+    if (nodeVersion === "v22.17.0" || nodeVersion === "v22.17.1") {
```
conico974, line 57: Could we avoid copying things twice here, we are already copying everything in L48

sommeeeer: I could use `filter` in `fs.CopySyncOptions` above instead. WDYT?

conico974: Why not just this logic to copy everything in the broken node version, and else just what's on L48

.changeset/hip-lies-matter.md
```diff
@@ -0,0 +1,9 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: Workaround for broken symlink dereferencing in Node 22.17.0 and 22.17.1
+
+The `dereference: true` option in `fs.cpSync()` is broken on version 22.17.0 and 22.17.1. This fix will do it manually for the binaries in `node_modules/.bin`. We can revert this once fixed upstream.
```
vicb: "We can revert this once fixed upstream"

IMO we should not revert the workaround as people might still use an affected version.

sommeeeer: Good point Victor

sommeeeer:
Thanks for the reviews! Could you take another look when you have time Nico? We could in theory just skip the `node_modules/.bin` folder aswell, cause CLI commands wont be used in a server function bundle.

conico974 (approved):
LGTM Thanks

Issue #940, opened 2025-08-02
Next.config.ts redirects produces decoded Location header


leog:
When configuring `next.config.ts` redirects as follows, and you visit `/about?param=test%2F%2F` the resultant Location header has the query parameter decoded, which produces an error in Cloudfront.

This worked correctly up until v3.5.5 according to my research, from v3.5.6 onwards this is reproducible, including v3.7.2. Also tested v3.6.3 which includes #871, directly tackling this issue, but no luck.

```
import type { NextConfig } from "next";

const nextConfig: NextConfig = {
  async redirects() {
    return [
      {
        source: "/about",
        destination: "/",
        permanent: false,
      },
    ];
  },
};

export default nextConfig;
```

conico974:
Fixed in #941

PR #941, opened 2025-08-02
fix: Ensure redirects are properly encoded and relative/absolute

sommeeeer:
Currently in draft mode. It will pass all our tests, but requires more manual testing. Especially the middleware part.

For #940 and https://github.com/opennextjs/opennextjs-cloudflare/issues/809

For redirects coming from next config we want that the `Location` header value is encoded properly according to RFC. In `next start` and deployed to Vercel the header can also be relative or absolute. 

I am a bit unsure what to do about `NextResponse.redirect()` coming from the middleware. Should we encode it aswell? I'll keep investigating this but are open for suggestions.

### DIFF
--- a/packages/open-next/src/core/routing/middleware.ts
+++ b/packages/open-next/src/core/routing/middleware.ts
@@ -15,6 +15,7 @@ import {
   convertBodyToReadableStream,
   getMiddlewareMatch,
   isExternal,
+  normalizeLocationHeader,
 } from "./util.js";
 
 const middlewareManifest = MiddlewareManifest;
@@ -94,6 +95,15 @@ export async function handleMiddleware(
     url,
     body: convertBodyToReadableStream(internalEvent.method, internalEvent.body),
   } as unknown as Request);
+  if (result.headers.has("Location")) {
+    result.headers.set(
+      "Location",
+      normalizeLocationHeader(
+        result.headers.get("Location") as string,
+        internalEvent.url,
+      ),
+    );
+  }
   const statusCode = result.status;
 
   /* Apply override headers from middleware
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -437,3 +437,27 @@ export async function invalidateCDNOnRequest(
     ]);
   }
 }
+
+/**
+ * Normalizes the Location header to either be a relative path or a full URL.
+ * If the Location header is relative to the host, it will return a relative path.
+ * If it is an absolute URL, it will return the full URL.
+ * Both cases will ensure that the URL is properly encoded according to RFC
+ *
+ * @param location The Location header value
+ * @param url The original request URL
+ * @returns A normalized Location header value
+ */
+export function normalizeLocationHeader(location: string, url: string): string {
+  const locationUrl = new URL(location);
+  const host = new URL(url).host;
+  
+  // Encode the search parameters to ensure they are valid according to RFC
+  const encodedSearch = locationUrl.searchParams.toString() ? `?${locationUrl.searchParams.toString()}` : '';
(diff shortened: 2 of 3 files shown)


leog:
Confirmed it works as expected using your changes @sommeeeer:
https://github.com/leog/OpenNext-AWS-Normalize-the-Location-header-in-redirects/commit/540caa06802785d6b5519df39d7747e0eb09deb4

https://d13iq8evz1d8fc.cloudfront.net/about?param=test%2F%2F

Very promising, thank you! 💪

packages/open-next/src/core/routing/middleware.ts
```diff
@@ -94,7 +97,18 @@ export async function handleMiddleware(
     url,
     body: convertBodyToReadableStream(internalEvent.method, internalEvent.body),
   } as unknown as Request);
+
   const statusCode = result.status;
+  // We want to normalize the Location header if its a redirect response. Headers.has() doesn't care about the case of the header name.
+  if (REDIRECTS.has(statusCode) && result.headers.has("Location")) {
+    result.headers.set(
```
conico974: You cannot do that here, if you return a fetch from the middleware that is a redirect, you will not be able to set the headers as they will be immutable.
You'll need to do it here https://github.com/opennextjs/opennextjs-aws/blob/2c8a649be4f5ca5a677d7251e7636379f6270534/packages/open-next/src/core/routing/middleware.ts#L139-L153

packages/open-next/src/core/routing/util.ts
```diff
@@ -437,3 +437,41 @@ export async function invalidateCDNOnRequest(
...
+ *
+ * @param location The Location header value
+ * @param baseUrl The base URL to use for relative paths (i.e the original request URL)
+ * @returns An encoded absolute or relative Location header value
+ */
+export function normalizeLocationHeader(
+  location: string,
+  baseUrl: string,
+): string {
+  try {
```
conico974: Why not use `URL.canParse` instead of a try catch block

packages/open-next/src/core/routing/util.ts
```diff
@@ -437,3 +437,41 @@ export async function invalidateCDNOnRequest(
...
+    const locationUrl = new URL(location);
+    const origin = new URL(baseUrl).origin;
+
+    // Encode the search parameters to ensure they are valid according to RFC
+    const encodedSearch = locationUrl.searchParams.toString()
```
conico974: Use a variable for `locationUrl.searchParams.toString()` instead of calling it twice

packages/open-next/src/core/routing/util.ts
```diff
@@ -437,3 +437,41 @@ export async function invalidateCDNOnRequest(
...
+      ? `?${locationUrl.searchParams.toString()}`
+      : "";
+    const href =
```
conico974: I prefer a template literals here. Not a big deal if you prefer it this way.

packages/open-next/src/core/routing/middleware.ts
sommeeeer: Alright, thanks for the clarification on that. I'll address that right away.

sommeeeer:
I turned this into draft mode again. I need to do some more thouroghly testing as it seems I overlooked the fact that Next actually does not encode the search paramaters in the Location header coming from the middleware.

Update: The PR is ready now we should be matching Next behavior now. They encode query params in redirects from Next Config, but not with `URLSearchParams`. They use `stringify` from `node:querystring`. Query parameters in Redirects from the middleware is not touched.

conico974 (approved):
LGTM Thanks!!

sommeeeer:
Thanks for the review Nico!

PR #944, opened 2025-08-03
Add shouldBypassTagCache option for incremental caches

conico974:
Introduce a new option to bypass tag cache validation from the incremental caches, along with unit tests to ensure expected behavior.

This is useful in 2 situations: 
- When you don't need the tag cache only for some routes/data cache entry
- When you're sure that a route has not been revalidated. For example in cloudflare when a cache entry is coming from the cache api and that you have automatic cache invalidation.

### DIFF
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -63,11 +63,9 @@ export default class Cache {
 
       const _tags = [...(tags ?? []), ...(softTags ?? [])];
       const _lastModified = cachedEntry.lastModified ?? Date.now();
-      const _hasBeenRevalidated = cachedEntry.shouldBypassTagCache ? false : await hasBeenRevalidated(
-        key,
-        _tags,
-        cachedEntry,
-      );
+      const _hasBeenRevalidated = cachedEntry.shouldBypassTagCache
+        ? false
+        : await hasBeenRevalidated(key, _tags, cachedEntry);
 
       if (_hasBeenRevalidated) return null;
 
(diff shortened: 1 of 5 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -82,11 +80,13 @@ export default class Cache {
...
         if (path) {
-          const hasPathBeenUpdated = await hasBeenRevalidated(
-            path.replace("_N_T_/", ""),
-            [],
-            cachedEntry,
-          );
+          const hasPathBeenUpdated = cachedEntry.shouldBypassTagCache
+            ? false
+            : await hasBeenRevalidated(
+                path.replace("_N_T_/", ""),
```
khuezy: Is `_N_T_/` worth extracting as a constant? I'm honestly not sure what the NT stands for. Next Tag?

khuezy (approved):
LGTM! Just 1 note.

packages/open-next/src/adapters/cache.ts
conico974: Not sure what it stands for to be honest. It's just the prefix that next uses for soft tag (i.e. route and layout).
You're right, we should probably use a constant here

sommeeeer: https://github.com/sommeeeer/next.js/blob/canary/packages/next/dist/lib/constants.d.ts#L22

sommeeeer: https://github.com/vercel/next.js/blob/d3303229b397f2017a14b0712668d52a860efc7e/packages/next/src/lib/constants.ts#L35

PR #945, opened 2025-08-04
fix: Validate statusCode is number and not NaN in OpenNextNodeResponse constructor

sommeeeer:


packages/open-next/src/http/openNextResponse.ts
```diff
@@ -74,7 +74,8 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
     statusCode?: number,
   ) {
     super();
-    if (statusCode !== undefined) {
+    // We only set the status code if it is not a NaN and it is a number
+    if (!Number.isNaN(statusCode) && typeof statusCode === "number") {
```
vicb:
```suggestion
    if (Number.isInteger(statusCode)) {
```

vicb (approved):
Thanks, LGTM with 1 nit

We should port the tests to the cloudflare adapter when this is merged there.

sommeeeer:
Thanks for the review Victor!

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -74,7 +74,8 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
     statusCode?: number,
   ) {
     super();
-    if (statusCode !== undefined) {
+    // We only set the status code if it is not a NaN and it is a number
+    if (Number.isInteger(statusCode)) {
```
conico974: We should also not apply it if it's not in a correct range as well. 
Technically, setting a value to 10 billions here would work, but it doesn't make any sense.

sommeeeer: Alright will do so. For some reason TypeScript is not happy about `Number.isInteger()`. Should I cast it or go for something else?

sommeeeer: `if (statusCode && Number.isInteger(statusCode) && statusCode >= 100 && statusCode <= 599)`

Are we happy about this one?

sommeeeer:
Thanks for the review Nico!

PR #948, opened 2025-08-04
add(patches): Drop babel from getting pulled in by node-environment

### LINKED ISSUE #947: Upstream babel drop patch
We should upstream https://github.com/opennextjs/opennextjs-cloudflare/pull/843

Some reported babel still present is 15.2.5

Actually I think I made a mistake when I moved the patches to aws.

The [original patch](https://github.com/opennextjs/opennextjs-cloudflare/pull/702/files#diff-7738cd8f60bec3bd7aea223b2ecc0f4e09f3231606791e0228669dd1c23b3186R60) was patching `node-environment.js` - I must have screw up when moving the code into this repo

edit: we can drop the [uneffective patch](https://github.com/opennextjs/opennextjs-aws/blob/e1048ac9bfffa79d32b1cab810619c41a66187c3/packages/open-next/src/build/patch/patches/patchNextServer.ts#L95-L99) and replace with the one from cloudflare.

Once this is done we can remove the patch from cloudflare.

/cc @conico974 @sommeeeer

sommeeeer:
Closes #947

I applied it on `>=15.0.0`.

### DIFF
--- a/packages/open-next/src/build/patch/patches/patchNodeEnvironment.ts
+++ b/packages/open-next/src/build/patch/patches/patchNodeEnvironment.ts
@@ -1,0 +1,34 @@
+import { Lang } from "@ast-grep/napi";
+import { getCrossPlatformPathRegex } from "utils/regex.js";
+import { createPatchCode } from "../astCodePatcher.js";
+import type { CodePatcher } from "../codePatcher.js";
+
+/**
+ * Drops `require("./node-environment-extensions/error-inspect");`
+ *
+ * This is to avoid pulling babel (~4MB)
+ */
+export const rule = `
+rule:
+  pattern: require("./node-environment-extensions/error-inspect");
+fix: |-
+  // Removed by OpenNext
+  // require("./node-environment-extensions/error-inspect");
+`;
+
+export const patchNodeEnvironment: CodePatcher = {
+  name: "patch-node-environment-error-inspect",
+  patches: [
+    {
+      pathFilter: getCrossPlatformPathRegex(
+        String.raw`/next/dist/server/node-environment\.js$`,
+        {
+          escape: false,
+        },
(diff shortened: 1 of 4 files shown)


sommeeeer:
cc @vicb

packages/tests-unit/tests/build/patch/patches/patchNodeEnvironment.test.ts
```diff
@@ -0,0 +1,48 @@
+import { rule } from "@opennextjs/aws/build/patch/patches/patchNodeEnvironment";
```
vicb: I think you need a .js here

vicb (approved):
Lgtm with one comment. 
Thanks!

sommeeeer:
Thanks for the review Victor!

packages/open-next/src/build/patch/patches/patchNodeEnvironment.ts
```diff
@@ -0,0 +1,34 @@
...
+  patches: [
+    {
+      pathFilter: getCrossPlatformPathRegex(
+        String.raw`/next/dist/server/node-environment\.js$`,
+        {
+          escape: false,
+        },
+      ),
+      contentFilter: /error-inspect/,
+      patchCode: createPatchCode(rule, Lang.JavaScript),
```
vicb: nit: the default is TS which is a superset of JS so that should not be needed. Most of the type we do not explicitly pass JS.

```suggestion
      patchCode: createPatchCode(rule),
```

sommeeeer: Alright, that makes sense!

vicb (approved):
Ship it 🚀
Thanks!

PR #950, opened 2025-08-05
chore: Exclude more packages

sommeeeer:
Add more packages that can be excluded so they don't get copied into the final bundles `node_modules`. Also add debug to determine which packages actually gets skipped. I've run the E2E successfully. These packages will be skipped now:

- `typescript`
- `next/dist/compiled/babel`
- `next/dist/compiled/babel-packages`
- `next/dist/compiled/amphtml-validator`

It will decrease the unzipped folder by 15mb and the zipped one with 4mb:
 
```bash
❯ du -hs * | sort -rh
46M	with-deps
31M	without-deps
11M	lambda-with-deps.zip
7,0M	lambda-without-deps.zip
```

### DIFF
--- a/.changeset/beige-ways-switch.md
+++ b/.changeset/beige-ways-switch.md
@@ -4,7 +4,7 @@
 
 chore: Exclude more packages
 
-Added a debug to determine which packages that gets excluded from the final bundle's `node_modules`. Will skip these packages now aswell:
+Will skip these packages now aswell:
 
 - "typescript"
 - "next/dist/compiled/babel"


packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -39,10 +39,14 @@ const EXCLUDED_PACKAGES = [
...
+  "typescript",
+  "next/dist/compiled/babel",
+  "next/dist/compiled/babel-packages",
+  "next/dist/compiled/amphtml-validator",
 ];
 
-function isExcluded(srcPath: string) {
-  return EXCLUDED_PACKAGES.some((excluded) =>
+function isExcluded(srcPath: string): string | undefined {
+  return EXCLUDED_PACKAGES.find((excluded) =>
```
conico974: Why use find here ?

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -250,11 +254,17 @@ File ${serverPath} does not exist
 
   // Only files that are actually copied
   const tracedFiles: string[] = [];
-
+  // Packages that are excluded and not copied
+  const excludedPackages = new Set<string>();
```
conico974: Why a Set here ?

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -250,11 +254,17 @@ File ${serverPath} does not exist
...
   //Actually copy the files
   filesToCopy.forEach((to, from) => {
-    // We don't want to copy excluded packages (i.e sharp)
-    if (isExcluded(from)) {
+    // We don't want to copy excluded packages (e.g. sharp)
+    const excluded = isExcluded(from);
+    if (excluded) {
+      if (excluded && !excludedPackages.has(excluded)) {
```
conico974: I really don't understand the logic. What was the issue with the isExcluded from before ?

packages/open-next/src/build/copyTracedFiles.ts
sommeeeer: To keep trac kof the packages that are excluded, so we don't `logger.debug()` them more than once.

packages/open-next/src/build/copyTracedFiles.ts
sommeeeer: I just wanted to add a debug for the packages that actually gets excluded.

packages/open-next/src/build/copyTracedFiles.ts
sommeeeer: Same reason. I wanted to keep track of the packages that gets excluded. So we can logger.debug them.

conico974: I don't think that's useful at all, or at least i don't see how. I'd rather have this documented in the docs

sommeeeer: Aight, i'll revert that and make a PR for the docs.

conico974 (approved):
Thanks!!

sommeeeer:
Thanks for the review Nico!

PR #952, opened 2025-08-07
fix: Add automatic response cleanup via AbortSignal

sommeeeer:
To make `request.signal.onabort` work in route handlers. Currently only works for `cloudflare-node`, `node` and `express-dev`. It will close the `OpenNextNodeResponse` we pass into `NextServer`'s request handler when the original response in the wrapper is closed. This will ensure that the AbortSignal associated with `request.signal` is properly triggered when a client disconnects during a streaming response.

I did try to get it to work with `awsLambda.streamifyResponse`. It looks like we get no event emitted on `responseStream` when a client disconnects. I even tried with a barebone Lambda like this:
```ts
exports.handler = awslambda.streamifyResponse(
    async (event, responseStream, context) => {
        responseStream.setContentType('text/event-stream')
        responseStream.on("close", ()=> {
          // this never gets emitted at any point of time
          console.log("responseStream closed");
        });
        for (let i = 0; i < 20; i++) {
            await sleep(1000);
            responseStream.write(`event: ping\n`);
            responseStream.write(`data: ${i}\n\n`);
        }
        responseStream.end();
    }
);

### DIFF
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -87,8 +87,8 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
 
     // We want to destroy this response when the original response is closed. (i.e when the client disconnects)
     // This is to support `request.signal.onabort` in route handlers
-    if (streamCreator?.onClose) {
-      streamCreator.onClose(() => {
+    if (streamCreator?.abortSignal) {
+      streamCreator.abortSignal.addEventListener("abort", () => {
         this.destroy();
       });
     }
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -85,6 +85,7 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
       this.statusCode = statusCode;
     }
 
+    // https://github.com/vercel/next.js/blob/ea08bf2/packages/next/src/server/web/spec-extension/adapters/next-request.ts#L46-L54
     // We want to destroy this response when the original response/request is closed. (i.e when the client disconnects)
     // This is to support `request.signal.onabort` in route handlers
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by sommeeeer** [comment]
> It's basically what you did, but using `AbortSignal` instead of our own implementation

Thanks for the review Nico! This was an excellent idea. 

I am now using an `AbortSignal` instead of a callback. However, to get this to work in cloudflare I had to add the initial `request.signal` you retrieve from `fetch` to the ALS store. The request.signal you retrieve in the handler [here](https://github.com/sommeeeer/opennextjs-aws/blob/d3e4b4820447cc7a1177689333be523f2d111731/packages/open-next/src/overrides/wrappers/cloudflare-node.ts#L16) will not be `===` to the one you retrieve in your workers `fetch`. It doesn't trigger any event when the client disconnects/aborts either. Im not sure why thats the case, but im gonna look into it later. For now this PR would require us to put the `signal` in `opennextjs-cloudflare` [here](https://github.com/opennextjs/opennextjs-cloudflare/blob/060a03e431f4133f7ba005dbba2ba07d2eaa5d82/packages/cloudflare/src/cli/templates/init.ts#L34)

Update: Instead we pass the original `AbortSignal` you retrieve in cloudflare to the wrapper.

File: packages/open-next/src/http/openNextResponse.ts [hunk 1]
```diff
@@ -84,6 +84,14 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
     ) {
       this.statusCode = statusCode;
     }
+
+    // We want to destroy this response when the original response is closed. (i.e when the client disconnects)
+    // This is to support `request.signal.onabort` in route handlers
+    if (streamCreator?.abortSignal) {
```
khuezy: Since we're already using the conditional chain, it might be cleaner just to do:
`streamCreator?.abortSignal?.addEventListener("abort", () => this.destroy());`

sommeeeer:
Thanks for the review Kheuzy!

PR #954, opened 2025-08-11
Add binary content type handling in cache interceptor

conico974:
Introduce support for handling binary content types in the cache interceptor, along with tests for route type retrieval from the cache.

### DIFF
--- a/packages/open-next/src/core/routing/cacheInterceptor.ts
+++ b/packages/open-next/src/core/routing/cacheInterceptor.ts
@@ -9,6 +9,7 @@ import { getTagsFromValue, hasBeenRevalidated } from "utils/cache";
 import { debug } from "../../adapters/logger";
 import { localizePath } from "./i18n";
 import { generateMessageGroupId } from "./queue";
+import { isBinaryContentType } from "utils/binary";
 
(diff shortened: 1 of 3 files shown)


packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -268,6 +269,31 @@ export async function cacheInterceptor(
...
+        case "route": {
+          const cacheControl = await computeCacheControl(
+            localizedPath,
+            cachedData.value.body,
+            host,
+            cachedData.value.revalidate,
+            cachedData.lastModified,
+          );
+
+          const isBinary = isBinaryContentType(
```
sommeeeer, line 281: If the header is `undefined` here isBinary will be true. Can it be undefined here?

conico974: How would that be true ? `isBinaryContentType("undefined")` should not return true

sommeeeer: Ah yeah your right. My apologies. That would indeed return `false`.

khuezy (approved):
LGTM!

PR #955, opened 2025-08-13
fix(build): Add 22.18.0 to affected nodeVersion in dereference symlink

sommeeeer:
`22.18.0` is also affected. Should we consider skipping the `.bin` folder? I cannot see a reason for having it in the bundle.

### DIFF
--- a/packages/open-next/src/build/installDeps.ts
+++ b/packages/open-next/src/build/installDeps.ts
@@ -7,6 +7,8 @@ import type { InstallOptions } from "types/open-next";
 
 import logger from "../logger.js";
 
+const AFFECTED_NODE_VERSIONS = ["22.17.0", "22.17.1", "22.18.0"];
+
 export function installDependencies(
   outputDir: string,
   installOptions?: InstallOptions,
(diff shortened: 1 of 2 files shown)


packages/open-next/src/build/installDeps.ts
```diff
@@ -7,7 +7,7 @@ import type { InstallOptions } from "types/open-next";
 
 import logger from "../logger.js";
 
-const AFFECTED_NODE_VERSIONS = ["22.17.0", "22.17.1", "22.18.0"];
+const AFFECTED_NODE_VERSIONS = ["v22.17.0", "v22.17.1", "v22.18.0"];
```
conico974: Why did you have to add the V here ?

sommeeeer: The `v` is included in `process.version`

sommeeeer: We can use `process.versions.node` instead if you prefer the list without `v`'s

conico974 (approved):
LGTM Thanks!

PR #959, opened 2025-08-26
fix(cacheInterceptor): route cache handling

vicb:
Thanks Nico for pointing the root cause [in this comment](https://github.com/opennextjs/opennextjs-cloudflare/pull/858#issuecomment-3223320787)


vicb:
Thanks for the review!

Issue #961, opened 2025-08-26
createAssets fails in case of broken JSON in fetch cache


grabmateusz:
There is an issue in code:
https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/build/createAssets.ts
Code like:
```
  Object.entries(cacheFilesPath).forEach(([cacheFilePath, files]) => {
    const cacheFileMeta = files.meta
      ? JSON.parse(fs.readFileSync(files.meta, "utf8"))
      : undefined;
  //...
  });
```
is failing because of unsafe JSON parsing - whenever Next.js creates a file in cache that is not valid JSON file, deployment of OpenNext fails with:
```
SyntaxError: Unexpected non-whitespace character after JSON at position 34335
    at JSON.parse (<anonymous>)
    at file:///app/node_modules/@opennextjs/aws/dist/build/createAssets.js:141:35
    at file:///app/node_modules/@opennextjs/aws/dist/build/helper.js:159:13
    at Array.forEach (<anonymous>)
    at Module.traverseFiles (file:///app/node_modules/@opennextjs/aws/dist/build/helper.js:151:51)
    at createCacheAssets (file:///app/node_modules/@opennextjs/aws/dist/build/createAssets.js:139:21)
    at build (file:///app/node_modules/@opennextjs/cloudflare/dist/cli/build/build.js:59:44)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async buildCommand (file:///app/node_modules/@opennextjs/cloudflare/dist/cli/commands/build.js:13:5)
```
(The fact that Next.js generates incorrect JSON file in data cache is other problem, but not a subject of this ticket).
The goal of this ticket is to request fix in the way JSON is being parsed in mentioned createAssets.ts file.

khuezy:
Interesting, do you have an example of the incorrect JSON? We can forward it to the vercel team for them to look at if that's not expected.

sommeeeer:
Yeah, and also what version of Next is this happening on? A reproduction would be awesome!

grabmateusz:
Hi, I will submit incorrect JSON example later when I reproduce the issue again (it is reproducible very often, but not in 100% cases).
Regarding Next.js version at the moment it's 15.3.4, but it has been reproducible on multiple previous 15.x versions.
Regarding reproduction steps I would have to create standalone repository, I will try to do that.
[...]

PR #963, opened 2025-08-26
fix: createAssets fails in case of broken JSON in fetch cache

grabmateusz:
MR closes https://github.com/opennextjs/opennextjs-aws/issues/961

### DIFF
--- a/packages/open-next/src/build/createAssets.ts
+++ b/packages/open-next/src/build/createAssets.ts
@@ -165,7 +165,7 @@ export function createCacheAssets(options: buildHelper.BuildOptions) {
     const cacheJson = files.json
       ? safeParseJsonFile(fs.readFileSync(files.json, "utf8"), cacheFilePath)
       : undefined;
-    if (!cacheFileMeta || !cacheJson) {
+    if ((files.meta && !cacheFileMeta) || (files.json && !cacheJson)) {
       logger.warn(`Skipping invalid cache file: ${cacheFilePath}`);
       return;
     }
--- a/packages/open-next/src/build/createAssets.ts
+++ b/packages/open-next/src/build/createAssets.ts
@@ -162,13 +162,18 @@ export function createCacheAssets(options: buildHelper.BuildOptions) {
     const cacheFileMeta = files.meta
       ? safeParseJsonFile(fs.readFileSync(files.meta, "utf8"), cacheFilePath)
       : undefined;
+    const cacheJson = files.json
+      ? safeParseJsonFile(fs.readFileSync(files.json, "utf8"), cacheFilePath)
+      : undefined;
+    if (!cacheFileMeta || !cacheJson) {
+      logger.warn(`Skipping invalid cache file: ${cacheFilePath}`);
+      return;
+    }
     const cacheFileContent = {
       type: files.body ? "route" : files.json ? "page" : "app",
       meta: cacheFileMeta,
       html: files.html ? fs.readFileSync(files.html, "utf8") : undefined,
-      json: files.json
-        ? safeParseJsonFile(fs.readFileSync(files.json, "utf8"), cacheFilePath)
-        : undefined,
+      json: cacheJson,
       rsc: files.rsc ? fs.readFileSync(files.rsc, "utf8") : undefined,
       body: files.body
         ? fs
--- a/packages/open-next/src/build/createAssets.ts
+++ b/packages/open-next/src/build/createAssets.ts
@@ -6,7 +6,7 @@ import logger from "../logger.js";
 import type { TagCacheMetaFile } from "../types/cache.js";
 import { isBinaryContentType } from "../utils/binary.js";
 import * as buildHelper from "./helper.js";
-import { safeJsonParse } from "utils/safe-json-parse.js";
+import { safeParseJsonFile } from "utils/safe-json-parse.js";
 
 /**
  * Copy the static assets to the output folder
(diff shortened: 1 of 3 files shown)


packages/open-next/src/build/createAssets.ts
```diff
@@ -159,14 +160,14 @@ export function createCacheAssets(options: buildHelper.BuildOptions) {
...
-      ? JSON.parse(fs.readFileSync(files.meta, "utf8"))
+      ? safeJsonParse(fs.readFileSync(files.meta, "utf8"))
       : undefined;
     const cacheFileContent = {
       type: files.body ? "route" : files.json ? "page" : "app",
       meta: cacheFileMeta,
       html: files.html ? fs.readFileSync(files.html, "utf8") : undefined,
       json: files.json
-        ? JSON.parse(fs.readFileSync(files.json, "utf8"))
+        ? safeJsonParse(fs.readFileSync(files.json, "utf8"))
```
khuezy: Would it make sense to not write the cacheFile if the JSON was invalid?

grabmateusz: It would make sense, however:
1. Source of the broken JSON being cached is in external project (next), which needs to be reproduced and handled there independently,
2. Does it make sense to break whole deployment if one of the fetch cache JSON files is corrupted?

khuezy: 1. I've pinged the Vercel team. I'll update in the main issue when I hear back.
2. I don't think so, it should hit the next server if the cache is missing... assuming the server doesn't somehow regenerate the wrong JSON? If it does, I'm not sure what will happen, probably throw an error.

grabmateusz: re1: Thank you :)
re2: That is correct, missing fetch cache will cause server to perform HTTP request once it will be needed.

khuezy: If  the cache exists but the meta/json is undefined, what would happen? I'm not sure. I'm assuming it will fail?

conico974: I agree with khuezy here, if one of them fail, we should not write this entry at all. It may kind of work, but you're in a very weird state

packages/open-next/src/utils/safe-json-parse.ts
```diff
@@ -0,0 +1,7 @@
+export function safeJsonParse<T = any>(input: string, fallback?: T): T | undefined {
+    try {
+      return JSON.parse(input);
+    } catch (err) {
+      return fallback;
```
conico974: We should log the broken entry

packages/open-next/src/build/createAssets.ts
```diff
@@ -159,14 +160,14 @@ export function createCacheAssets(options: buildHelper.BuildOptions) {
...
-      ? JSON.parse(fs.readFileSync(files.meta, "utf8"))
+      ? safeParseJsonFile(fs.readFileSync(files.meta, "utf8"), cacheFilePath)
       : undefined;
     const cacheFileContent = {
       type: files.body ? "route" : files.json ? "page" : "app",
       meta: cacheFileMeta,
       html: files.html ? fs.readFileSync(files.html, "utf8") : undefined,
       json: files.json
-        ? JSON.parse(fs.readFileSync(files.json, "utf8"))
+        ? safeParseJsonFile(fs.readFileSync(files.json, "utf8"), cacheFilePath)
```
khuezy: Sorry for not being clear, what we're trying to say is that we shouldn't write the path if `meta` and (or?) `json` are `undefined`.

eg
```
const fileJson = safeParseJsonFile(fs.readFileSync(files.json, "utf8"), cacheFilePath)
if (!fileJson) return
```

If the `json` was malformed, we don't want to push it to the bucket since `next-server` would find it but would likely error if it didn't have any valid content. If it were able to gracefully fallback, it would just be extra work.

I'm not sure what a valid state would be. Can a valid cache file have a `meta` but no `json`?

grabmateusz: No worries, all clear, PR adjusted.

packages/open-next/src/utils/safe-json-parse.ts
grabmateusz: Logging added:
Screenshot 2025-08-28 at 17 50 59

packages/open-next/src/build/createAssets.ts
grabmateusz: Thanks for the feedback, PR updated, please revisit. I will close conversations I consider resolved, but feel free to open them if you see topic unanswered.

packages/open-next/src/utils/safe-json-parse.ts
sommeeeer: Were you able to create a minimal reproduction BTW?

packages/open-next/src/build/createAssets.ts
```diff
@@ -159,15 +160,20 @@ export function createCacheAssets(options: buildHelper.BuildOptions) {
   // Generate cache file
   Object.entries(cacheFilesPath).forEach(([cacheFilePath, files]) => {
     const cacheFileMeta = files.meta
-      ? JSON.parse(fs.readFileSync(files.meta, "utf8"))
+      ? safeParseJsonFile(fs.readFileSync(files.meta, "utf8"), cacheFilePath)
       : undefined;
+    const cacheJson = files.json
+      ? safeParseJsonFile(fs.readFileSync(files.json, "utf8"), cacheFilePath)
+      : undefined;
+    if (!cacheFileMeta || !cacheJson) {
```
conico974: That is not correct meta does not exist for page router, doing that means that we are skipping any entry from page router.
You should do something like `if ((files.meta && !cacheFileMeta) || (files.json && !cacheJson))`

conico974 (changes requested):

packages/open-next/src/build/createAssets.ts
khuezy: Thanks for the catch.

grabmateusz: Thanks @conico974, I confirm issue originated in App Router app and was tested only there. I've applied requested fix, please rereview.

conico974 (approved):

conico974:
@grabmateusz Could you fix linting please. I'll merge once done

Issue #958, opened 2025-08-25
Update examples stack to use SST


khuezy:
https://github.com/opennextjs/opennextjs-aws/blob/main/examples/sst/stacks/AppRouter.ts#L4

We're currently using a custom CDK implementation (this was before ion released,) we should switch over to ion since that's what most people are using (coming from SST.)

ref: [sst nextjs docs](https://sst.dev/docs/start/aws/nextjs/)

khuezy:
@conico974 @sommeeeer thoughts?

sommeeeer:
I think we should do it. Should be pretty straight forward, however `app-pages-router` might need some changes. Currently its using splitted functions which is not supported in latest SST. After they changed the routing logic to be in a cloudfront function instead of cloudfront behaviors. From version `>=3.10` its not working anymore.

You can see the splitted function here: https://github.com/opennextjs/opennextjs-aws/blob/9a83eb754ca82f830b1cc06d2d1470cd19173e0b/examples/app-pages-router/open-next.config.ts#L3-L8

khuezy:
Was that intentional that the sst 3.10+ no longer supports splitted functions?

sommeeeer:
> Was that intentional that the sst 3.10+ no longer supports splitted functions?

Yeah,  or you can actually read the Discord [thread](https://discord.com/channels/983865673656705025/1352177768095420427/1352178927166160957) when it was released. I think its possible to make it work, but it would require some changes in SST. Especially `ssr-site.ts` and `nextjs.ts`

conico974:
I think if we switch to ion, we should rewrite the whole e2e stuff to run locally first. Ion doesn't support everything (splitted functions, lambda@edge, external middleware...), so i wouldn't move everything there either.

If we want to do that, we should do it properly so if I had to make a plan here, it would be : 
1. Make e2e work locally so that we can run CI for it on every PR (not only on main)
2. Move like one app to the default ion construct (we can use multiple config file)
3. Build new more complex construct for the other apps (something like here https://github.com/conico974/open-next-deploy)

PR #964, opened 2025-08-28
add(dev-overrides): In memory tagCache with nextMode

sommeeeer:
This is for #958. Our goal is to make E2E work locally
 
With this new override `revalidateTag` and `revalidatePath` will work out of the box running OpenNext locally. This override can be used to support composable caching when that becomes stable aswell. There will come more PRs to make E2E work locally.

### DIFF
--- a/packages/open-next/src/overrides/tagCache/fs-dev-nextMode.ts
+++ b/packages/open-next/src/overrides/tagCache/fs-dev-nextMode.ts
@@ -29,8 +29,8 @@ export default {
     }
 
     const hasRevalidatedTag = tags.some((tag) => {
-      const lastModifiedTime = tagsMap.get(tag);
-      return lastModifiedTime ? lastModifiedTime > (lastModified ?? 0) : false;
+      const tagRevalidatedAt = tagsMap.get(tag);
+      return tagRevalidatedAt ? tagRevalidatedAt > (lastModified ?? 0) : false;
     });
 
     debug("hasBeenRevalidated result:", hasRevalidatedTag);


packages/open-next/src/overrides/tagCache/fs-dev-nextMode.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    debug("hasBeenRevalidated result:", hasRevalidatedTag);
+    return hasRevalidatedTag;
+  },
+  writeTags: async (tagsToWrite: string[]) => {
+    if (
+      globalThis.openNextConfig.dangerous?.disableTagCache ||
+      tagsToWrite.length === 0
+    ) {
+      return Promise.resolve();
```
khuezy: Since the function is `async`, you can just return here and below

sommeeeer: I prefer `Promise.resolve()` to be explicit. However, I'll remove it right away!

sommeeeer:
Thanks for the review Kheuzy!

packages/open-next/src/overrides/tagCache/fs-dev-nextMode.ts
```diff
@@ -0,0 +1,88 @@
...
+    path: { S: buildKey(tag) },
+    tag: { S: buildKey(tag) },
+    revalidatedAt: { N: `${revalidatedAt ?? Date.now()}` },
+  };
+}
+
+export default {
+  name: "fs-dev-nextMode",
+  mode: "nextMode",
+  getLastRevalidated: async (tagsToCheck: string[]) => {
```
khuezy: I'm not familiar w/ the base spec for tag cache. I see `writeTags`, shouldn't there be a `readTags`?

sommeeeer: This is for the new `tagCache` mode called `nextMode`. It works a bit different, and also is necessary to make composable cache work when that becomes stable. 

`hasBeenRevalidated()` will check if any tags have been revalidated. You can see it in action here:
https://github.com/opennextjs/opennextjs-aws/blob/31c3740ce7e7058295f2e268b7da251f8afccf5f/packages/open-next/src/utils/cache.ts#L25-L27

khuezy: Which part of the code reads and load the tags that was written to the file system?

sommeeeer: https://github.com/opennextjs/opennextjs-aws/blob/31c3740/packages/open-next/src/build/createAssets.ts#L253-L269 - the `dynamodb-cache.json` gets written here during build. In our `fs-dev-nextMode` override we read that file on top level and put it in a `let tags = ....`

khuezy: cool thanks!

packages/open-next/src/overrides/tagCache/fs-dev-nextMode.ts
```diff
@@ -0,0 +1,88 @@
...
+    // Not supported for now
+    // TODO: Implement getLastRevalidated
+    return 0;
```
conico974: You should implement this one

packages/open-next/src/overrides/tagCache/fs-dev-nextMode.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import type { NextModeTagCache } from "types/overrides";
+import { getMonorepoRelativePath } from "utils/normalize-path";
+import { debug } from "../../adapters/logger";
+
+const tagFile = path.join(
+  getMonorepoRelativePath(),
+  "dynamodb-provider/dynamodb-cache.json",
+);
+const tagContent = fs.readFileSync(tagFile, "utf-8");
```
conico974: You don't need the file. The whole point of the `nextMode` is to not need to prepopulate the tag cache.
Pls remove that

packages/open-next/src/overrides/tagCache/fs-dev-nextMode.ts
```diff
@@ -0,0 +1,88 @@
...
+
+let tags = JSON.parse(tagContent) as {
+  tag: { S: string };
+  path: { S: string };
+  revalidatedAt: { N: string };
+}[];
+
+function buildKey(key: string) {
+  const { NEXT_BUILD_ID } = process.env;
+  return path.posix.join(NEXT_BUILD_ID ?? "", key);
```
conico974: Don't add the BUILD_ID, it's in memory anyway

packages/open-next/src/overrides/tagCache/fs-dev-nextMode.ts
```diff
@@ -0,0 +1,88 @@
...
+import { debug } from "../../adapters/logger";
+
+const tagFile = path.join(
+  getMonorepoRelativePath(),
+  "dynamodb-provider/dynamodb-cache.json",
+);
+const tagContent = fs.readFileSync(tagFile, "utf-8");
+
+let tags = JSON.parse(tagContent) as {
+  tag: { S: string };
```
conico974: You don't need this structure, just a `Map<string, number>`

packages/open-next/src/overrides/tagCache/fs-dev-nextMode.ts
conico974: Composable cache works with the original tag cache as well. And you don't need the original file

sommeeeer:
Thanks Nico! Was a much better idea to use a Map. This `tagCache` interface is way easier to use aswell!

PR #966, opened 2025-08-28
fix(dev-overrides): Make Image optimization work locally

sommeeeer:
This is the only thing we need I think to make it work locally. By not specifying an `arch`, `npm install` command will choose it for us. (i.e our system one)

PR #965, opened 2025-08-28
fix(express-dev): Disable x-powered-by header

sommeeeer:
For #958 

This is to make the E2E work when ran locally.

PR #968, opened 2025-08-29
chore(app-pages-router): Put proxy in front of the server functions

sommeeeer:
This PR will make `examples/app-pages-router` run locally. We put a proxy infront of the splitted server functions. Ended up using `express-http-proxy`. I also installed `tsx` so we can run that TypeScript file easily. 

Depends on https://github.com/opennextjs/opennextjs-aws/pull/964, https://github.com/opennextjs/opennextjs-aws/pull/965, https://github.com/opennextjs/opennextjs-aws/pull/966


sommeeeer:
Thanks Nico!

PR #972, opened 2025-09-01
refactor: `compileOpenNextConfig` now take a path to the config file

james-elicx:
per https://github.com/opennextjs/opennextjs-cloudflare/pull/862#pullrequestreview-3172299357

~Using this PR, the CLI will treat absolute paths as absolute, instead of always treating them as relative. Relative paths should not start with a `/`, i.e. you should use `./config-path.ts` instead of `/config-path.ts`.~

The behavior of the CLI is unchanged after this PR, that is:

```bash
open-next build --config-path /relative/path
```

will continue to treat `/relative/path` as relative to the `baseDir` == `process.cwd()`.

### DIFF
--- a/.changeset/three-oranges-whisper.md
+++ b/.changeset/three-oranges-whisper.md
@@ -1,7 +1,7 @@
 ---
-"@opennextjs/aws": minor
+"@opennextjs/aws": patch
 ---
 
-fix: support absolute paths for OpenNext config compilation utility
+refactor: `compileOpenNextConfig` now takes `openNextConfigPath` only and no more `baseDir`.
 
-`compileOpenNextConfig` will now support the ability to pass in either absolute or relative paths, instead of treating any input as a relative path.
+`openNextConfigPath` is now in line with fs APIs: it is either absolute or relative to the working directory (`cwd`).
--- a/packages/open-next/src/build/compileConfig.ts
+++ b/packages/open-next/src/build/compileConfig.ts
@@ -13,18 +13,18 @@ import { validateConfig } from "./validateConfig.js";
  *
  * The configuration is always compiled for Node.js and for the edge only if needed.
  *
- * @param sourcePath Absolute path to the configuration file.
+ * @param openNextConfigPath Path to the configuration file. Absolute or relative to cwd.
  * @param nodeExternals Coma separated list of Externals for the Node.js compilation.
  * @param compileEdge Force compiling for the edge runtime when true
  * @return The configuration and the build directory.
  */
 export async function compileOpenNextConfig(
-  sourcePath: string,
+  openNextConfigPath: string,
   { nodeExternals = "", compileEdge = false } = {},
 ) {
   const buildDir = fs.mkdtempSync(path.join(os.tmpdir(), "open-next-tmp"));
   let configPath = compileOpenNextConfigNode(
-    sourcePath,
+    openNextConfigPath,
     buildDir,
     nodeExternals.split(","),
   );
@@ -47,7 +47,11 @@ export async function compileOpenNextConfig(
     (config.middleware?.external && config.middleware.runtime !== "node") ||
     Object.values(config.functions || {}).some((fn) => fn.runtime === "edge");
   if (usesEdgeRuntime || compileEdge) {
-    compileOpenNextConfigEdge(sourcePath, buildDir, config.edgeExternals ?? []);
+    compileOpenNextConfigEdge(
+      openNextConfigPath,
+      buildDir,
+      config.edgeExternals ?? [],
+    );
   } else {
     // Skip compiling for the edge runtime.
     logger.debug(
@@ -58,8 +62,16 @@ export async function compileOpenNextConfig(
   return { config, buildDir };
 }
 
+/**
+ * Compiles the OpenNext configuration for Node.
+ *
+ * @param openNextConfigPath Path to the configuration file. Absolute or relative to cwd.
+ * @param outputDir Folder where to output the compiled config file (`open-next.config.mjs`).
+ * @param externals List of packages that should not be bundled.
+ * @return Path to the compiled config.
+ */
 export function compileOpenNextConfigNode(
-  sourcePath: string,
+  openNextConfigPath: string,
   outputDir: string,
   externals: string[],
 ) {
@@ -67,13 +79,13 @@ export function compileOpenNextConfigNode(
   logger.debug("Compiling open-next.config.ts for Node.", outputPath);
 
   //Check if open-next.config.ts exists
-  if (!fs.existsSync(sourcePath)) {
+  if (!fs.existsSync(openNextConfigPath)) {
     //Create a simple open-next.config.mjs file
     logger.debug("Cannot find open-next.config.ts. Using default config.");
     fs.writeFileSync(outputPath, "export default { default: { } };");
   } else {
     buildSync({
-      entryPoints: [sourcePath],
+      entryPoints: [openNextConfigPath],
       outfile: outputPath,
       bundle: true,
       format: "esm",
@@ -94,8 +106,16 @@ export function compileOpenNextConfigNode(
   return outputPath;
 }
 
+/**
+ * Compiles the OpenNext configuration for Edge.
+ *
+ * @param openNextConfigPath Path to the configuration file. Absolute or relative to cwd.
+ * @param outputDir Folder where to output the compiled config file (`open-next.config.edge.mjs`).
+ * @param externals List of packages that should not be bundled.
+ * @return Path to the compiled config.
+ */
 export function compileOpenNextConfigEdge(
-  sourcePath: string,
+  openNextConfigPath: string,
   outputDir: string,
   externals: string[],
 ) {
@@ -103,7 +123,7 @@ export function compileOpenNextConfigEdge(
   logger.debug("Compiling open-next.config.ts for edge runtime.", outputPath);
 
   buildSync({
-    entryPoints: [sourcePath],
+    entryPoints: [openNextConfigPath],
     outfile: outputPath,
     bundle: true,
     format: "esm",
@@ -116,4 +136,6 @@ export function compileOpenNextConfigEdge(
       "process.env.NODE_ENV": '"production"',
     },
   });
+
+  return outputPath;
 }


vicb (approved):
LGTM, thanks!

sommeeeer (approved):
LGTM, we should probably update the [docs](https://opennext.js.org/aws/config#build-arguments) to address this change once merged. I did test this with an absolute path: 

 ```bash
 # This did work :)
node /my/open-next/dist/index.js --config-path ~/hi/open-next.config.ts
```

**Comment by vicb** [comment]
> LGTM, we should probably update the [docs](https://opennext.js.org/aws/config#build-arguments) to address this change once merged. I did test this with an absolute path:
> 
> ```shell
> # This did work :)
> node /home/user/open-next/dist/index.js build --config-path ~/hi/open-next.config.ts
> ```

A couple comments:

I didn't realize earlier that aws supports `--config-path`.

Would the current PR be a change in behavior, i.e. `build --config-path /path/to` would have been considered a relative path before but now requires `build --config-path ./path/to`.

If that is the case this is a change in behavior. I believe we can call this a "fix" rather than a breaking change but it would be nice to add details in the changeset.

Second comment is that we can use `--config-path` in cloudflare too, WDYT @james-elicx ?
I don't have a strong preference on this one. On one hand this would make both the CLI consistent but on the other hand cloudflare already uses `--config` to point to a wrangler config.
So I'm fine either way but probably something we can discuss before merging the cloudflare PR.

File: .changeset/three-oranges-whisper.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": minor
+---
+
+feat: support absolute paths when compiling opennext configs
```
james-elicx:
```suggestion
fix: support absolute paths for OpenNext configs

Compiling OpenNext configs will now support the ability to pass in either absolute or relative paths, instead of treating any input as a relative path.

Further, the CLI will now treat absolute paths as absolute, instead of always treating them as relative. Relative paths should not start with a `/`, i.e. you should use `./config-path.ts` instead of `/config-path.ts`.
```

james-elicx:
> > LGTM, we should probably update the [docs](https://opennext.js.org/aws/config#build-arguments) to address this change once merged. I did test this with an absolute path:
> > ```shell
> > # This did work :)
> > node /home/user/open-next/dist/index.js build --config-path ~/hi/open-next.config.ts
> > ```
> 
> A couple comments:
> 
> I didn't realize earlier that aws supports `--config-path`.
> 
> Would the current PR be a change in behavior, i.e. `build --config-path /path/to` would have been considered a relative path before but now requires `build --config-path ./path/to`.
> 
> If that is the case this is a change in behavior. I believe we can call this a "fix" rather than a breaking change but it would be nice to add details in the changeset.

Good point. I've updated the changeset. Alternatively, we could remove the support for passing in absolute paths in the AWS CLI and leave it until they do a future breaking change?

> Second comment is that we can use `--config-path` in cloudflare too, WDYT @james-elicx ? I don't have a strong preference on this one. On one hand this would make both the CLI consistent but on the other hand cloudflare already uses `--config` to point to a wrangler config. So I'm fine either way but probably something we can discuss before merging the cloudflare PR.

That would be a breaking change because we use the config path flag for the Wrangler config as part of allowing users to pass any wrangler flags to the cli.

vicb:
> Good point. I've updated the changeset. Alternatively, we could remove the support for passing in absolute paths in the AWS CLI and leave it until they do a future breaking change?

@conico974 and others, WDYT?

> That would be a breaking change because we use the config path flag for the Wrangler config as part of allowing users to pass any wrangler flags to the cli.

I think it's `-c` / `--config` as I wrote earlier though? 
i.e. it would not be a breaking change but `--config` vs `--config-path` could be confusing (the first one being the wrangler config and the second one the ON config)

james-elicx:
> > That would be a breaking change because we use the config path flag for the Wrangler config as part of allowing users to pass any wrangler flags to the cli.
> 
> I think it's `-c` / `--config` as I wrote earlier though? i.e. it would not be a breaking change but `--config` vs `--config-path` could be confusing (the first one being the wrangler config and the second one the ON config)

Ah yes, I see, Wrangler's is `config` but we also gave `configPath` as the flag name in cloudflare, with `config` and `c` as aliases 🤦‍♂️. So that would be a breaking change still in cloudflare.

sommeeeer:
> Would the current PR be a change in behavior, i.e. `build --config-path /path/to` would have been considered a relative path before but now requires `build --config-path ./path/to`.

Ah, good catch! I confirmed this is the case.

vicb:
> Wrangler's is `config` but we also gave `configPath` as the flag name in cloudflare, with `config` and `c` as aliases 🤦‍♂️. So that would be a breaking change still in cloudflare.

@james-elicx What about:
- rename the arg to `config` with a single `c` alias (instead of `configPath`)
- add a "new" arg `configPath`

The new behavior would be:
- error early when both flags are set
- use either `config` or `configPath` when only one of them is set
- print a deprecation warning when `configPath` is used
- remove the deprecated `configPath` in the next major release

vicb (changes requested):
@james-elicx I have updated the PR description to describe the breaking change in this PR:

```bash
open-next build --config-path /relative/to/baseDir
```

will not treat `/relative/to/baseDir` as an absolute path (you would have to use `./relative/to/baseDir`)

I think this is still correct?

If it is, we probably do not want that breaking outside of a major aws release

james-elicx:
> Alternatively, we could remove the support for passing in absolute paths in the AWS CLI and leave it until they do a future breaking change?

Yes, that's why I asked if we should remove that part of the change until AWS does a major release. I have no problem removing that if people prefer not to do it as part of this

vicb:
> Yes, that's why I asked if we should remove that part of the change until AWS does a major release. I have no problem removing that if people prefer not to do it as part of this

We should not introduce a breaking change in aws, confirmed offline with @conico974.

For Cloudflare we have to support both absolute (`/...`) and relative (`./...`) paths - that how `wrangler --c(onfig)` works and we don't want a different behavior for the Open Next config flag - that is consistency with the Cloudflare CLI is more important than consistency across aws and clouflare.

Ideally this PR can be updated to allow cf CLI support for both absolute and relative paths while not changing the behavior of aws CLI (always relative).

james-elicx:
Thanks for confirming. I've removed that part of the change.

packages/open-next/src/build/compileConfig.ts
```diff
@@ -13,22 +13,15 @@ import { validateConfig } from "./validateConfig.js";
  *
  * The configuration is always compiled for Node.js and for the edge only if needed.
  *
- * @param baseDir Directory where to look for the configuration.
- * @param openNextConfigPath Override the default configuration when provided. Relative to baseDir.
+ * @param sourcePath Path to the configuration file.
```
vicb: I think we can add what is supported amongst:
- absolute paths
- relative paths - if they are supported, what is the base path (`cwd`, app root directory, ...) ?

.changeset/three-oranges-whisper.md
```diff
@@ -0,0 +1,7 @@
+---
+"@opennextjs/aws": minor
+---
+
+fix: support absolute paths for OpenNext config compilation utility
+
+Compiling OpenNext configs will now support the ability to pass in either absolute or relative paths, instead of treating any input as a relative path.
```
vicb:
```suggestion
`compileOpenNextConfig` will now support the ability to pass in either absolute or relative paths, instead of treating any input as a relative path.
```

.changeset/three-oranges-whisper.md
```diff
@@ -0,0 +1,7 @@
+---
+"@opennextjs/aws": minor
```
vicb: I think we usually treat internal API changes as patch changes.

I'm fine with either but maybe @conico974 has a stronger opinion?

packages/open-next/src/build/compileConfig.ts
james-elicx: Is this okay? How relative paths are treated will depend on how esbuild and existsSync choose to interpret them.

```suggestion
* @param sourcePath Path to the configuration file. Either absolute or relative to the current directory.
```

vicb: > Is this okay?

If aws always uses an absolute path (sorry no time to check right now, need to run) then it should be ok to only accept absolute path - and we would make sure to also pass an absolute path from the cloudflare adapter.

packages/open-next/src/build.ts
```diff
@@ -32,8 +33,7 @@ export async function build(
   const openNextDistDir = url.fileURLToPath(new URL(".", import.meta.url));
 
   const { config, buildDir } = await compileOpenNextConfig(
-    baseDir,
-    openNextConfigPath,
+    path.join(baseDir, openNextConfigPath ?? "open-next.config.ts"),
```
vicb, line 36: @james-elicx `baseDir` is retrieved above on l 32:

```ts
const baseDir = process.cwd();
```

This means that `compileOpenNextConfig` is always called with an absolute path (`path.join(baseDir, openNextConfigPath ?? "open-next.config.ts"`).

So I think we shuold enforce and document that in `compileOpenNextConfig`.

packages/open-next/src/build/compileConfig.ts
james-elicx:
```suggestion
 * @param sourcePath Absolute path to the configuration file.
```

packages/open-next/src/build.ts
james-elicx: updated the jsdoc comment

vicb:
@james-elicx I have added a proposal fixup to this PR, might help speed up merging this - or not :)

The idea is that:
- The CLI behavior is not changed in `open-next build --config-path /relative/path`, `/relative/path` is still relative to the baseDir
- The *internal* `compileOpenNextConfig()` takes `openNextConfigPath` instead of `openNextConfigPath` and `baseDir`. `openNextConfigPath` is actually more akin to what's we usually call a path, that is absolute when starting with a `/`.

This change is a patch/refactor because only an internal API is not backward compatible.

Minor changes also in this proposal:
- more JS doc
- make `compileOpenNextConfigEdge` return the path to the config file to be consistent with the Node conterpart.

What do you think about the proposal?

conico974 (approved):
Sorry guys i took so long to respond here. I was convinced I did :smile:
LGTM, I definitely didn't want to introduce a breaking change here

james-elicx:
> @james-elicx I have added a proposal fixup to this PR, might help speed up merging this - or not :)
> 
> The idea is that:
> 
> * The CLI behavior is not changed in `open-next build --config-path /relative/path`, `/relative/path` is still relative to the baseDir
> * The _internal_ `compileOpenNextConfig()` takes `openNextConfigPath` instead of `openNextConfigPath` and `baseDir`. `openNextConfigPath` is actually more akin to what's we usually call a path, that is absolute when starting with a `/`.
> 
> This change is a patch/refactor because only an internal API is not backward compatible.
> 
> Minor changes also in this proposal:
> 
> * more JS doc
> * make `compileOpenNextConfigEdge` return the path to the config file to be consistent with the Node conterpart.
> 
> What do you think about the proposal?

Looks like they were just nitpicks about variable names and jsdoc, so looks good to me.

vicb:
> Looks like they were just nitpicks about variable names and jsdoc, so looks good to me.

Right!

Thanks for your work on this PR!

vicb (approved):

PR #974, opened 2025-09-03
fix: Ensure middleware rewrite status code is properly propagated in cache interceptor

sommeeeer:
When you have a rewrite in middleware like this:
```ts
return NextResponse.rewrite(u, {
  status: 403,
});
```
The rewrite status code is not propogated to the cache interceptor's responses when you have an external middleware. 

We can also safely remove the `x-next-cache-tags` header on the response.

I did update some tests that were flakey aswell. We need to register the event listener before the navigation.

### DIFF
--- a/packages/open-next/src/adapters/middleware.ts
+++ b/packages/open-next/src/adapters/middleware.ts
@@ -125,7 +125,10 @@ const defaultHandler = async (
         }
       }
 
-      result.headers[INTERNAL_EVENT_REQUEST_ID] = requestId;
+      if (process.env.OPEN_NEXT_REQUEST_ID_HEADER || globalThis.openNextDebug) {
+        result.headers[INTERNAL_EVENT_REQUEST_ID] = requestId;
+      }
+
       debug("Middleware response", result);
       return result;
     },
--- a/packages/open-next/src/types/open-next.ts
+++ b/packages/open-next/src/types/open-next.ts
@@ -32,6 +32,12 @@ export type InternalEvent = {
   readonly remoteAddress: string;
 } & BaseEventOrResult<"core">;
 
+export type MiddlewareEvent = InternalEvent & {
+  responseHeaders?: Record<string, string | string[]>;
+  isExternalRewrite?: boolean;
+  rewriteStatusCode?: number;
+};
+
 export type InternalResult = {
   statusCode: number;
   headers: Record<string, string | string[]>;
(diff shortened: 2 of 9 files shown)


packages/open-next/src/adapters/middleware.ts
```diff
@@ -125,7 +125,13 @@ const defaultHandler = async (
         }
       }
 
-      result.headers[INTERNAL_EVENT_REQUEST_ID] = requestId;
+      if (process.env.OPEN_NEXT_REQUEST_ID_HEADER || globalThis.openNextDebug) {
+        result.headers[INTERNAL_EVENT_REQUEST_ID] = requestId;
+      }
+
+      // Can safely delete this header here before returning as it is only used for the cache tags internally
+      delete result.headers["x-next-cache-tags"];
```
conico974: This is not the correct place to do that. It is only set by us in the cache interceptor, that's where we should not add it.

packages/open-next/src/types/open-next.ts
```diff
@@ -30,6 +30,8 @@ export type InternalEvent = {
   readonly query: Record<string, string | string[]>;
   readonly cookies: Record<string, string>;
   readonly remoteAddress: string;
+  // Used for NextResponse.rewrite's status code in middleware
+  readonly rewriteStatusCode?: number;
```
conico974: Why do we have to add this ?

packages/tests-e2e/tests/appRouter/middleware.rewrite.test.ts
```diff
@@ -23,21 +23,21 @@ test("Middleware Rewrite", async ({ page }) => {
...
 test("Middleware Rewrite External Image", async ({ page }) => {
-  await page.goto("/rewrite-external");
   page.on("response", async (response) => {
     expect(response.status()).toBe(200);
     expect(response.headers()["content-type"]).toBe("image/png");
     expect(response.headers()["cache-control"]).toBe("max-age=600");
     const bodyBuffer = await response.body();
     expect(validateMd5(bodyBuffer, OPENNEXT_PNG_MD5)).toBe(true);
   });
+  await page.goto("/rewrite-external");
```
conico974, line 33: Why change the order here ?

sommeeeer: Had to register the event listener before the navigation cause else it would never run the callback. The test was wrong in the first place.

packages/open-next/src/types/open-next.ts
sommeeeer: For the `event` passed into `cacheInterceptor`.

conico974: `cacheInterceptor` is already after the middleware, it should just take a `MiddlewareEvent`

sommeeeer:
Alright, thanks for the review Nico! This one should be ready to go now :)

conico974 (approved):

PR #975, opened 2025-09-03
fix(build): Refactor fallback values to be dummy overrides in createEdgeBundle

sommeeeer:
If you have `enableCacheInterception` enabled `includeCache` will be true there, meaning `esbuild` will potentially bundle `aws4fetch` and the fallback overrides(`sqs-lite`, `dynamodb-lite` & `s3-lite`) in the middleware handler file. This can happen if you build from the Cloudflare adapter cause the `overrides` can contain anonymous functions and the `override` function would return `undefined`.

To see this your self, build the `examples/e2e/app-router` in cloudflare repo and look for `aws4fetch`, `sqs-lite` etc.. in the middleware handler file. Do we really need to pass any values [here](https://github.com/sommeeeer/opennextjs-aws/blob/0e962f1a006449df2d7444ed021a9b6b9a287fd2/packages/open-next/src/build/edge/createEdgeBundle.ts#L86-L91) tho? I can't find any of them bundled in the middleware handler, even with this fix.

```bash
# before
124K    .open-next/middleware/handler.mjs
# after
104K    .open-next/middleware/handler.mjs
```

### DIFF
--- a/packages/open-next/src/build/edge/createEdgeBundle.ts
+++ b/packages/open-next/src/build/edge/createEdgeBundle.ts
@@ -59,7 +59,8 @@ export async function buildEdgeBundle({
 }: BuildEdgeBundleOptions) {
   const isInCloudflare = await isEdgeRuntime(overrides);
   function override<T extends keyof Override>(target: T) {
-    return typeof overrides?.[target] === "string"
+    return typeof overrides?.[target] === "string" ||
+      typeof overrides?.[target] === "function"
       ? overrides[target]
       : undefined;
   }
--- a/packages/open-next/src/build/edge/createEdgeBundle.ts
+++ b/packages/open-next/src/build/edge/createEdgeBundle.ts
@@ -59,8 +59,7 @@ export async function buildEdgeBundle({
 }: BuildEdgeBundleOptions) {
   const isInCloudflare = await isEdgeRuntime(overrides);
   function override<T extends keyof Override>(target: T) {
-    return typeof overrides?.[target] === "string" ||
-      typeof overrides?.[target] === "function"
+    return typeof overrides?.[target] === "string"
       ? overrides[target]
       : undefined;
   }
@@ -83,9 +82,9 @@ export async function buildEdgeBundle({
             converter: override("converter") ?? defaultConverter,
             ...(includeCache
               ? {
-                  tagCache: override("tagCache") ?? "dynamodb-lite",
-                  incrementalCache: override("incrementalCache") ?? "s3-lite",
-                  queue: override("queue") ?? "sqs-lite",
+                  tagCache: override("tagCache") ?? "dummy",
+                  incrementalCache: override("incrementalCache") ?? "dummy",
+                  queue: override("queue") ?? "dummy",
                 }
               : {}),
             originResolver: override("originResolver") ?? "pattern-env",
--- a/packages/open-next/src/build/edge/createEdgeBundle.ts
+++ b/packages/open-next/src/build/edge/createEdgeBundle.ts
@@ -59,7 +59,8 @@ export async function buildEdgeBundle({
 }: BuildEdgeBundleOptions) {
   const isInCloudflare = await isEdgeRuntime(overrides);
   function override<T extends keyof Override>(target: T) {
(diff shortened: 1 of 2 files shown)


packages/open-next/src/build/edge/createEdgeBundle.ts
```diff
@@ -59,7 +59,8 @@ export async function buildEdgeBundle({
 }: BuildEdgeBundleOptions) {
   const isInCloudflare = await isEdgeRuntime(overrides);
   function override<T extends keyof Override>(target: T) {
-    return typeof overrides?.[target] === "string"
+    return typeof overrides?.[target] === "string" ||
+      typeof overrides?.[target] === "function"
```
conico974, line 63: Not sure i understand what you mean here. `typeof (() => whatever)` is a function

sommeeeer: The `overrides` object can contain functions instead of strings when you build from Cloudflare adapter. This means that the [`override`](https://github.com/opennextjs/opennextjs-aws/blob/ba02f6b0d393e42ab232c26a6af0d5e18a365ba2/packages/open-next/src/build/edge/createEdgeBundle.ts#L62-L67) function will return `undefined` for those values:
```ts
overrides: {
    wrapper: 'cloudflare-edge',
    converter: 'edge',
    proxyExternalRequest: 'fetch',
    incrementalCache: [Function (anonymous)],
    tagCache: [Function (anonymous)],
    queue: [Function (anonymous)],
    originResolver: undefined
}
```

If you also have `enableCacheInterception` set as `true`, `includeCache` in `buildEdgeBundle` is true, so [these](https://github.com/opennextjs/opennextjs-aws/blob/ba02f6b0d393e42ab232c26a6af0d5e18a365ba2/packages/open-next/src/build/edge/createEdgeBundle.ts#L85-L87) lines will all be the fallback values cause `override("tagCache")` would return `undefined` in that case. Thus, it ends up bundling the `*-lite` overrides and also `aws4fetch` which is imported there.

packages/open-next/src/build/edge/createEdgeBundle.ts
```diff
@@ -82,9 +82,9 @@ export async function buildEdgeBundle({
             converter: override("converter") ?? defaultConverter,
             ...(includeCache
               ? {
-                  tagCache: override("tagCache") ?? "dynamodb-lite",
-                  incrementalCache: override("incrementalCache") ?? "s3-lite",
-                  queue: override("queue") ?? "sqs-lite",
+                  tagCache: override("tagCache") ?? "dummy",
```
conico974: This is not what we want to do at all. 
We should fallback to dummy **only** when tagCache is a function (same for the others). When nothing is provided, we should still use the current one

conico974 (changes requested):

packages/open-next/src/build/edge/createEdgeBundle.ts
sommeeeer: Are you fine with doing just this:
```ts
function override<T extends keyof Override>(target: T) {
  const value = overrides?.[target];
  return typeof value === "function"
    ? "dummy"
    : typeof value === "string"
      ? value
      : undefined;
}
```

Or you would you just prefer I'll do it seperately in the `includeCache` object spread

sommeeeer: Yeah, and sorry about the confusion. Of course we need to provide default values if your building from AWS

conico974: Yeah your solution is OK, but ideally it should  be on the resolve plugin that when we provide a function we always fallback to dummy

sommeeeer: Actually I think the PR was fine as it was. We already do what you described ^

https://github.com/sommeeeer/opennextjs-aws/blob/dc0579bf/packages/open-next/src/plugins/resolve.ts#L36-L44 - this function in resolve will return dummy for functions as long as we make sure to also send the function as overrideValue into openNextResolvePlugin from createEdgeBundle like we do in this PR now.

PR #977, opened 2025-09-05
fix: Avoid merging Location header on response when its an array

sommeeeer:
This is for https://github.com/opennextjs/opennextjs-cloudflare/issues/875#issuecomment-3258248276

Next will return a header like `Location: ["/target", "/target"];` on the response for a `redirect("/target");` in a static `page.tsx` when you dont have an `incrementalCache`. We should not merge that header's value with `join(",");` 

Affected line in Next: https://github.com/vercel/next.js/blob/ea08bf27/packages/next/src/server/base-server.ts#L1521

This is me `console.log(res._res.headers)` before and after that line with a reproduction:
image

### DIFF
--- a/packages/open-next/src/http/util.ts
+++ b/packages/open-next/src/http/util.ts
@@ -12,7 +12,8 @@ export const parseHeaders = (
     if (value === undefined) {
       continue;
     }
-    // Next can sometimes return an Array for the Location header
+    // Next can return an Array for the Location header
+    // We dont want to merge that into a comma-separated string
     // See: https://github.com/opennextjs/opennextjs-cloudflare/issues/875#issuecomment-3258248276
     if (key.toLowerCase() === "location" && Array.isArray(value)) {
       result[key.toLowerCase()] = value[0];
--- a/packages/open-next/src/http/util.ts
+++ b/packages/open-next/src/http/util.ts
@@ -15,7 +15,7 @@ export const parseHeaders = (
     }
     const keyLower = key.toLowerCase();
     /**
-     * Next can return an Array for the Location header
+     * Next can return an Array for the Location header when you return null from a get in the cacheHandler on a page that has a redirect()
      * We dont want to merge that into a comma-separated string
      * If they are the same just return one of them
      * Otherwise return the last one
--- a/packages/open-next/src/http/util.ts
+++ b/packages/open-next/src/http/util.ts
@@ -1,4 +1,5 @@
 import type http from "node:http";
+import logger from "../logger";
 
 export const parseHeaders = (
   headers?: http.OutgoingHttpHeader[] | http.OutgoingHttpHeaders,
@@ -25,6 +26,9 @@ export const parseHeaders = (
       if (value[0] === value[1]) {
         result[keyLower] = value[0];
       } else {
+        logger.warn(
+          "Multiple different values for Location header found. Using the last one",
+        );
         result[keyLower] = value[value.length - 1];
       }
       continue;
--- a/packages/open-next/src/http/util.ts
+++ b/packages/open-next/src/http/util.ts
@@ -12,14 +12,24 @@ export const parseHeaders = (
     if (value === undefined) {
       continue;
     }
-    // Next can return an Array for the Location header
-    // We dont want to merge that into a comma-separated string
-    // See: https://github.com/opennextjs/opennextjs-cloudflare/issues/875#issuecomment-3258248276
-    if (key.toLowerCase() === "location" && Array.isArray(value)) {
-      result[key.toLowerCase()] = value[0];
+    const keyLower = key.toLowerCase();
+    /**
+     * Next can return an Array for the Location header
+     * We dont want to merge that into a comma-separated string
+     * If they are the same just return one of them
+     * Otherwise return the last one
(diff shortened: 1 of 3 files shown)


vicb:
Thanks for putting this PR together @sommeeeer !

A few questions as you have looked into this:

Do I understand correctly than Next doesn't really care about what's returned when there is no incremental cache... as they always have an incremental cache?

Should we rather error when we have static/cached content and no incremental cache?

If the header value is an array `[location1, location2]`, what are `location1` and `location2`? 
Are there always exactly 2 locations?
Are location1 and location2 always the same (i.e. why should we pick `[0]`)

sommeeeer:
> Do I understand correctly than Next doesn't really care about what's returned when there is no incremental cache... as they always have an incremental cache?

Yeah, or they should SSR that page then, but for some reason in this case the Location's value is an array. Well, when I think more about this. Why would you ever want to setup an OpenNext/Next project without ISR. You could just use the `static` export then. Perhaps I should close this PR? WDYT?

> Are location1 and location2 always the same (i.e. why should we pick `[0]`)

Yeah, they are always the same. I guess somewhere in Next's code they put that header value twice when  `get` returns null from the `cacheHandler`.

vicb:
> > Do I understand correctly than Next doesn't really care about what's returned when there is no incremental cache... as they always have an incremental cache?
> 
> Yeah, or they should SSR that page then, but for some reason in this case the Location's value is an array. Well, when I think more about this. Why would you ever want to setup an OpenNext/Next project without ISR. You could just use the `static` export then. Perhaps I should close this PR? WDYT?
> 
> > Are location1 and location2 always the same (i.e. why should we pick `[0]`)
> 
> Yeah, they are always the same. I guess somewhere in Next's code they put that header value twice when `get` returns null from the `cacheHandler`.

Not sure what is the best way to handle this.

Maybe the Open Next build should fail when Next has cache artifacts and Open Next has no cache configured?
There should be no reason to force configuring cache for a fully dynamic app.

I think it's worth leaving this PR opened until we discuss more about it.

sommeeeer:
> Maybe the Open Next build should fail when Next has cache artifacts and Open Next has no cache configured?

Thats actually a great idea, however I think we should discuss this further as you mentioned. This decision can have consequences and needs to be thought about thoroughly.

I'll leave it open for now.

conico974:
> Maybe the Open Next build should fail when Next has cache artifacts and Open Next has no cache configured?

It should be a big warning not an error. If your app is fully SSR, it could make sense to have no Incremental cache.
And splitted function don't necessarily need the cache either.

This error could also happen if you have a cache configured, but it errors out. In this case we return null from the cache handler (same thing as with the dummy incremental cache) and we have this bug

As to what exactly to do here, before picking the value, we should check if they are the same, if it's the case we're fine.
If not, maybe pick the last one (logically the last one should be the latest added there), but we should log something here.

sommeeeer:
Thanks for the reviews! 

> As to what exactly to do here, before picking the value, we should check if they are the same, if it's the case we're fine. If not, maybe pick the last one (logically the last one should be the latest added there), but we should log something here.

Alright, I addressed this in a review commit.

packages/open-next/src/http/util.ts
```diff
@@ -12,7 +13,27 @@ export const parseHeaders = (
...
+     * Next can return an Array for the Location header
+     * We dont want to merge that into a comma-separated string
+     * If they are the same just return one of them
+     * Otherwise return the last one
+     * See: https://github.com/opennextjs/opennextjs-cloudflare/issues/875#issuecomment-3258248276
+     * and https://github.com/opennextjs/opennextjs-aws/pull/977#issuecomment-3261763114
+     */
+    if (keyLower === "location" && Array.isArray(value)) {
+      if (value[0] === value[1]) {
+        result[keyLower] = value[0];
```
vicb, line 27: are we sure they are always exactly 2 values in the array?

sommeeeer: > are we sure they are always exactly 2 values in the array?

Yeah, and it could only happen if it returns `null` from a get in the cacheHandler.

vicb: Maybe you can add the info in the JS doc ?

vicb (approved):

brunns:
Hi folks.

Since this was released, we are seeing these messages in our logs:
```json
{
    "timestamp": "2025-10-12T18:33:25.780Z",
    "level": "WARN",
    "requestId": "c3b84e73-52b0-4e6e-8d16-22a2040160e9",
    "message": "WARN Multiple different values for Location header found. Using the last one"
}
```

Any chance it's related? FWIW, our code is here: https://github.com/NHSDigital/vaccinations-app/blob/main/src/middleware.ts#L34

Thanks!

sommeeeer:
Hi @brunns 

Yeah that warning is related to this PR. However, I was not able to reproduce this with having a `NextResponse.redirect` like you have in your `src/middleware.ts`. If you `console.log({key, value});` inside the `else {` block above the `console.warn` what does it give you?

brunns:
Thanks @sommeeeer. I'll have a look around elsewhere in our codebase, see if I can find the issue.

donnabelsey:
We have found the source of the warning message; when the array is of length 1 a warning is printed by accident. PR raised #1016 to fix. Thanks.

PR #980, opened 2025-09-10
fix: prevent unnecessary flattening of headers during convertRes

cboar:
I ran into an issue where multiple `Set-Cookie` headers were being joined into a single header, which is not valid. I found that the `convertRes` function flattens all headers right before constructing the `InternalResult`. Since `InternalResult` allows `string[]` headers, I rewrote the `parseHeaders` to keep these as is. Since the method is also used elsewhere, I renamed the original implementation to `flattenHeaders` to better reflect its behavior.


cboar:
@sommeeeer I see #977 solves a similar issue exclusively for `Location`, but this also seems to affect `Set-Cookie` when multiple cookies are set.

For context, I am using a custom wrapper built for AWS ALB responses (not API Gateway), it has been helpful to not join the headers returned from InternalResult.

Issue #973, opened 2025-09-02
Headers from next.config override headers defined in a middleware


dirkdev98:
Setting default response headers in the next.config are not overwritten by response headers set in a middleware. Based on the [docs](https://nextjs.org/docs/app/api-reference/file-conventions/middleware#execution-order) and previous deployments in a standalone builds I'd expect this to work. 

I think the offending lines in OpenNext are here: 

https://github.com/opennextjs/opennextjs-aws/blob/fe913bb8dc09b4b9eb782d28b7b5d52ad63377b6/packages/open-next/src/core/routingHandler.ts#L132-L135

I could send a PR, however I think this is a breaking change for existing users, so might not be ideal to land in v3?

sommeeeer:
You have pointed out the correct place. The order is wrong there. I ran the E2E on the change, and it did pass 100%, however you might be right that there might be people out there relying on this. If we decide to make a PR for this, we should probably do it in a minor (or even a major). Until then you can just use `pnpm patch` or similiar if you need it.

dirkdev98:
Thanks for confirming. Yes this ain't blocking, enough ways to work around it!

Let me know if you want this fixed in a minor, if so I can get a PR going.

sommeeeer:
> Let me know if you want this fixed in a minor, if so I can get a PR going.

That decision is up to Kheuzy or Conico. We'll have to wait and see what they want. Thanks for your research and willingnesss

khuezy:
I'll have to defer to Conico here since, as I vaguely recall, he worked on the headers ordering per Nextjs' docs. Was that order always like that or did they change it?
If this is originally wrong, then we did a big whooops in the code review 😢

conico974:
I recall testing this back at the time and it was the other way around, I could be wrong though, it's been awhile.
This is definitely a breaking change, and something i'd put on a major. 
And since there is an easy workaround, i don't consider it an urgent matter. You could still create a PR and use the prerelease if you want, but we probably won't merge it right now.
We could also put it under a flag to enable this behavior (and this would not be a breaking change)

sommeeeer:
> We could also put it under a flag to enable this behavior (and this would not be a breaking change)

Thats actually an excellent idea.

khuezy:
> I recall testing this back at the time and it was the other way around

Same here, I recall us specifically discussing this somewhere (on github or on the discord)

dirkdev98:
[...]

PR #981, opened 2025-09-12
feat: add option to align header priorization execution order

dirkdev98:
Fixes opennextjs/opennextjs-aws#973

I'm not sure if the added configuration option should be under `dangerous`, but just put it there since it changes some behavior. Also alternative names are welcome :) 

Should I also create a PR to the docs repo, or is that auto-synced based on types / comment extraction?

### DIFF
--- a/examples/app-router/app/headers/execution-order/page.tsx
+++ b/examples/app-router/app/headers/execution-order/page.tsx
@@ -1,0 +1,9 @@
+export default async function HeadersExecutionOrder() {
+  return (
+    <div>
+      <h1>
+        Used to check execution order of middleware vs next.config headers.
+      </h1>
+    </div>
+  );
+}
--- a/examples/app-router/app/headers/execution-order/page.tsx
+++ b/examples/app-router/app/headers/execution-order/page.tsx
@@ -1,9 +1,0 @@
-export default async function HeadersExecutionOrder() {
-  return (
-    <div>
-      <h1>
-        Used to check execution order of middleware vs next.config headers.
-      </h1>
-    </div>
-  );
-}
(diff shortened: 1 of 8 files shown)


sommeeeer:
> Should I also create a PR to the docs repo, or is that auto-synced based on types / comment extraction?

No, and it would be nice to be added there.

dirkdev98:
>   Did you run the E2E?

Yep, the headers one specifically against a Next.js only build and the `appRouter` suite fully against a local OpenNext build. Had some failures on image-optimization which I haven't really looked at yet. But I assumed that's caused by only spinning up the server function.

I forgot to push the `open-next.config.local.ts` change, which I have added to the PR now.

examples/app-router/app/headers/execution-order/page.tsx
```diff
@@ -0,0 +1,9 @@
+export default async function HeadersExecutionOrder() {
```
conico974: Do we really need a route ? Can't we just return from the middleware ?

packages/open-next/src/types/open-next.ts
```diff
@@ -80,6 +80,16 @@ export interface DangerousOptions {
   headersAndCookiesPriority?: (
     event: InternalEvent,
   ) => "middleware" | "handler";
+
+  /**
+   * Configuration option to prioritize headers set via middleware over headers set via the option in the Next config.
+   * This brings OpenNext behavior inline with the documented execution order.
```
conico974: Execution order is not really the same thing either, if you set some headers on a route they will be overriden by the middleware, which does not follow the execution order either.

We should also add that the default will change in next major

examples/app-router/app/headers/execution-order/page.tsx
dirkdev98: Thanks for the review! Nope, moved it to a middleware response.

packages/open-next/src/types/open-next.ts
dirkdev98: Updated!

PR #982, opened 2025-09-18
fix: allow bypassing the Tag Cache in the interceptor

vicb:
`shouldBypassTagCache` was implemented in https://github.com/opennextjs/opennextjs-aws/pull/944.

It is useful to bypass the Tag Cache when we know that an entry has not been revalidated.

The original PR missed an updated in the cache interceptor resulting in the Tag Cache never being bypassed when it is in use.

/cc @dario-piotrowicz

### DIFF
--- a/.changeset/gold-penguins-walk.md
+++ b/.changeset/gold-penguins-walk.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": minor
+---
+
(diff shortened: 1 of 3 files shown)


conico974 (approved):

vicb:
I'm going to add a test before merging the PR

PR #985, opened 2025-09-24
Update npm install command to use --os flag

niklaswallerstedt:
Resolves a problem where sharp after 0.36.2 would not install correctly.

- uses `--os` instead of platform
- updates `sharp`to latest version

### DIFF
--- a/packages/open-next/src/build/createImageOptimizationBundle.ts
+++ b/packages/open-next/src/build/createImageOptimizationBundle.ts
@@ -110,7 +110,7 @@ export async function createImageOptimizationBundle(
   // Target should be same as used by Lambda, see https://github.com/sst/sst/blob/ca6f763fdfddd099ce2260202d0ce48c72e211ea/packages/sst/src/constructs/NextjsSite.ts#L114
   // For SHARP_IGNORE_GLOBAL_LIBVIPS see: https://github.com/lovell/sharp/blob/main/docs/install.md#aws-lambda
 
-  const sharpVersion = process.env.SHARP_VERSION ?? "0.32.6";
+  const sharpVersion = process.env.SHARP_VERSION ?? "0.34.4";
 
   installDependencies(
     outputPath,
--- a/packages/open-next/src/build/installDeps.ts
+++ b/packages/open-next/src/build/installDeps.ts
@@ -36,7 +36,7 @@ export function installDependencies(
       : "";
 
     const additionalArgs = installOptions.additionalArgs ?? "";
-    const installCommand = `npm install --platform=linux ${archOption} ${targetOption} ${libcOption} ${additionalArgs} ${installOptions.packages.join(" ")}`;
+    const installCommand = `npm install --os=linux ${archOption} ${targetOption} ${libcOption} ${additionalArgs} ${installOptions.packages.join(" ")}`;
     execSync(installCommand, {
       stdio: "pipe",
       cwd: tempInstallDir,
--- a/packages/open-next/src/types/open-next.ts
+++ b/packages/open-next/src/types/open-next.ts
@@ -279,7 +279,7 @@ export interface InstallOptions {
    * @example
    * ```ts
    * install: {
-   *  packages: ["sharp@0.32"]
+   *  packages: ["sharp@0.34"]
    * }
    * ```
    */


packages/open-next/src/build/installDeps.ts
```diff
@@ -36,7 +36,7 @@ export function installDependencies(
       : "";
 
     const additionalArgs = installOptions.additionalArgs ?? "";
-    const installCommand = `npm install --platform=linux ${archOption} ${targetOption} ${libcOption} ${additionalArgs} ${installOptions.packages.join(" ")}`;
+    const installCommand = `npm install --os=linux ${archOption} ${targetOption} ${libcOption} ${additionalArgs} ${installOptions.packages.join(" ")}`;
```
conico974, line 39: We should provide both if it does not cause any issue, not just replace it

.changeset/spicy-mangos-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": minor
```
conico974: It's a patch, not a minor

packages/open-next/src/build/createImageOptimizationBundle.ts
```diff
@@ -110,7 +110,7 @@ export async function createImageOptimizationBundle(
   // Target should be same as used by Lambda, see https://github.com/sst/sst/blob/ca6f763fdfddd099ce2260202d0ce48c72e211ea/packages/sst/src/constructs/NextjsSite.ts#L114
   // For SHARP_IGNORE_GLOBAL_LIBVIPS see: https://github.com/lovell/sharp/blob/main/docs/install.md#aws-lambda
 
-  const sharpVersion = process.env.SHARP_VERSION ?? "0.32.6";
+  const sharpVersion = process.env.SHARP_VERSION ?? "0.34.4";
```
conico974: Do we have to bump that one ? Did you test it against other versions of Next ?
OpenNext support Next 12, it has to work from Next 12 to 15

packages/open-next/src/build/installDeps.ts
vicb: @conico974 It looks like there is no `--platform` option. What about updating the code to use `--os=${os.platform()}`

It looks like it would help with local e2e on Mac

/cc @sommeeeer

packages/open-next/src/build/createImageOptimizationBundle.ts
vicb: We should probably revert this from this PR

.changeset/spicy-mangos-shop.md
vicb:
```suggestion
"@opennextjs/aws": patch
```

packages/open-next/src/build/installDeps.ts
vicb: See https://sharp.pixelplumbing.com/install/#npm-v10 (kudos to Magnus for finding that link)

sommeeeer: We could introduce another optional parameter to `InstallOptions` that is `os` and defaults to linux. Then [here](https://github.com/sommeeeer/opennextjs-aws/blob/8a70e257465710da0cc8e84d49de27a9bcec1cf7/packages/open-next/src/build/createImageOptimizationBundle.ts#L115-L123) we set it to what you suggested if the `config.imageOptimization.loader` is `fs-dev`. WDYT?

Update: However, does `glibc` work on macOS?

vicb: +1 to your idea @sommeeeer 

We could also detect the CPU (i.e. arm64 vs x64)

> Update: However, does glibc work on macOS?

I'll test and report.

conico974: I don't remember when we added the `--platform` options TBH, it's a very old one.
I'm fine with adding the `os` option.
Autodetecting is something that we need to be able to override though, this could break some deployment otherwise.

niklaswallerstedt: Hey, been busy with other things and haven't gotten round to this PR. But if I understand correctly, autodetecting would break my deployment since I usually deploy through GitHub actions but on rare occasions I've had to deploy from local to fix issues.

packages/open-next/src/build/createImageOptimizationBundle.ts
niklaswallerstedt: Ok, reverting!

packages/open-next/src/build/installDeps.ts
conico974: What they suggest is only if you use the `fs-dev` image loader (which is likely not your case).
But even in this case I think that if the user provide some options we should respect it

vicb:
The pre-releasing step fails... merging anyway

PR #988, opened 2025-10-01
fix: bypass next tag cache when there are no tags to check

vicb:
Before this PR `hasBeenRevalidated` might be called with an empty tag list.

However we can early return `false` in this case


conico974 (approved):

vicb:
Thanks for the review Nico!

vicb:
Oops, I need to update the tests...

PR #991, opened 2025-10-07
improve performance of responses

anonrig:
For benchmarks that check TTFB, opennextjs was performing poorly. This was due to several reasons such as returning the body as a huge chunk with unnecessary copies. To summarize:

- Eliminated Buffer.concat() usage - to avoid copying the entire buffer into memory. From now on, we use ReadableStream.from() to stream chunks without copying.
- Eliminating Buffer.concat() on body length calculation - previously in order to get the body length, we were copying the entire buffer into memory (using Buffer.concat()) and disposing it after we calculated the length. This created huge GC pressure and degraded the performance.
- Chunk-by-chunk writing: Removed getBody() call that copied and concatenated all chunks into a single buffer before writing. From now on, we write each chunks individually - which reduces the TTFB.

I've also updated CI to use Node.js v20 since ReadableStream.from() requires it, but also v18 is EOL.

----

This improves the responses by roughly 20%. 

----

PS: AI took my trust in pull-request descriptions with lists - but I wrote it regardless...

### DIFF
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -106,11 +106,13 @@ export function convertRes(res: OpenNextNodeResponse): InternalResult {
     isBinaryContentType(headers["content-type"]) ||
     !!headers["content-encoding"];
   const body = new ReadableStream({
-    start(controller) {
-      for (const chunk of res._chunks ?? []) {
-        controller.enqueue(chunk);
+    pull(controller) {
+      if (!res._chunks || res._chunks.length === 0) {
+        controller.close();
+        return;
       }
-      controller.close();
+
+      controller.enqueue(res._chunks.shift());
     },
   });
   return {
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -105,14 +105,14 @@ export function convertRes(res: OpenNextNodeResponse): InternalResult {
   const isBase64Encoded =
     isBinaryContentType(headers["content-type"]) ||
     !!headers["content-encoding"];
-  const body =
-    res._chunks == null
-      ? new ReadableStream({
-          start(controller) {
-            controller.close();
-          },
-        })
-      : ReadableStream.from(res._chunks);
+  const body = new ReadableStream({
+    start(controller) {
+      for (const chunk of res._chunks ?? []) {
+        controller.enqueue(chunk);
+      }
+      controller.close();
+    },
+  });
   return {
     type: "core",
     statusCode,
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -1,7 +1,6 @@
 import crypto from "node:crypto";
 import type { OutgoingHttpHeaders } from "node:http";
 import { parse as parseQs, stringify as stringifyQs } from "node:querystring";
-import { Readable } from "node:stream";
 
 import { BuildId, HtmlPages, NextConfig } from "config/index.js";
 import type { IncomingMessage } from "http/index.js";
@@ -18,6 +17,7 @@ import type {
   StreamCreator,
 } from "types/open-next.js";
 
+import { ReadableStream } from "node:stream/web";
 import { debug, error } from "../../adapters/logger.js";
 import { isBinaryContentType } from "../../utils/binary.js";
 import { localizePath } from "./i18n/index.js";
@@ -105,9 +105,14 @@ export function convertRes(res: OpenNextNodeResponse): InternalResult {
   const isBase64Encoded =
     isBinaryContentType(headers["content-type"]) ||
     !!headers["content-encoding"];
-  // We cannot convert the OpenNextNodeResponse to a ReadableStream directly
-  // You can look in the `aws-lambda.ts` file for some context
-  const body = Readable.toWeb(Readable.from(res.getBody()));
+  const body =
+    res._chunks == null
+      ? new ReadableStream({
+          start(controller) {
+            controller.close();
+          },
+        })
+      : ReadableStream.from(res._chunks);
   return {
     type: "core",
     statusCode,
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -282,6 +282,14 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
     return Buffer.concat(this._chunks);
   }
 
+  getBodyLength(): number {
+    let size = 0;
+    for (const chunk of this._chunks) {
+      size += chunk.length;
+    }
+    return size;
+  }
+
   private _internalWrite(chunk: any, encoding: BufferEncoding) {
     this._chunks.push(Buffer.from(chunk, encoding));
     this.push(chunk, encoding);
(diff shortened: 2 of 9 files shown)


packages/open-next/src/core/routing/util.ts
```diff
@@ -105,14 +105,11 @@ export function convertRes(res: OpenNextNodeResponse): InternalResult {
...
     !!headers["content-encoding"];
-  // We cannot convert the OpenNextNodeResponse to a ReadableStream directly
-  // You can look in the `aws-lambda.ts` file for some context
-  const body = Readable.toWeb(Readable.from(res.getBody()));
   return {
     type: "core",
     statusCode,
     headers,
-    body,
+    body: Readable.from(res._chunks),
```
jasnell: Some explanation here is worthwhile: the current implementation of `res.getBody()` does a `Buffer.concat(res._chunks)`, which means the entire buffer is copied into memory, then passed into a Node.js `stream.Readable` that gets further wrapped in a `ReadableStream` adapter implementation... when it could just be `ReadableStream.from(res._chunks)`

Btw, @anonrig ... should this be `ReadableStream.from(res._chunks)`? `Readable.from(...)` will return a node `stream.Readable` which means this is getting a different type than the original.

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -310,7 +318,7 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
     globalThis.__openNextAls
       ?.getStore()
       ?.pendingPromiseRunner.add(this.onEnd(this.headers));
-    const bodyLength = this.getBody().length;
```
jasnell, line 313: Some explanation here: Since `getBody()` does a `Buffer.concat(this._chunks)`, the original code here is unnecessarily copying the entire buffer and throwing it away just to get the length.

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -366,8 +374,10 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
   }
 
   send() {
-    const body = this.getBody();
-    this.end(body);
+    for (const chunk of this._chunks) {
+      this.write(chunk);
+    }
+    this.end();
```
jasnell, line 380: some explanation here: the `getBody()` is doing a `Buffer.concat(_chunks)` which is then passed into the stream as a single write, which is unnecessary. We can just write out each individual chunk.

@anonrig ... did you make sure that `this.write(chunk)` is not just deferring to `_internalWrite`?

packages/open-next/src/core/routing/util.ts
anonrig: You're right but I'm seeing an error with the proposed change.

```
tests-unit:test: TypeError: Cannot read properties of undefined (reading 'Symbol(Symbol.asyncIterator)')
tests-unit:test:  ❯ Module.convertRes packages/open-next/src/core/routing/util.ts:112:26
tests-unit:test:     110|     statusCode,
tests-unit:test:     111|     headers,
tests-unit:test:     112|     body: ReadableStream.from(res._chunks),
tests-unit:test:        |                          ^
tests-unit:test:     113|     isBase64Encoded,
tests-unit:test:     114|   };
tests-unit:test:  ❯ packages/tests-unit/tests/core/routing/util.test.ts:242:20
```

jasnell: Ok, then it likely needs to be wrapped in an async generator...

```js
async function* gen() {
  for (const chunk of res._chunks) {
    yield chunk;
  }
}

ReadableStream.from(gen());
```

jasnell: Or check to make sure that `res._chunks` is not undefined here! If it is, pass back a closed `ReadableStream`

packages/open-next/src/http/openNextResponse.ts
anonrig: this class doesn't override write method.

mhart:
This is still buffering all chunks and delivering them as one though right?

We should definitely look at streaming without doing that (you only need to buffer until headers have been delivered, can stream after that).

Eg, using Next's MockResponse: https://github.com/vercel/next.js/blob/v15.3.0-canary.13/packages/next/src/server/lib/mock-request.ts and then just wiring up `resWriter` to the output stream: https://github.com/vercel/next.js/blob/a1c27450e97ec3985c32f763355f8a564cd562bd/packages/next/src/server/lib/mock-request.ts#L238-L240

(or using https://github.com/mhart/fetch-to-node or similar)

anonrig:
> This is still buffering all chunks and delivering them as one though right?

Yes, technically we shouldn't store them in memory and stream them directly without copying/storing.

jasnell:
well, splitting the `write(...)` and `end()` in the `send()` should at least (hopefully) stream it out a bit better but we need to test/verify further. These were just some low hanging fruit.

packages/open-next/src/core/routing/util.ts
```diff
@@ -105,9 +105,14 @@ export function convertRes(res: OpenNextNodeResponse): InternalResult {
...
-  // You can look in the `aws-lambda.ts` file for some context
-  const body = Readable.toWeb(Readable.from(res.getBody()));
+  const body =
+    res._chunks == null
+      ? new ReadableStream({
+          start(controller) {
+            controller.close();
+          },
+        })
+      : ReadableStream.from(res._chunks);
```
mhart: Everything in this repo is targeting Node 18 (including all esbuild configs and typescript configs), so I suspect this may require a major bump.

jasnell: Ugh, yep...

```
Welcome to Node.js v18.20.8.
Type ".help" for more information.
> ReadableStream.from
undefined
> 
```

I have to assume this is why the current code was used.

We can avoid the semver-major bump with a bit more work tho, it would just require an impl like..

```js
async function* gen() {
  for (const chunk of _chunks) {
    yield chunk;
  }
}
const generator = gen();
new ReadableStream({
  async pull(controller) {
    const next = await generator.next();
    if (next.done) {
      controller.close();
    } else {
      controller.enqueue(next.value);
    }
  }
});
```

Or something along those lines. Bit more work. Bumping the major would be preferred, however, especially since 18 is so far behind.

jasnell: Question tho... if this is targeting 18.x... is CI not targeting that?

packages/open-next/src/core/routing/util.ts
```diff
@@ -105,9 +105,14 @@ export function convertRes(res: OpenNextNodeResponse): InternalResult {
   const isBase64Encoded =
     isBinaryContentType(headers["content-type"]) ||
     !!headers["content-encoding"];
-  // We cannot convert the OpenNextNodeResponse to a ReadableStream directly
-  // You can look in the `aws-lambda.ts` file for some context
-  const body = Readable.toWeb(Readable.from(res.getBody()));
+  const body = new ReadableStream({
```
vicb, line 108: @conico974 would this work with aws-lambda ?

conico974: In theory it should, but I need to test in a real deployment

vicb:
> > This is still buffering all chunks and delivering them as one though right?
> 
> Yes, technically we shouldn't store them in memory and stream them directly without copying/storing.

Correct, see #992

packages/open-next/src/core/routing/util.ts
vicb: Thanks @conico974 for testing the code on aws 🙏

conico974 (approved):
LGTM, Thanks a lot for that guys.

vicb (approved):
Thanks all for this work on this 🎉

sommeeeer:
Awesome :+1:

PR #992, opened 2025-10-08
perf(OpenNextResponse): do not store the chunks for streamed responses

vicb:
/cc @anonrig @jasnell @mhart

### DIFF
--- a/.changeset/honest-lamps-smell.md
+++ b/.changeset/honest-lamps-smell.md
@@ -7,4 +7,5 @@ perf(OpenNextResponse): do not store the chunks for streamed responses
 There is no need to store the chunks for streamed responses.
 Not storing the chunks allows saving memory.
 
-Note that `OpenNextHandler` will now return an empty body for streamed responses.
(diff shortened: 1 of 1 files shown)


vicb:
We should probably keep track of `push`es for the AWS hack in `_flush`.

~I'll update this PR when https://github.com/opennextjs/opennextjs-aws/pull/991 is merged.~
~We could have a `private bodyLength = 0` updated when chunks are pushed and used that in `_flush`~

.changeset/honest-lamps-smell.md
```diff
@@ -0,0 +1,10 @@
+---
+"@opennextjs/aws": minor
+---
+
+perf(OpenNextResponse): do not store the chunks for streamed responses
+
+There is no need to store the chunks for streamed responses.
+Not storing the chunks allows saving memory.
+
+Note that `OpenNextHandler` will now return an empty body for streamed responses.
```
conico974:
```suggestion
BREAKING CHANGE: Note that `OpenNextHandler` will now return an empty body if your wrapper provides a `StreamCreator`
This could break custom converters.
```

PR #994, opened 2025-10-08
refactor(StreamCreator): allow opting-out of retaining the chunks

vicb:
The aws-lambda wrapper uses a fake `StreamCreator` and the chunks can not be discarded. They are now retained by default but can be discarded for implementation that don't need them (i.e. Cloudflare).


vicb:
Thanks Nico for the review and testing!

PR #995, opened 2025-10-08
update node.js to v20

anonrig:
Updates minimum required Node.js version to v20


rafaell-lycan:
> For upgrading to Node 20, we also need to upgrade the actual AWS CI deployment, which still use Node 18. And it probably means that we need to drop support for at least Next 12, and very likely 13 as well

Yeah, that makes sense. Given that Next.js v14 has been out for a while (2023) I think that dropping support for older Next versions is reasonable.

wdyt?

khuezy:
I think supporting `current` and the 2 versions prior seems reasonable.

nicholas-c:
Chiming in here, AWS have just released node 24 support. In terms of supporting next.js versions, Vercel themselves only support 15 and 16 (latest two versions) at the time of writing. 

https://nextjs.org/support-policy

In AWS terms, node 20 will be deprecated from April '26 - https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

PR #996, opened 2025-10-08
improve performance of streams

anonrig:
Overally reduces unnecessary allocations/copies.

### DIFF
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -105,15 +105,14 @@ export function convertRes(res: OpenNextNodeResponse): InternalResult {
   const isBase64Encoded =
     isBinaryContentType(headers["content-type"]) ||
     !!headers["content-encoding"];
-  let index = 0;
   const body = new ReadableStream({
     pull(controller) {
-      if (!res._chunks || index >= res._chunks.length) {
+      if (!res._chunks || res._chunks.length === 0) {
         controller.close();
         return;
       }
 
-      controller.enqueue(res._chunks[index++]);
+      controller.enqueue(res._chunks.shift());
     },
   });
   return {
--- a/packages/open-next/src/utils/binary.ts
+++ b/packages/open-next/src/utils/binary.ts
@@ -62,6 +62,6 @@ const commonBinaryMimeTypes = new Set([
 export function isBinaryContentType(contentType?: string | null) {
   if (!contentType) return false;
 
-  const value = contentType.split(";").at(0);
+  const value = contentType.split(";")[0];
   return commonBinaryMimeTypes.has(value);
 }
--- a/packages/open-next/src/utils/stream.ts
+++ b/packages/open-next/src/utils/stream.ts
@@ -4,65 +4,63 @@ export async function fromReadableStream(
   stream: ReadableStream<Uint8Array>,
   base64?: boolean,
 ): Promise<string> {
-  const reader = stream.getReader();
   const chunks: Uint8Array[] = [];
   let totalLength = 0;
 
-  try {
-    while (true) {
-      const { done, value } = await reader.read();
-      if (done) break;
-      chunks.push(value);
-      totalLength += value.length;
-    }
-
-    if (chunks.length === 0) {
-      return "";
-    }
-
-    if (chunks.length === 1) {
-      return Buffer.from(chunks[0]).toString(base64 ? "base64" : "utf8");
-    }
-
-    // Pre-allocate buffer with exact size to avoid reallocation
-    const buffer = Buffer.allocUnsafe(totalLength);
-    let offset = 0;
-    for (const chunk of chunks) {
-      buffer.set(chunk, offset);
-      offset += chunk.length;
-    }
-
-    return buffer.toString(base64 ? "base64" : "utf8");
-  } finally {
-    reader.releaseLock();
+  for await (const chunk of stream) {
+    chunks.push(chunk);
+    totalLength += chunk.length;
   }
+
+  if (chunks.length === 0) {
+    return "";
+  }
+
+  if (chunks.length === 1) {
+    return Buffer.from(chunks[0]).toString(base64 ? "base64" : "utf8");
+  }
+
+  // Pre-allocate buffer with exact size to avoid reallocation
+  const buffer = Buffer.alloc(totalLength);
+  let offset = 0;
+  for (const chunk of chunks) {
+    buffer.set(chunk, offset);
+    offset += chunk.length;
+  }
+
+  return buffer.toString(base64 ? "base64" : "utf8");
 }
 
 export function toReadableStream(
   value: string,
   isBase64?: boolean,
 ): ReadableStream {
-  const buffer = Buffer.from(value, isBase64 ? "base64" : "utf8");
-
-  return new ReadableStream({
-    start(controller) {
-      controller.enqueue(buffer);
-      controller.close();
+  return new ReadableStream(
+    {
+      pull(controller) {
+        // Defer the Buffer.from conversion to when the stream is actually read.
+        controller.enqueue(Buffer.from(value, isBase64 ? "base64" : "utf8"));
+        controller.close();
+      },
     },
-  });
+    { highWaterMark: 0 },
+  );
 }
 
 let maybeSomethingBuffer: Buffer | undefined;
 
 export function emptyReadableStream(): ReadableStream {
   if (process.env.OPEN_NEXT_FORCE_NON_EMPTY_RESPONSE === "true") {
-    return new ReadableStream({
-      start(controller) {
-        maybeSomethingBuffer ??= Buffer.from("SOMETHING");
-        controller.enqueue(maybeSomethingBuffer);
-        controller.close();
+    return new ReadableStream(
+      {
+        pull(controller) {
+          maybeSomethingBuffer ??= Buffer.from("SOMETHING");
+          controller.enqueue(maybeSomethingBuffer);
+          controller.close();
+        },
       },
-    });
+      { highWaterMark: 0 },
+    );
   }
   return new ReadableStream({
     start(controller) {


packages/open-next/src/core/routing/util.ts
```diff
@@ -105,14 +105,15 @@ export function convertRes(res: OpenNextNodeResponse): InternalResult {
...
+  let index = 0;
   const body = new ReadableStream({
     pull(controller) {
-      if (!res._chunks || res._chunks.length === 0) {
+      if (!res._chunks || index >= res._chunks.length) {
         controller.close();
         return;
       }
 
-      controller.enqueue(res._chunks.shift());
```
vicb, line 115: The intent here was to allow freeing memory early.

What's the problem with that? Having to move memory on `shift`s? it should be only refs so not too much?

packages/open-next/src/utils/binary.ts
```diff
@@ -60,8 +60,9 @@ const commonBinaryMimeTypes = new Set([
 ]);
 
 export function isBinaryContentType(contentType?: string | null) {
-  if (!contentType) return false;
+  if (contentType == null) return false;
```
vicb: not equivalent (empty string)

anonrig: It will be handled by line 66

packages/open-next/src/core/routing/util.ts
anonrig: This constantly triggers GC and when streaming this would create unnecessary GC pressure, which would impact the performance. Since we already allocated it, there is no benefit of an additional GC pressure

packages/open-next/src/utils/binary.ts
```diff
@@ -60,8 +60,9 @@ const commonBinaryMimeTypes = new Set([
...
 
-  const value = contentType?.split(";")[0] ?? "";
+  const value = contentType.split(";").at(0);
+  if (value == null) return false;
```
vicb: can value ever be `== null` ?

anonrig: if array length is 0, .at(0) will return undefined, and value == null will be true.

packages/open-next/src/utils/stream.ts
```diff
@@ -1,43 +1,72 @@
...
-        .read()
-        .then(({ done, value }) => {
-          if (done) {
-            resolve(Buffer.concat(chunks).toString(base64 ? "base64" : "utf8"));
-            return;
-          }
-          chunks.push(value);
-          pump();
-        })
-        .catch(reject);
```
vicb, line 23: Does the new version ever error?

anonrig: Yes, await reader.read() can throw an error. just like line 23.

packages/open-next/src/utils/binary.ts
vicb: Let's not change the coding standards on this repo for no reason
```suggestion
  if (!contentType) return false;
```

vicb: See https://github.com/opennextjs/opennextjs-aws/pull/996/files#r2414473979

packages/open-next/src/utils/binary.ts
vicb: I don't think `split()` ever returns an empty array

```suggestion
```

vicb: See https://github.com/opennextjs/opennextjs-aws/pull/996/files#r2414476395

packages/open-next/src/utils/stream.ts
vicb: Oops I saw a `catch` instead of a `finally` below

packages/open-next/src/utils/binary.ts
jasnell: split should only return an empty array if both the string and the split are empty, e.g. `''.split('')`.. so yeah, given that this is passing `;`, `value` should never `== null`

jasnell: The only time `b.split(c)` can return an empty array is if *both* `b` and `c` are empty strings.

packages/open-next/src/core/routing/util.ts
jasnell: does the shift *actually* add to the memory pressure? the removed object still exists so wouldn't be eligible for GC and the shift() itself does not trigger a GC. It does shift a bit of memory around within the array but not a GC as far as I know.

vicb: My reasoning is that it would be better to GC the chunk pas early as possible while waiting on slow IO rather than pushing the GC of all chunks to the end, but  I don't know what I'm talking about :)

erikcorry: I don't think this shift is very heavy on the GC unless there's something I'm missing.  In terms of reachability the shifted object goes out of the _chunks array.

erikcorry: If the array is bigger than 100 (kMaxCopyElements) then some crazy shit happens with the backing array moving forwards one word so that all the elements can change index without having to do a memmove, but we probably don't trigger that and it's not very hard on the GC in terms of performance (terrifying in terms of correctness, but that's a different story).

jasnell: Bottom line is that I do not believe that removing the `shift()` here is correct and @vicb is likely correct. We actually do want the object reference to be removed from the array unless these happen to be really big arrays.

The other possibility here is that we invert this somehow and use `pop()` instead, which is always O(1) as opposed to shift potentially being O(N) with the reindexing that's necessary... but then that also raises issues about adding things to the array. That's something we can explore... for this particular PR tho, I think keeping the `shift()` is appropriate unless profiling is somehow actually showing that the `shift()` is triggering GC somehow.

packages/open-next/src/utils/stream.ts
```diff
@@ -1,43 +1,72 @@
-import { Readable } from "node:stream";
-import type { ReadableStream } from "node:stream/web";
+import { ReadableStream } from "node:stream/web";
 
-export function fromReadableStream(
+export async function fromReadableStream(
```
jasnell, line 3: there's a potential behavioral difference here. `stream.getReader()` can throw. In the original method, it would throw synchronously, with this change it's converted into a promise rejection. It'll need to be determined if that change in behavior makes this a semver-major.

packages/open-next/src/utils/stream.ts
```diff
@@ -1,43 +1,72 @@
...
-          if (done) {
-            resolve(Buffer.concat(chunks).toString(base64 ? "base64" : "utf8"));
-            return;
-          }
-          chunks.push(value);
-          pump();
-        })
-        .catch(reject);
+  try {
+    while (true) {
```
jasnell: using `for await (const chunk of stream)` might be better than the `while (true)` with the `await reader.read()`

packages/open-next/src/utils/stream.ts
```diff
@@ -1,43 +1,72 @@
...
+      const { done, value } = await reader.read();
+      if (done) break;
+      chunks.push(value);
+      totalLength += value.length;
```
jasnell: does there need to be validation that `value` actually is a typed array?

packages/open-next/src/utils/stream.ts
```diff
@@ -1,43 +1,72 @@
...
+    if (chunks.length === 0) {
+      return "";
+    }
+
+    if (chunks.length === 1) {
+      return Buffer.from(chunks[0]).toString(base64 ? "base64" : "utf8");
+    }
+
+    // Pre-allocate buffer with exact size to avoid reallocation
+    const buffer = Buffer.allocUnsafe(totalLength);
```
jasnell:
```suggestion
    const buffer = Buffer.alloc(totalLength);
```

Just safer... and in workers `Buffer.allocUnsafe(n)` is the same thing anyway.

packages/open-next/src/utils/stream.ts
```diff
@@ -1,43 +1,72 @@
...
+    let offset = 0;
+    for (const chunk of chunks) {
+      buffer.set(chunk, offset);
+      offset += chunk.length;
+    }
+
+    return buffer.toString(base64 ? "base64" : "utf8");
```
jasnell: For the case where `base64` is `false`... would using a `TextDecoder` here in streaming mode actually be better than the buffer concat followed by `utf8` conversion? I honestly don't know as I haven't compared the performance of both...

anonrig: I had the same question but I'm not sure.

packages/open-next/src/utils/stream.ts
anonrig: Removed stream.getReader()

packages/open-next/src/utils/stream.ts
```diff
@@ -1,43 +1,72 @@
...
-  return Readable.toWeb(
-    Readable.from(Buffer.from(value, isBase64 ? "base64" : "utf8")),
-  );
+  const buffer = Buffer.from(value, isBase64 ? "base64" : "utf8");
+
+  return new ReadableStream({
+    start(controller) {
+      controller.enqueue(buffer);
+      controller.close();
+    },
```
jasnell, line 45: I wonder if it would be slightly better here to have this be

```js
return new ReadableStream({
  pull(controller) {
    controller.enqueue(Buffer.from(value, isBase64 ? 'base64' : 'utf8'))
    controller.close();
  }
}, { highWaterMark: 0 });
```

The reasoning is simple:

1. Defer the Buffer.from conversion to when the stream is actually read.
2. By doing the `const buffer = Buffer.from(...)` outside of the pull, you're guaranteeing that start will capture both the original `value` and the copied `buffer`, doubling the amount of memory that it is retaining. By deferring the Buffer.from to the lexical scope of the pull function execution, you'll defer the copy and reduce the amount of memory the closure is retaining. Keep in mind that the `start` function is held until the readable stream close is processed, so the `value` will be retained for that long, but the `buffer` won't be retained by the closure at all (just the queue)

anonrig: Agreed.

packages/open-next/src/utils/stream.ts
```diff
@@ -1,43 +1,67 @@
...
 export function emptyReadableStream(): ReadableStream {
   if (process.env.OPEN_NEXT_FORCE_NON_EMPTY_RESPONSE === "true") {
-    return Readable.toWeb(Readable.from([Buffer.from("SOMETHING")]));
+    return new ReadableStream({
+      start(controller) {
+        maybeSomethingBuffer ??= Buffer.from("SOMETHING");
+        controller.enqueue(maybeSomethingBuffer);
+        controller.close();
+      },
+    });
```
jasnell: Let's make these lazier...

```suggestion
    return new ReadableStream({
      pull(controller) {
        maybeSomethingBuffer ??= Buffer.from("SOMETHING");
        controller.enqueue(maybeSomethingBuffer);
        controller.close();
      },
    }, { highWaterMark: 0 });
```

That way the buffer is never actually allocated until and unless the stream is actually read.

packages/open-next/src/utils/stream.ts
jasnell: Well, yes and no. `for await (const chunk of stream)` still calls `getReader()` internally, so the issue with the potential change in error handling is still present.

packages/open-next/src/utils/stream.ts
```diff
@@ -1,43 +1,70 @@
...
-            return;
-          }
-          chunks.push(value);
-          pump();
-        })
-        .catch(reject);
-    }
-    pump();
-  });
+  for await (const chunk of stream) {
```
jasnell, line 10: there's still a difference in error handling here that might make this semver-major. `for await (const chunk of stream)` still calls `getReader()` under the covers.

conico974: Given where this one is used right now, I don't think it's worth a major really (or even a minor).
It's only used in some converters or in our own implementation of the composable cache.
For end users none of these places are directly accessed, so I think we're fine really.

jasnell: sounds good!

anonrig:
@conico974 would you mind landing this?

PR #1000, opened 2025-10-10
chore: github actions

vicb:
- Revert Node version change to see if it helps with the release workflow
- Drop installing the deps from `.github/workflows/v2-release.yml` as it is now part of the setup.


vicb:
@sommeeeer

sommeeeer:
Thanks a lot Victor!

PR #1002, opened 2025-10-10
improve composable-cache, converters and header parsing performance

anonrig:
We saw that under high concurrency, Nth request was taking a lot more time than others. This was caused by the composable cache. Prior to this pull-request, cache was consuming all data stream (the response of the page), just to access it and later convert it to a stream (and discard the data cached). This created a lot of GC pressure and caused a lot of unnecessary memory usage.

### DIFF
--- a/packages/open-next/src/overrides/originResolver/pattern-env.ts
+++ b/packages/open-next/src/overrides/originResolver/pattern-env.ts
@@ -3,34 +3,71 @@ import type { OriginResolver } from "types/overrides";
 
 import { debug, error } from "../../adapters/logger";
 
+// Cache parsed origin and compiled patterns at module level
+let cachedOrigin: Record<string, Origin> | null = null;
+const cachedPatterns: Array<{
+  key: string;
+  patterns: string[];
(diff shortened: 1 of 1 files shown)


packages/open-next/src/utils/stream.ts
```diff
@@ -20,13 +20,9 @@ export async function fromReadableStream(
...
-  // Pre-allocate buffer with exact size to avoid reallocation
-  const buffer = Buffer.alloc(totalLength);
-  let offset = 0;
-  for (const chunk of chunks) {
-    buffer.set(chunk, offset);
-    offset += chunk.length;
-  }
+  // Use Buffer.concat which is more efficient than manual allocation and copy
+  // It handles the allocation and copy in optimized native code
+  const buffer = Buffer.concat(chunks, totalLength);
```
anonrig, line 25: Once I look at this with a different perspective, I think this might be faster, but I didn't run a separate benchmark for this. @jasnell @vicb

anonrig:
Closing in favor of separate pull-requests.

PR #1004, opened 2025-10-10
improve composable-cache performance

anonrig:
We saw that under high concurrency, Nth request was taking a lot more time than others. This was caused by the composable cache. Prior to this pull-request, cache was consuming all data stream (the response of the page), just to access it and later convert it to a stream (and discard the data cached). This created a lot of GC pressure and caused a lot of unnecessary memory usage.

### DIFF
--- a/packages/open-next/src/adapters/composable-cache.ts
+++ b/packages/open-next/src/adapters/composable-cache.ts
@@ -1,9 +1,15 @@
-import type { ComposableCacheEntry, ComposableCacheHandler } from "types/cache";
+import type {
+  ComposableCacheEntry,
+  ComposableCacheHandler,
+  StoredComposableCacheEntry,
+} from "types/cache";
 import { writeTags } from "utils/cache";
-import { fromReadableStream, toReadableStream } from "utils/stream";
 import { debug } from "./logger";
 
-const pendingWritePromiseMap = new Map<string, Promise<ComposableCacheEntry>>();
+const pendingWritePromiseMap = new Map<
+  string,
+  Promise<StoredComposableCacheEntry>
+>();
 
 export default {
   async get(cacheKey: string) {
@@ -11,7 +17,12 @@ export default {
       // We first check if we have a pending write for this cache key
       // If we do, we return the pending promise instead of fetching the cache
       const stored = pendingWritePromiseMap.get(cacheKey);
-      if (stored) return stored;
+      if (stored) {
+        return stored.then((val) => ({
+          ...val,
+          value: val.value.stream(),
+        }));
+      }
 
       const result = await globalThis.incrementalCache.get(
         cacheKey,
@@ -50,7 +61,7 @@ export default {
 
       return {
         ...result.value,
-        value: toReadableStream(result.value.value),
+        value: result.value.value.stream(),
       };
     } catch (e) {
       debug("Cannot read composable cache entry");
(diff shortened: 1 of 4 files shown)


.changeset/curly-beans-exercise.md
conico974, line 1: We don't need this one

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -69,73 +54,77 @@ export default {
...
   async set(cacheKey: string, pendingEntry: Promise<ComposableCacheEntry>) {
-    const promiseEntry = pendingEntry.then(async (entry) => ({
-      ...entry,
-      value: await fromReadableStream(entry.value),
-    }));
-    pendingWritePromiseMap.set(cacheKey, promiseEntry);
+    const teedPromise = pendingEntry.then((entry) => {
+      // Optimization: We avoid consuming and stringifying the stream here,
+      // because it creates double copies just to be discarded when this function
+      // ends. This avoids unnecessary memory usage, and reduces GC pressure.
```
conico974: Wouldn't this cause an issue if we read this more than once in the get ? We then have a single ReadableStream that multiple reader would try to access.
We could teed on get as well, but it means we'd end up with a copy in memory and multiple ReadableStream, not sure if it's worth it or not ?

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -69,73 +54,77 @@ export default {
...
+        await writeTags(tagsToWrite);
       }
     }
   },
 
-  async refreshTags() {
-    // We don't do anything for now, do we want to do something here ???
-    return;
-  },
+  async refreshTags() {},
```
conico974: Add the comment back

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -69,73 +54,77 @@ export default {
...
   async getExpiration(...tags: string[]) {
-    if (globalThis.tagCache.mode === "nextMode") {
-      return globalThis.tagCache.getLastRevalidated(tags);
-    }
-    // We always return 0 here, original tag cache are handled directly in the get part
-    // TODO: We need to test this more, i'm not entirely sure that this is working as expected
-    return 0;
+    return globalThis.tagCache.mode === "nextMode"
+      ? globalThis.tagCache.getLastRevalidated(tags)
+      : 0;
```
conico974: Could you add back the comment

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -69,73 +54,77 @@ export default {
...
   async expireTags(...tags: string[]) {
     if (globalThis.tagCache.mode === "nextMode") {
       return writeTags(tags);
     }
+
     const tagCache = globalThis.tagCache;
     const revalidatedAt = Date.now();
-    // For the original mode, we have more work to do here.
-    // We need to find all paths linked to to these tags
     const pathsToUpdate = await Promise.all(
```
conico974, line 133: Add the comment back

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -69,73 +54,77 @@ export default {
...
-    await writeTags(Array.from(setToWrite));
+    await writeTags(Array.from(dedupeMap.values()));
   },
 
-  // This one is necessary for older versions of next
-  async receiveExpiredTags(...tags: string[]) {
-    // This function does absolutely nothing
-    return;
-  },
+  async receiveExpiredTags() {},
```
conico974, line 152: Same add the comment back, just the external one internal is useless

packages/open-next/src/adapters/composable-cache.ts
anonrig: well, the goal is to avoid having unnecessary memory allocations. Both optimizations have quirks, but right now, concurrent calls get stalled and have more impact due to this unnecessary copy. i'm open to suggestions.

jasnell: Might take a bit of experimentation to get exactly right. I'm also not convinced that this new bit is correct either. Another potential approach you could take would be to use a `Blob` to buffer the data and use it's `.stream()` method to get individual `ReadableStream` instances that read from the single in memory buffer. That way you're not relying on, or victim to, `ReadableStream` tees terrible default backpressure management that ends up creating multiple buffers in memory. But it's going to take a bit of careful thought to get right.

Regardless, I think this is going to require some careful consideration to get correct. @anonrig is absolutely correct that the current code is needlessly bottlenecking with too many additional copies.

packages/open-next/src/utils/stream.ts
```diff
@@ -20,13 +20,9 @@ export async function fromReadableStream(
...
-  // Pre-allocate buffer with exact size to avoid reallocation
-  const buffer = Buffer.alloc(totalLength);
-  let offset = 0;
-  for (const chunk of chunks) {
-    buffer.set(chunk, offset);
-    offset += chunk.length;
-  }
+  // Use Buffer.concat which is more efficient than manual allocation and copy
+  // It handles the allocation and copy in optimized native code
+  const buffer = Buffer.concat(chunks, totalLength);
```
jasnell: This should likely be bench marked more. I've seen cases with smaller buffers where `buffer.set` was faster. There's likely a threshold at which the additional overhead of the JS-to-C++ boundary call used by `Buffer.concat` outweighs the `buffer.set(...)` call.

anonrig: You're right. I'll revert this for now.

packages/open-next/src/adapters/composable-cache.ts
anonrig: I've taken another shot and stored a blob on the composable cache rather than text or stream. So that, any caller can create a readable stream from the blob with almost no cost.

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -69,28 +70,41 @@ export default {
...
       pendingWritePromiseMap.delete(cacheKey);
     });
+
     await globalThis.incrementalCache.set(
       cacheKey,
-      {
-        ...entry,
-        value: entry.value,
-      },
+      entryWithBlob,
```
conico974, line 92: Pretty sure that the incrementalCache here expect a string for `entry.value` for the set here

anonrig: I actually changed the type, so we store a blob now.

conico974: But that's not enough, if we'd want to do that we need to change every implementation [here](https://github.com/opennextjs/opennextjs-aws/tree/main/packages/open-next/src/overrides/incrementalCache) and the ones in cloudflare.
And I'm not sure what we gain from doing that really, someone will serialize that in the `incrementalCache.set` anyway.
We can use a blob in the map, but not for the incremental cache

anonrig: ok, so that would be problematic. because if we store it as string, we would need to convert it into stream everytime we need to interact with it, since next.js waits for ReadableStream. We want to avoid having this unnecessary allocation.

conico974: Yeah but that's a big breaking change (definitely major).
And that's actually already kind of planned for next major, but i think it requires some thought about how we want to store it really. 
The way I see it is an entire refactor of the incremental cache and splitting the data into actual data (that would be blob/stream) and metadata that would be serializable small data

PR #1005, opened 2025-10-10
perf(cloudflare): optimize readable/writable piping

anonrig:
Prior to this pull-request we were copying every single chunk. This change make sure that we avoid unnecessary copies in any way that we can.

Part of https://github.com/opennextjs/opennextjs-aws/pull/1002

### DIFF
--- a/.changeset/sour-socks-rescue.md
+++ b/.changeset/sour-socks-rescue.md
@@ -2,4 +2,4 @@
 "@opennextjs/aws": patch
 ---
 
-Improve Cloudflare readable/writable streaming performance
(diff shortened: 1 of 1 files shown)


.changeset/sour-socks-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Improve Cloudflare readable/writable streaming performance
```
vicb:
```suggestion
perf(cloudflare): optimize readable/writable piping
```

PR #1006, opened 2025-10-10
perf(converters): Improve performance

anonrig:
Improves the performance of converters by skipping unnecessary work.

### DIFF
--- a/packages/open-next/src/overrides/converters/node.ts
+++ b/packages/open-next/src/overrides/converters/node.ts
@@ -8,28 +8,12 @@ import { extractHostFromHeaders, getQueryFromSearchParams } from "./utils.js";
 const converter: Converter = {
   convertFrom: async (req: IncomingMessage & { protocol?: string }) => {
     const body = await new Promise<Buffer>((resolve) => {
-      const contentLength = req.headers["content-length"];
-      const expectedLength = contentLength
-        ? Number.parseInt(contentLength, 10)
-        : undefined;
       const chunks: Uint8Array[] = [];
-      let receivedLength = 0;
-
       req.on("data", (chunk) => {
         chunks.push(chunk);
-        receivedLength += chunk.length;
       });
       req.on("end", () => {
-        // Use pre-allocated buffer if we have content-length and it matches
-        if (
-          expectedLength &&
-          receivedLength === expectedLength &&
-          chunks.length === 1
-        ) {
-          resolve(Buffer.from(chunks[0]));
-        } else {
-          resolve(Buffer.concat(chunks));
-        }
+        resolve(Buffer.concat(chunks));
       });
     });
 


packages/open-next/src/core/routing/util.ts
```diff
@@ -150,15 +150,9 @@ export function convertToQueryString(query: Record<string, string | string[]>) {
...
   const query = new URLSearchParams(querystring);
-  const queryObject: Record<string, string[] | string> = {};
-
-  for (const key of query.keys()) {
-    const queries = query.getAll(key);
-    queryObject[key] = queries.length > 1 ? queries : queries[0];
-  }
-
-  return queryObject;
+  return getQueryFromIterator(query.entries());
```
vicb: @anonrig ok to revert this? This is not immediately obvious to me that it does the same thing

anonrig: yes it's ok

.changeset/dirty-rabbits-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Improve performance of converters used by all adapters
```
vicb:
```suggestion
perf(converters): Improve performance
```

packages/open-next/src/core/routing/util.ts
vicb:
```suggestion
 const queryObject: Record<string, string[] | string> = {};

  for (const key of query.keys()) {
    const queries = query.getAll(key);
    queryObject[key] = queries.length > 1 ? queries : queries[0];
  }

  return queryObject;
```

packages/open-next/src/core/routing/util.ts
```diff
@@ -150,8 +150,9 @@ export function convertToQueryString(query: Record<string, string | string[]>) {
  * @__PURE__
  */
 export function convertToQuery(querystring: string) {
+  if (!querystring) return {};
   const query = new URLSearchParams(querystring);
-  const queryObject: Record<string, string[] | string> = {};
+ const queryObject: Record<string, string[] | string> = {};
```
vicb:
```suggestion
  const queryObject: Record<string, string[] | string> = {};
```

packages/open-next/src/overrides/converters/node.ts
```diff
@@ -8,12 +8,28 @@ import { extractHostFromHeaders, getQueryFromSearchParams } from "./utils.js";
 const converter: Converter = {
   convertFrom: async (req: IncomingMessage & { protocol?: string }) => {
     const body = await new Promise<Buffer>((resolve) => {
+      const contentLength = req.headers["content-length"];
+      const expectedLength = contentLength
+        ? Number.parseInt(contentLength, 10)
```
conico974: If I send a fake `content-length` header here, I can make it crash. We should use a safe version of the parse

vicb: Let's revert this code for now and handle later.

PR #1007, opened 2025-10-10
replace delete with assigning to undefined

anonrig:
Replacing `delete X.y` with `X.y = undefined` avoids killing any optimizations done by V8 to access the properties of an object faster.

### DIFF
--- a/packages/open-next/src/core/requestHandler.ts
+++ b/packages/open-next/src/core/requestHandler.ts
@@ -108,7 +108,8 @@ export async function openNextHandler(
           overwrittenResponseHeaders[key] = value;
         }
         headers[key] = value;
-        delete headers[rawKey];
+        // @ts-expect-error Setting it to undefined is faster than deleting the attribute.
+        headers[rawKey] = undefined;
       }
 
       if (
@@ -238,7 +239,7 @@ async function processRequest(
   // @ts-ignore
   // Next.js doesn't parse body if the property exists
   // https://github.com/dougmoscrop/serverless-http/issues/227
-  delete req.body;
+  req.body = undefined;
 
   // Here we try to apply as much request metadata as possible
   // We apply every metadata from `resolve-routes` https://github.com/vercel/next.js/blob/916f105b97211de50f8580f0b39c9e7c60de4886/packages/next/src/server/lib/router-utils/resolve-routes.ts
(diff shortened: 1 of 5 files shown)


packages/open-next/src/core/requestHandler.ts
```diff
@@ -238,7 +239,7 @@ async function processRequest(
   // @ts-ignore
   // Next.js doesn't parse body if the property exists
   // https://github.com/dougmoscrop/serverless-http/issues/227
-  delete req.body;
+  req.body = undefined;
```
vicb, line 242: The comment above explicitely says "the property exists"

packages/open-next/src/http/util.ts
```diff
@@ -14,6 +14,12 @@ export const parseHeaders = (
       continue;
     }
     const keyLower = key.toLowerCase();
+    if (keyLower === "set-cookie") {
```
vicb, line 17: is that a change in behavior?

packages/open-next/src/core/requestHandler.ts
jasnell: Might be worth an upstream pr to next.js tho I suspect it's been tried before.

The key challenge with using `delete` here is that it immediately puts `req` into a slower / non-optimized mode for all property accesses. Whether or not that moves the needle or not in performance depends on usage pattern. In this particular case, let's leave this as `delete req.body` and see if we can try to fix upstream.

but as a general rule, we should be avoiding `delete` like the plague.

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -188,10 +188,6 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
 
     const parsedHeaders = parseHeaders(this.headers);
 
-    // We need to remove the set-cookie header from the parsed headers because
-    // it does not handle multiple set-cookie headers properly
-    delete parsedHeaders[SET_COOKIE_HEADER];
-
```
jasnell, line 194: I see this is removed but not replaced?

jasnell: oh, I see, you're removing it in the util.ts change instead.

packages/open-next/src/http/util.ts
```diff
@@ -14,6 +14,12 @@ export const parseHeaders = (
...
+      // We need to remove the set-cookie header from the parsed headers because
+      // it does not handle multiple set-cookie headers properly
+      continue;
```
jasnell, line 20: side note... I keep thinking that it would be nice if `Object.entries(...)` (and the iterable pattern in general) would accept a filter function...

```js
Object.entries(headers, filter);
```

...that could optimize this pattern more and make it so the collection only had to be iterated through once... That's for another time tho.

PR #1008, opened 2025-10-10
avoid creating unnecessary headers instance

anonrig:
Let's make sure that we don't copy all of the headers, just to mutate it.

Part of https://github.com/opennextjs/opennextjs-aws/pull/1002

### DIFF
--- a/packages/open-next/src/overrides/wrappers/cloudflare-node.ts
+++ b/packages/open-next/src/overrides/wrappers/cloudflare-node.ts
@@ -41,15 +41,16 @@ const handler: WrapperHandler<InternalEvent, InternalResult> =
       }): Writable {
         const { statusCode, cookies, headers } = prelude;
 
-        const responseHeaders = new Headers(headers);
-        for (const cookie of cookies) {
-          responseHeaders.append("Set-Cookie", cookie);
-        }
-
         // TODO(vicb): this is a workaround to make PPR work with `wrangler dev`
         // See https://github.com/cloudflare/workers-sdk/issues/8004
         if (url.hostname === "localhost") {
-          responseHeaders.set("Content-Encoding", "identity");
+          headers["content-encoding"] = "identity";
+        }
+
+        // Build headers array - set-cookie must be added separately for each cookie
+        const headerEntries: [string, string][] = Object.entries(headers);
+        for (const cookie of cookies) {
+          headerEntries.push(["set-cookie", cookie]);
         }
 
         const { readable, writable } = new TransformStream({
@@ -60,7 +61,7 @@ const handler: WrapperHandler<InternalEvent, InternalResult> =
         const body = NULL_BODY_STATUSES.has(statusCode) ? null : readable;
         const response = new Response(body, {
           status: statusCode,
-          headers: responseHeaders,
+          headers: headerEntries,
         });
         resolveResponse(response);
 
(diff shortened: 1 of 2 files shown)


packages/open-next/src/overrides/wrappers/cloudflare-node.ts
```diff
@@ -41,15 +41,12 @@ const handler: WrapperHandler<InternalEvent, InternalResult> =
       }): Writable {
         const { statusCode, cookies, headers } = prelude;
 
-        const responseHeaders = new Headers(headers);
-        for (const cookie of cookies) {
-          responseHeaders.append("Set-Cookie", cookie);
-        }
+        headers["set-cookie"] = cookies.join(",");
```
vicb: I think that's wrong.
I you have `a=1,b=2` it will create a single `a` cookie with the value `1,b=2`

anonrig:
```
> let a = new Headers()
undefined
> a.append('set-cookie', 1)
undefined
> a.append('set-cookie', 2)
undefined
> a
Headers { 'set-cookie': '1, 2' }
```

This is true. Other than the missing extra space

vicb: image

vicb: ^ with a space

anonrig:
```suggestion
        headers["set-cookie"] = cookies.join(", ");
```

vicb: @anonrig Looks like the actual syntax would be: `a="1";b="2"` (quotes required)

anonrig: I see. Let me fix it.

packages/open-next/src/overrides/wrappers/cloudflare-node.ts
```diff
@@ -41,15 +41,16 @@ const handler: WrapperHandler<InternalEvent, InternalResult> =
...
         // See https://github.com/cloudflare/workers-sdk/issues/8004
         if (url.hostname === "localhost") {
-          responseHeaders.set("Content-Encoding", "identity");
+          headers["content-encoding"] = "identity";
+        }
+
+        // Build headers array - set-cookie must be added separately for each cookie
+        const headerEntries: [string, string][] = Object.entries(headers);
+        for (const cookie of cookies) {
+          headerEntries.push(["set-cookie", cookie]);
```
jasnell, line 53: Hmm.. while this might be more efficient I'm not convinced this is right. This is going to end up skipping any kind of validation on the value of `cookie`, which could end up causing problems elsewhere.

packages/open-next/src/overrides/wrappers/cloudflare-node.ts
```diff
@@ -41,15 +41,16 @@ const handler: WrapperHandler<InternalEvent, InternalResult> =
...
-        const responseHeaders = new Headers(headers);
-        for (const cookie of cookies) {
-          responseHeaders.append("Set-Cookie", cookie);
-        }
-
         // TODO(vicb): this is a workaround to make PPR work with `wrangler dev`
         // See https://github.com/cloudflare/workers-sdk/issues/8004
         if (url.hostname === "localhost") {
-          responseHeaders.set("Content-Encoding", "identity");
+          headers["content-encoding"] = "identity";
```
stof, line 47: this now mutates the headers object coming from outside the function, which changes its behavior (it now has a side effect on the caller values).

PR #1010, opened 2025-10-10
avoid unnecessary buffer copies on opennext.js response

anonrig:
Several optimizations but I think the most impactful one is to avoid having an unnecessary copy everytime getBody() is called.

### DIFF
--- a/packages/open-next/src/http/openNextResponse.ts
+++ b/packages/open-next/src/http/openNextResponse.ts
@@ -14,6 +14,14 @@ import { parseHeaders, parseSetCookieHeader } from "./util";
 
 const SET_COOKIE_HEADER = "set-cookie";
 const CANNOT_BE_USED = "This cannot be used in OpenNext";
+const ERROR_CACHE_CONTROL_HEADER =
+  "private, no-cache, no-store, max-age=0, must-revalidate";
+
+// Cache environment variable checks at module load time
+const FORCE_NON_EMPTY_RESPONSE =
+  process.env.OPEN_NEXT_FORCE_NON_EMPTY_RESPONSE === "true";
+const DANGEROUSLY_SET_ERROR_HEADERS =
+  process.env.OPEN_NEXT_DANGEROUSLY_SET_ERROR_HEADERS === "true";
 
 // We only need to implement the methods that are used by next.js
 export class OpenNextNodeResponse extends Transform implements ServerResponse {
@@ -26,6 +34,7 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
   private _cookies: string[] = [];
   private responseStream?: Writable;
   private bodyLength = 0;
+  private _cachedBody?: Buffer;
 
   // To comply with the ServerResponse interface :
   strictContentLength = false;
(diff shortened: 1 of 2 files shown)


packages/open-next/src/http/openNextResponse.ts
```diff
@@ -404,16 +426,15 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
       return;
     }
     // We only check for 404 and 500 errors
     // The rest should be errors that are handled by the user and they should set the cache headers themselves
     if (this.statusCode === 404 || this.statusCode === 500) {
       // For some reason calling this.setHeader("Cache-Control", "no-cache, no-store, must-revalidate") does not work here
       // The function is not even called, i'm probably missing something obvious
-      this.headers["cache-control"] =
-        "private, no-cache, no-store, max-age=0, must-revalidate";
+      this.headers["cache-control"] = ERROR_CACHE_CONTROL_HEADER;
```
vicb, line 437: Does that make any difference?
I understand where you come from but wouldn't Node do the same thing here?

anonrig: every little bit helps

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -281,7 +306,8 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
   }
 
   getBody() {
-    return Buffer.concat(this._chunks);
+    this._cachedBody ??= Buffer.concat(this._chunks);
```
vicb, line 309: Do we have evidence that `getBody` is called multiple times?

anonrig: I don't have it, but it doesn't make this optimization less important since this method enables developers to copy the whole body at any time.

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -290,6 +316,8 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
     if (this.streamCreator?.retainChunks !== false) {
       // Avoid keeping chunks around when the `StreamCreator` supports it to save memory
       this._chunks.push(buffer);
+      // Invalidate cached body when new chunks are added
+      this._cachedBody = undefined;
```
jasnell, line 320: This can be a bit problematic for memory. This means that we're persistently doubling the amount of memory consumed. Sure it means we eliminate multiple potential copies but at the cost of definitely holding more memory.

Instead, if we're going to cache this, we should replace `this._chunks` with the combined buffer so that the separate chunks can be reclaimed.

I think a better approach is to see if we can eliminate this entirely, and instead rely on regular stream buffering.

Ultimately I'm not sure if this is the right change.

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -103,7 +112,7 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
 
   get finished() {
     return this.responseStream
-      ? this.responseStream?.writableFinished
+      ? this.responseStream.writableFinished
```
jasnell, line 115: @anonrig ... just a general note. This looks like a drive-by change unrelated to the actual unnecessary buffer copies. For additional changes like this, can you put these into separate commits moving forward? Makes it a bit difficult to separate what is or is not relevant to the actual allocation/copy improvements.

PR #1011, opened 2025-10-10
fix(cloudflare): ignore errors when the stream is already closed

vicb:
Fix:

image

https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/18413835283/job/52472646373#step:7:31


vicb:
Thanks for the reviews, appreciated!

PR #1013, opened 2025-10-11
perf: avoid unnecessary buffer copy in internalWrite

jasnell:
See the code comments for details.

One challenge/question: it's not clear that this function is actually being tested with `pnpm run test`... to check I added a `throw new Error('boom')` and ran `pnpm run test` and everything still passed.

/cc @vicb @anonrig

### DIFF
--- a/.changeset/olive-cars-count.md
+++ b/.changeset/olive-cars-count.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+perf: avoid unnecessary buffer copy in internalWrite


vicb:
> See the code comments for details.
> 
> One challenge/question: it's not clear that this function is actually being tested with `pnpm run test`... to check I added a `throw new Error('boom')` and ran `pnpm run test` and everything still passed.
> 
> /cc @vicb @anonrig

See the [README](https://github.com/opennextjs/opennextjs-aws/blob/main/README.md#testing) to run the e2e tests

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -285,13 +285,20 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
   }
 
   private _internalWrite(chunk: any, encoding: BufferEncoding) {
-    const buffer = Buffer.from(chunk, encoding);
+    // When encoding === 'buffer', chunk is already a Buffer
+    // and does not need to be converted again.
+    // @ts-expect-error TS2367 'encoding' can be 'buffer', but it's not in the
+    // official type definition
+    const buffer = encoding === "buffer" ? chunk : Buffer.from(chunk, encoding);
```
vicb, line 292: Image

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -285,13 +285,20 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
     this.bodyLength += buffer.length;
     if (this.streamCreator?.retainChunks !== false) {
       // Avoid keeping chunks around when the `StreamCreator` supports it to save memory
       this._chunks.push(buffer);
     }
-    this.push(chunk, encoding);
+    // We already have the data as a buffer, let's push it as is to avoid
+    // unnecessary additional conversion down the stream pipeline.
+    // @ts-expect-error TS2345 'buffer' is not in the official type definition
+    this.push(buffer, "buffer");
```
vicb: I have tried `push(buffer, "whatever")` and this works so the encoding only seem to matter for string values.

[The docs](https://nodejs.org/api/stream.html#readablepushchunk-encoding) say "Encoding of string chunks. Must be a valid Buffer encoding, such as 'utf8' or 'ascii'." - and indeed invalid encodings fail for strings.

But using a non existent encoding for buffer looks like an UB we should not rely on.

packages/open-next/src/http/openNextResponse.ts
vicb, line 292: Maybe the only code change in this PR should be:

```ts
const buffer = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk, encoding);
```

vicb: which of course would fail here if someone modifies the buffer. i.e. we can image a pattern when some code populates a buffer and pushes it before re-using it for the next compute.

jasnell: That's fundamentally different.

```js
const t = new stream.Transform({
  transform(chunk, encoding, callback) {
    console.log(chunk, encoding);
    callback();
  }
});

t.write('hello');
// Prints: <Buffer 68 65 6c 6c 6f> buffer

t.write(Buffer.from('hello'))
// Prints: <Buffer 68 65 6c 6c 6f> buffer
```

The `buffer` value is how Node.js streams communicate that the `chunk` is not an encoded string and does not need to be decoded, and in this case it means you do not need to be copying the `chunk` by using `Buffer.from(chunk, encoding)`

packages/open-next/src/http/openNextResponse.ts
jasnell: It's not undefined behavior at all. It's directly handled in the code. It's not a "non existent encoding".

* https://github.com/nodejs/node/blob/main/lib/internal/streams/writable.js#L465
* https://nodejs.org/docs/latest/api/stream.html#transform_transformchunk-encoding-callback
* https://nodejs.org/docs/latest/api/stream.html#writable_writevchunks-callback

packages/open-next/src/http/openNextResponse.ts
jasnell: This is already a reality in Node.js.

```js
const t = new stream.Transform({
  transform(chunk) {
    // enqueues the chunk.
    // <Buffer 68 65 6c 6c 6f>
    this.push(chunk);
  }
});

const b = Buffer.from('hello');
// <Buffer 68 65 6c 6c 6f>

t.write(b);

b.fill(0);
// <Buffer 00 00 00 00 00>

// consume the buffered chunk
t.on('data', console.log);
// <Buffer 00 00 00 00 00>
```

packages/open-next/src/http/openNextResponse.ts
jasnell: Passing `'buffer'` to `push` is allowed even if not strictly necessary. I included it here so that the intent is absolutely clear. It has no cost, is not an error, and just works. There will never be a case when `push(buffer, 'buffer')` means anything other than `buffer` being a `Buffer`.

packages/open-next/src/http/openNextResponse.ts
jasnell: the other bit, `Buffer.isBuffer(chunk)` is just an alias for `chunk instanceof Buffer` which is generally slower in general `encoding === 'buffer'`.

```
(higher number is better)
misc/c.js p="instanceof" n=4096: 16,418,083.94226414
misc/c.js p="encoding" n=4096: 36,453,280.88427685
```

vicb:
Thanks for the additional details @jasnell.

IIUC the first change makes sense to avoid having to copy the buffer when it is already a buffer.

Still not clear about `this.push(buffer, "buffer");`:
- the doc explicitely says it is not allowed
- it does not seem to make any difference [in the code](https://github.com/nodejs/node/blob/main/lib/internal/streams/readable.js#L385)

Can you measure any perf difference for that second change?

packages/open-next/src/http/openNextResponse.ts
vicb: @jasnell I'll change this for just `this.push(buffer);` as the encoding is not used for buffers anyway and merge the PR

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -285,13 +285,20 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
+    // @ts-expect-error TS2367 'encoding' can be 'buffer', but it's not in the
+    // official type definition
+    const buffer = encoding === "buffer" ? chunk : Buffer.from(chunk, encoding);
     this.bodyLength += buffer.length;
     if (this.streamCreator?.retainChunks !== false) {
       // Avoid keeping chunks around when the `StreamCreator` supports it to save memory
       this._chunks.push(buffer);
     }
-    this.push(chunk, encoding);
+    // We already have the data as a buffer, let's push it as is to avoid
```
vicb:
```suggestion
    // No need to pass the encoding for buffers
```

packages/open-next/src/http/openNextResponse.ts
vicb:
```suggestion
    this.push(buffer);
```

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -285,13 +285,20 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
+    // official type definition
+    const buffer = encoding === "buffer" ? chunk : Buffer.from(chunk, encoding);
     this.bodyLength += buffer.length;
     if (this.streamCreator?.retainChunks !== false) {
       // Avoid keeping chunks around when the `StreamCreator` supports it to save memory
       this._chunks.push(buffer);
     }
-    this.push(chunk, encoding);
+    // No need to pass the encoding for buffers
+    // unnecessary additional conversion down the stream pipeline.
```
vicb:
```suggestion
```

packages/open-next/src/http/openNextResponse.ts
```diff
@@ -285,13 +285,19 @@ export class OpenNextNodeResponse extends Transform implements ServerResponse {
...
+    // official type definition
+    const buffer = encoding === "buffer" ? chunk : Buffer.from(chunk, encoding);
     this.bodyLength += buffer.length;
     if (this.streamCreator?.retainChunks !== false) {
       // Avoid keeping chunks around when the `StreamCreator` supports it to save memory
       this._chunks.push(buffer);
     }
-    this.push(chunk, encoding);
+    // No need to pass the encoding for buffers
+    // @ts-expect-error TS2345 'buffer' is not in the official type definition
```
vicb:
```suggestion
```

PR #1023, opened 2025-10-23
Replace styled logger with warn from adapters/logger

mhart:
This saves bringing in the `chalk` dependency and saves 18kb in a bundled OpenNext Cloudflare script.


mhart:
Not sure if there are other uses of this that might be bundled in different circumstances? But this is the only use I could see.

conico974 (approved):
Good catch, thanks!!

sommeeeer:
Thanks mhart!

PR #1029, opened 2025-10-28
Fix and enhancement for Next 16

### LINKED ISSUE #1022: Handle segment data in the incremental cache
See https://github.com/vercel/next.js/blob/2ed06f03e810b3a2dcec4b60a452621a08f4b068/packages/next/src/server/lib/incremental-cache/file-system-cache.ts#L204-L231

### LINKED ISSUE #1026: `fetchExternalImage` signature change -> e2e failure
The function takes 3 parameters since https://github.com/vercel/next.js/commit/8d415853f73c72263d24be3ea92b497d41f242f0

It cause the [e2e to fail](https://github.com/opennextjs/opennextjs-aws/actions/runs/18741263805/job/53458587617#step:12:19)

conico974:
Should fix most build failures in Next 16.
Support for incremental prefetching (Fix #1022)
Support for incremental prefetching in the cache interceptor
Fix #1026

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -22,8 +22,8 @@ catalogs:
       specifier: 0.25.4
       version: 0.25.4
     next:
-      specifier: 16.0.0
-      version: 16.0.0
+      specifier: 16.0.1
+      version: 16.0.1
     postcss:
       specifier: 8.4.27
       version: 8.4.27
@@ -73,7 +73,7 @@ importers:
         version: 2.1.1
       next:
         specifier: 'catalog:'
-        version: 16.0.0(react-dom@19.0.0(react@19.0.0))(react@19.0.0)
+        version: 16.0.1(react-dom@19.0.0(react@19.0.0))(react@19.0.0)
       react:
         specifier: 'catalog:'
         version: 19.0.0
@@ -116,7 +116,7 @@ importers:
         version: link:../shared
       next:
         specifier: 'catalog:'
-        version: 16.0.0(react-dom@19.0.0(react@19.0.0))(react@19.0.0)
+        version: 16.0.1(react-dom@19.0.0(react@19.0.0))(react@19.0.0)
       react:
         specifier: 'catalog:'
         version: 19.0.0
@@ -184,7 +184,7 @@ importers:
         version: link:../shared
       next:
         specifier: 'catalog:'
-        version: 16.0.0(react-dom@19.0.0(react@19.0.0))(react@19.0.0)
+        version: 16.0.1(react-dom@19.0.0(react@19.0.0))(react@19.0.0)
       react:
         specifier: 'catalog:'
         version: 19.0.0
@@ -1969,8 +1969,8 @@ packages:
   '@next/env@15.4.0-canary.14':
     resolution: {integrity: sha512-ynXM3n0AEcB1mwoOLgar27s/WoFyX0C8kpbfpc6bylq2rfS+q+KNla1WAVX3QdHyV82KyrqdMQAFOIyTZg4K9A==}
 
-  '@next/env@16.0.0':
-    resolution: {integrity: sha512-s5j2iFGp38QsG1LWRQaE2iUY3h1jc014/melHFfLdrsMJPqxqDQwWNwyQTcNoUSGZlCVZuM7t7JDMmSyRilsnA==}
+  '@next/env@16.0.1':
+    resolution: {integrity: sha512-LFvlK0TG2L3fEOX77OC35KowL8D7DlFF45C0OvKMC4hy8c/md1RC4UMNDlUGJqfCoCS2VWrZ4dSE6OjaX5+8mw==}
 
   '@next/swc-darwin-arm64@15.4.0-canary.14':
     resolution: {integrity: sha512-p62YaNcigaJlZ6IIubZPT+S4N0CXXkjqdIbC2Otr6LLxWsvdkHRgWaPLHauCxWw0zS7jczKY1w4ZfyX9l26sIQ==}
@@ -1978,8 +1978,8 @@ packages:
     cpu: [arm64]
     os: [darwin]
 
-  '@next/swc-darwin-arm64@16.0.0':
-    resolution: {integrity: sha512-/CntqDCnk5w2qIwMiF0a9r6+9qunZzFmU0cBX4T82LOflE72zzH6gnOjCwUXYKOBlQi8OpP/rMj8cBIr18x4TA==}
+  '@next/swc-darwin-arm64@16.0.1':
+    resolution: {integrity: sha512-R0YxRp6/4W7yG1nKbfu41bp3d96a0EalonQXiMe+1H9GTHfKxGNCGFNWUho18avRBPsO8T3RmdWuzmfurlQPbg==}
     engines: {node: '>= 10'}
     cpu: [arm64]
     os: [darwin]
@@ -1990,8 +1990,8 @@ packages:
     cpu: [x64]
     os: [darwin]
 
-  '@next/swc-darwin-x64@16.0.0':
-    resolution: {integrity: sha512-hB4GZnJGKa8m4efvTGNyii6qs76vTNl+3dKHTCAUaksN6KjYy4iEO3Q5ira405NW2PKb3EcqWiRaL9DrYJfMHg==}
+  '@next/swc-darwin-x64@16.0.1':
+    resolution: {integrity: sha512-kETZBocRux3xITiZtOtVoVvXyQLB7VBxN7L6EPqgI5paZiUlnsgYv4q8diTNYeHmF9EiehydOBo20lTttCbHAg==}
     engines: {node: '>= 10'}
     cpu: [x64]
     os: [darwin]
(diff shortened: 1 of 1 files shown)


khuezy:
Do we have to handle the `middleware` => `proxy` renaming in 16?

conico974:
> Do we have to handle the middleware => proxy renaming in 16?

Nope it doesn't change the name of the compiled file, but we have to update to Node 20 for e2e it seems....

packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
```diff
@@ -27,7 +27,8 @@ export async function optimizeImage(
   const { isAbsolute, href } = imageParams;
 
   const imageUpstream = isAbsolute
-    ? await fetchExternalImage(href)
+    ? //@ts-ignore
```
sommeeeer: Is this necessary? If so do an expect-error with description.

examples/app-router/app/og/opengraph-image.tsx
```diff
@@ -31,6 +31,9 @@ export default async function Image() {
       // For convenience, we can re-use the exported opengraph-image
       // size config to also set the ImageResponse's width and height.
       ...size,
+      headers: {
+        "cache-control": "public, immutable, no-transform, max-age=31536000",
```
sommeeeer, line 35: Why did you set this?

examples/app-router/app/isr-data-cache/page.tsx
```diff
@@ -4,7 +4,7 @@ async function getTime() {
   return new Date().toISOString();
 }
 
-const cachedTime = unstable_cache(getTime, { revalidate: false });
+const cachedTime = unstable_cache(getTime, ["getTime"], { revalidate: false });
```
sommeeeer, line 7: Is it still called `unstable_cache`? Isnt it stable yet?

sommeeeer (changes requested):
Very good job Nico! It seems to pass the E2E locally for me on Next 16. 

Just a few comments and also could you rename the middleware files to proxy in all the examples app. (`npx @next/codemod@canary middleware-to-proxy .`)

packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
vicb: Looks like it's related to #1026 (the signature changed with Next 16).
+1 to add a comment about that

packages/open-next/src/build/createAssets.ts
```diff
@@ -169,6 +173,24 @@ export function createCacheAssets(options: buildHelper.BuildOptions) {
       logger.warn(`Skipping invalid cache file: ${cacheFilePath}`);
       return;
     }
+    const segments: Record<string, string> = {};
+
+    // If we have a meta file, and it contains segmentPaths, we need to add them to the cache file
+    if (
+      cacheFileMeta &&
+      Array.isArray(cacheFileMeta.segmentPaths) &&
```
vicb:
```suggestion
      Array.isArray(cacheFileMeta?.segmentPaths) &&
```

vicb: `cacheFileMeta.segmentPaths.length > 0` could also be skipped maybe

vicb: Maybe something like:

```ts
const segments: Record<string, string> = Array.isArray(cacheFileMeta?.segmentPaths) ? 
  Object.fromEntries(cacheFileMeta.segmentPaths.map(...)) :
  {};
```

examples/app-router/app/isr-data-cache/page.tsx
conico974: Actually yeah, it will be superseded by `use cache`, but that's not the same thing

examples/app-router/app/og/opengraph-image.tsx
conico974: Next does not automatically set the cache on this anymore. Need to test with `next start`

khuezy:
Do we have any tests for `User-Agent`, someone reported that it gets dropped: [Discord](https://discord.com/channels/1283128968140161065/1288871672035803236/1434840309325365359)

Could be some weird delimiter character in the user agent that's causing the break, eg `;`

vicb:
> Do we have any tests for `User-Agent`, someone reported that it gets dropped: [Discord](https://discord.com/channels/1283128968140161065/1288871672035803236/1434840309325365359)
> 
> Could be some weird delimiter character in the user agent that's causing the break, eg `;`

Is this related to this PR or should it be a different question/issue?

examples/app-pages-router/next.config.ts
```diff
@@ -3,12 +3,8 @@ import type { NextConfig } from "next";
 const nextConfig: NextConfig = {
   poweredByHeader: false,
   cleanDistDir: true,
-  transpilePackages: ["@example/shared"],
+  // transpilePackages: ["@example/shared"],
```
vicb: Should this be removed completely?

khuezy:
It's unrelated, we can continue the discussion when the person opens a ticket.

examples/app-router/app/api/after/revalidate/route.ts
```diff
@@ -6,7 +6,8 @@ export function POST() {
     () =>
       new Promise<void>((resolve) =>
         setTimeout(() => {
-          revalidateTag("date");
+          // We want to expire the "date" tag immediately
+          revalidateTag("date", { expire: 0 });
```
vicb: nit to make things clearer:

```suggestion          
          revalidateTag("date", { 
            // We want to expire the "date" tag immediately
            expire: 0 
          });
```

examples/app-router/proxy.ts
```diff
@@ -1,7 +1,7 @@
 import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
 
-export function middleware(request: NextRequest) {
+export default function middleware(request: NextRequest) {
```
vicb: nit: adhere to [Next convetions](https://nextjs.org/docs/app/getting-started/proxy#example)
```suggestion
export default function proxy(request: NextRequest) {
```

packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
```diff
@@ -27,7 +27,8 @@ export async function optimizeImage(
   const { isAbsolute, href } = imageParams;
 
   const imageUpstream = isAbsolute
-    ? await fetchExternalImage(href)
+    ? //@ts-expect-error - fetchExternalImage signature has changed in Next.js 16, it has an extra boolean parameter.
+      await fetchExternalImage(href)
```
vicb, line 32: Ref: https://github.com/vercel/next.js/blob/bfe2ab4/packages/next/src/server/image-optimizer.ts#L711

Do we want to include that ref in the comment and explain why we choose
- dangerouslyAllowLocalIP = false 
- count = 3 
with Next 16?

packages/tests-e2e/tests/appRouter/og.test.ts
```diff
@@ -49,8 +49,8 @@ test("next/og (vercel/og) to work in API route", async ({ request }) => {
   const response = await request.get("api/og?title=opennext");
   expect(response.status()).toBe(200);
   expect(response.headers()["content-type"]).toBe("image/png");
-  expect(response.headers()["cache-control"]).toBe(
-    "public, immutable, no-transform, max-age=31536000",
-  );
+  // expect(response.headers()["cache-control"]).toBe(
```
vicb: what do we want to do with those expectations?

pnpm-workspace.yaml
```diff
@@ -3,7 +3,7 @@ packages:
   - "packages/*"
 
 catalog:
-  next: 15.4.2
+  next: 16.0.0
```
vicb: Update to 16.0.1

packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
conico974: I don't think we need to explain why really. This one probably need some big refactor at one point anyway

packages/tests-e2e/tests/appRouter/og.test.ts
conico974: If next doesn't do it anymore by default, we should remove it (just checked, that's the case in `next start`).
But I think it's still a good idea to test to add them manually. I keep it in one and not in the other

.changeset/thirty-bears-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": minor
```
vicb, line 2: Should this PR be a major change and drop support for Next < 14 ?

vicb:
Changes LGTM
~it looks like the lock file is not updated to 16.0.1~ added a fixup commit

.changeset/thirty-bears-tell.md
conico974: I don't think we should, this doesn't change much for earlier versions. Only big difference being Node 20, but I didn't change the esbuild target on purpose. Virtually no difference for earlier versions

vicb: Makes sense !

sommeeeer:
> @sommeeeer I can't retrieve what your requested change was?

It was just this: https://github.com/opennextjs/opennextjs-aws/pull/1029#pullrequestreview-339103772

sommeeeer (approved):
LGTM, gj on this one!

PR #1034, opened 2025-11-10
fix(e2e): Use button instead of checkbox in parallel test

sommeeeer:
After reading through playwright issues there seems to be multiple reported issues with checkboxes and its click API that looks similar to [ours](https://github.com/opennextjs/opennextjs-aws/actions/runs/19126088137/job/54656986882#step:15:39). With this PR we refactor `parallel.test.ts` to use buttons for managing the state.

I first deployed the `/examples` apps and ran the E2E on `main` right now. That seemed to flake on `parallel.test.ts`. 

I then went ahead to deploy with this PR, and now the E2E seems to pass 100%. 

I also reverted [this](https://github.com/opennextjs/opennextjs-aws/pull/1031) PR as it didn't seem to work as we hoped.

### DIFF
--- a/packages/tests-e2e/playwright.config.js
+++ b/packages/tests-e2e/playwright.config.js
@@ -31,14 +31,4 @@ export default defineConfig({
       },
     },
   ],
-  // Workaround for https://github.com/microsoft/playwright/issues/36371
-  // It seems to be failing in our Github action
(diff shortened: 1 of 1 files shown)


vicb (approved):
Thanks for fixing this 🙏

PR #1036, opened 2025-11-13
Optimize header fixing logic and improve tests

conico974:
Refactor header handling to ensure fixes apply only once and only if necessary. Update tests to validate the new behavior of cache-control headers.

This fixes stale cache control header (for ISR) overriding cache control set in the middleware

### DIFF
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -356,9 +356,11 @@ export async function revalidateIfRequired(
  * @__PURE__
  */
 export function fixISRHeaders(headers: OutgoingHttpHeaders) {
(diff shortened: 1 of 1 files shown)


packages/open-next/src/core/routing/util.ts
```diff
@@ -356,6 +356,11 @@ export async function revalidateIfRequired(
  * @__PURE__
  */
 export function fixISRHeaders(headers: OutgoingHttpHeaders) {
+  const regex = /s-maxage=(\d+)/;
```
vicb: nit: do not capture if there is no need (it is slower)
```suggestion
  const regex = /s-maxage=\d+/;
```

sommeeeer (approved):
LGTM

PR #1042, opened 2025-11-21
fix(cache): add null check for globalThis.openNextConfig

DBosley:
## What?
Adds null checks for `globalThis.openNextConfig` in the cache handler to prevent TypeError during Next.js 16 build phase.

## Why?
When using the Next.js 16 Adapters API, the cache handler class is instantiated during SSG/prerendering **before** `globalThis.openNextConfig` is initialized by the runtime handlers. This causes:

```
TypeError: Cannot read properties of undefined (reading 'dangerous')
```

## How?
Changed all occurrences of:
```typescript
if (globalThis.openNextConfig.dangerous?.disableIncrementalCache)
```
To:
```typescript
if (globalThis.openNextConfig?.dangerous?.disableIncrementalCache)
```

Fixed 4 locations in `packages/open-next/src/adapters/cache.ts`:
- `get()` method (line 48)
- `set()` method (line 207)
- `revalidateTag()` method (line 325)
- `updateTagsOnSet()` method (line 433)

## Testing
- Added 3 test cases that verify cache operations work when `globalThis.openNextConfig` is undefined
- All 50 existing cache tests continue to pass

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)

## Related Issues
This is a blocker for the Adapters API work in opennextjs-cloudflare.

### DIFF
--- a/.changeset/fix-opennextconfig-null-check.md
+++ b/.changeset/fix-opennextconfig-null-check.md
@@ -1,0 +1,10 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: add null check for globalThis.openNextConfig in cache handler
+
+Adds optional chaining when accessing globalThis.openNextConfig to prevent
+TypeError during Next.js 16 build phase when using the Adapters API. The cache
+handler can be instantiated during SSG/prerendering before openNextConfig is
+initialized by the runtime handlers.
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -45,7 +45,7 @@ export default class Cache {
           kind?: "FETCH";
         },
   ) {
-    if (globalThis.openNextConfig.dangerous?.disableIncrementalCache) {
+    if (globalThis.openNextConfig?.dangerous?.disableIncrementalCache) {
       return null;
     }
 
@@ -204,7 +204,7 @@ export default class Cache {
     data?: IncrementalCacheValue,
     ctx?: IncrementalCacheContext,
   ): Promise<void> {
-    if (globalThis.openNextConfig.dangerous?.disableIncrementalCache) {
+    if (globalThis.openNextConfig?.dangerous?.disableIncrementalCache) {
       return;
     }
     // This one might not even be necessary anymore
@@ -322,7 +322,7 @@ export default class Cache {
   }
 
   public async revalidateTag(tags: string | string[]) {
-    const config = globalThis.openNextConfig.dangerous;
+    const config = globalThis.openNextConfig?.dangerous;
     if (config?.disableTagCache || config?.disableIncrementalCache) {
       return;
     }
@@ -430,7 +430,7 @@ export default class Cache {
     ctx?: IncrementalCacheContext,
   ) {
     if (
-      globalThis.openNextConfig.dangerous?.disableTagCache ||
+      globalThis.openNextConfig?.dangerous?.disableTagCache ||
       globalThis.tagCache.mode === "nextMode" ||
       // Here it means it's a delete
       !data
--- a/packages/tests-unit/tests/adapters/cache.test.ts
+++ b/packages/tests-unit/tests/adapters/cache.test.ts
@@ -85,6 +85,34 @@ describe("CacheHandler", () => {
       expect(result).toBeNull();
     });
 
+    describe("undefined openNextConfig", () => {
+      it("Should not throw when globalThis.openNextConfig is undefined", async () => {
+        // @ts-expect-error - Testing undefined config scenario
+        globalThis.openNextConfig = undefined;
+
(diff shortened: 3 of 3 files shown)


vicb:
Is this PR AI generated?

DBosley:
> Is this PR AI generated?

@vicb 
I used claude to help me write the PR and to create a comprehensive description, but this wasn't done in an automated way if that's what you're asking. I'm trying to assist with the adapters-api changes, but ran into errors that stemmed from the above issue. This was the solution that seemed to work. I've got it linked to the cloudflare repo locally and this fix is working to get the adapters-api functional

Is the usage of AI tooling a problem? The goal is to be comprehensive in the tests, documentation, PR description, and completeness of the solution.

khuezy:
I won't speak on behalf of the cloudflare side, but on the aws side, there's nothing wrong w/ using AI. Thanks for the PR!

khuezy:
Might be worth investigating if we can move the instantiation of the openextConfig before so it's available at build time to avoid the type checking. I believe we had a similar issue a few months ago where we fixed some out of order function calls.

DBosley:
> Might be worth investigating if we can move the instantiation of the openextConfig before so it's available at build time to avoid the type checking. I believe we had a similar issue a few months ago where we fixed some out of order function calls.

I can take a look to see if that's a possibility instead of just null checking, if it's possible I'll close this and open a new PR

DBosley:
I dug into this to see if we could reorder things to avoid the optional chaining. Unfortunately it's not possible due to how ES module evaluation works.

The issue is that `util.ts` has a top-level `const nextServer = new NextServer.default({...})` which instantiates Next.js immediately during module evaluation. And `requestHandler.ts` imports from `util.ts`, which gets imported by `createMainHandler.ts`, which gets imported by `server-adapter.ts`. So the entire import chain resolves synchronously before we even get to the `await createMainHandler()` call.

The config is set inside `createMainHandler()` using `await import("./open-next.config.mjs")`, which is async. By the time that async import completes and sets `globalThis.openNextConfig`, Next.js has already been instantiated and the cache handler has already been loaded.

I traced through the execution order:
1. Module graph resolves (synchronous) - all imports walk the dependency tree
2. Module code executes (synchronous) - `util.ts` creates Next.js server, which loads cache handler
3. Top-level await (async) - `createMainHandler()` runs and sets the config

So the cache class is defined during step 2, but the config isn't available until step 3.

The only alternatives would be build-time injection (loses runtime config flexibility, and I found an earlier commit where this concept was actually removed `1981a47d`) or lazy Next.js initialization (substantial refactoring, would need to defer instantiation until first request). Both seem like worse solutions.

The only time the null chaining is needed is during the brief module evaluation phase at cold start. After that the config is guaranteed to be set before any requests are handled.

khuezy:
Ah ok, thanks for the investigation.

conico974:
> I dug into this to see if we could reorder things to avoid the optional chaining. Unfortunately it's not possible due to how ES module evaluation works.
> 
> The issue is that `util.ts` has a top-level `const nextServer = new NextServer.default({...})` which instantiates Next.js immediately during module evaluation. And `requestHandler.ts` imports from `util.ts`, which gets imported by `createMainHandler.ts`, which gets imported by `server-adapter.ts`. So the entire import chain resolves synchronously before we even get to the `await createMainHandler()` call.
> 
> The config is set inside `createMainHandler()` using `await import("./open-next.config.mjs")`, which is async. By the time that async import completes and sets `globalThis.openNextConfig`, Next.js has already been instantiated and the cache handler has already been loaded.
> 
> I traced through the execution order:
> 1. Module graph resolves (synchronous) - all imports walk the dependency tree
> 2. Module code executes (synchronous) - `util.ts` creates Next.js server, which loads cache handler
> 3. Top-level await (async) - `createMainHandler()` runs and sets the config
> 
> So the cache class is defined during step 2, but the config isn't available until step 3.
> 
> The only alternatives would be build-time injection (loses runtime config flexibility, and I found an earlier commit where this concept was actually removed `1981a47d`) or lazy Next.js initialization (substantial refactoring, would need to defer instantiation until first request). Both seem like worse solutions.
> 
> The only time the null chaining is needed is during the brief module evaluation phase at cold start. After that the config is guaranteed to be set before any requests are handled.

This is not the problem, this part is not used at all during prerendering/build time. 
The problem is inside the adapter itself (Next is using the provided cache handler at build time for the fetch cache) 

I don't think your solution is the right one, typing should be correct and enforced everywhere. 
There is 2 paths to fix that (and I think we should do both) :
- We need a specific cache handler for the adapter API, and cache interception should become mandatory there. 
- Ideally we'd need to split the `open-next.config.ts` into 2 part, a build and a runtime part. Things like incremental cache or tag cache should be runtime only as we don't need them at build time.

DBosley:
Ok, my bad. I misunderstood the problem the errors pointed to. I'll close this.

PR #1046, opened 2025-12-02
fix: Use URL parsing instead of substring matching in isExternal

javascripter:
## Summary

- Fix `isExternal()` to properly detect external URLs by parsing the URL and comparing the actual host, instead of using naive substring matching
- The previous `url.includes(host)` check incorrectly identified URLs as internal when the request host appeared anywhere in the URL string (path, query params, etc.)
- This broke rewrites to services like rendertron that embed the target URL in the path (e.g., `https://rendertron.example.com/render/https://myapp.com/page`)

## Test plan

- [x] Added unit tests for embedded URL cases
- [x] Added unit tests for host:port matching
- [x] All existing tests pass

I also applied this patch locally to our website (not public) to confirm that this fixes the bug mentioned in this PR.

| URL | Request Host | Old Result | New Result |
|-----|--------------|------------|------------|
| `https://example.com/render/https://myapp.com/page` | `myapp.com` | `false` ❌ | `true` ✓ |

### DIFF
--- a/.changeset/pink-years-teach.md
+++ b/.changeset/pink-years-teach.md
@@ -5,4 +5,4 @@
 fix: Correct external URL detection in isExternal using proper URL parsing
 
 Replaces substring-based host matching with URL parsing to correctly determine whether a rewritten URL is external.
-This fixes an issue where NextResponse.rewrite() would treat certain internal URLs as external when their pathname contained the host as a substring, causing unexpected 404s during middleware rewrites.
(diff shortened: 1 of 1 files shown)


vicb:
You also need to `pnpm lint:fix` to format the files before uploading the PR

javascripter:
@vicb Thanks for reviewing! I've added the changeset and fixed linting errors.

PR #1054, opened 2025-12-03
fix a typo in the e2e workflow

vicb:
"v18" is no more valid, removing this altogether as the version can be found in the details

image


vicb:
Thanks for the review!

PR #1055, opened 2025-12-04
fix: AWS ApiGW v1 (REST) converter incorrectly parses all query parameters as arrays

guerra-eb:
When deploying a NextJS application using OpenNext in AWS Api Gateway v1 (REST), the image optimizer function breaks.

APIGWv1 event payload always includes the incoming query strings in both the `multiValueQueryStringParameter` and `queryStringParameters` attributes, even if they are single valued. 

The converter was parsing all the incoming query parameters as arrays, even if they were single valued. This makes functions like the Image Optimizer to fail as it does not support array parameters. Also, it breaks functions that expect single valued query parameters that were working just fine with OpenNext v2.

This PR fixes the converter, parsing single valued query parameters as strings, and multi valued ones as arrays.

### DIFF
--- a/packages/tests-unit/tests/converters/aws-apigw-v1.test.ts
+++ b/packages/tests-unit/tests/converters/aws-apigw-v1.test.ts
@@ -191,7 +191,7 @@ describe("convertFrom", () => {
         test: "testB",
       },
       multiValueQueryStringParameters: {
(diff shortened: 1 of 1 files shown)


conico974 (approved):
LGTM Thanks

PR #1056, opened 2025-12-05
fix: detect Turbopack builds and set TURBOPACK env accordingly

relsunkaev:
## Summary

Fixes runtime errors when deploying Next.js 16 apps built with Turbopack to AWS Lambda.

## Problem

When building with Turbopack (default in Next.js 16), the standalone output only includes `*-turbo.runtime.prod.js` files (e.g., `pages-turbo.runtime.prod.js`), not the webpack variants (`pages.runtime.prod.js`).

The previous behavior unconditionally set `TURBOPACK=false` in `module.compiled.js`, which caused it to try loading non-existent runtime files at runtime:

```
Error: Cannot find module 'next/dist/compiled/next-server/pages.runtime.prod.js'
Require stack:
- /var/task/node_modules/next/dist/server/route-modules/pages/module.compiled.js
- /var/task/node_modules/next/dist/server/route-modules/pages/vendored/contexts/router-context.js
- /var/task/node_modules/next/dist/client/compat/router.js
```

## Solution

This PR detects whether the build was done with Turbopack by checking the traced files:
- If only `*-turbo.runtime.prod.js` files exist → Turbopack build → set `TURBOPACK=true`
- If non-turbo runtime files exist → Webpack build → set `TURBOPACK=false`

## Testing

Tested with a Next.js 16.0.7 app built with Turbopack and deployed to AWS Lambda via SST. Before the fix: 502 errors. After the fix: 200 OK.

### DIFF
--- a/packages/open-next/src/build/patch/patches/patchEnvVar.ts
+++ b/packages/open-next/src/build/patch/patches/patchEnvVar.ts
@@ -1,5 +1,5 @@
-import { createPatchCode } from "../astCodePatcher.js";
-import type { CodePatcher } from "../codePatcher";
+import { createPatchCode, patchCode } from "../astCodePatcher.js";
+import type { CodePatcher, PatchCodeFn } from "../codePatcher.js";
 
 /**
  * Creates a rule to replace `process.env.${envVar}` by `value` in the condition of if statements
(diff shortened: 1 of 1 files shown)


relsunkaev:
@vicb Not sure what's going on with the workflow not passing but this fixed our Next 16 Turbopack deploys. Currently running this as a pnpm patch.

packages/open-next/src/build/patch/patches/patchEnvVar.ts
```diff
@@ -21,6 +21,22 @@ fix:
   '${value}'
 `;
 
+function isTurbopackBuild(tracedFiles: string[]): boolean {
```
vicb, line 24: it would be nice to extract this into an helper

- so that it can be re-used across different files
- and ultimately moved to Next.js

We use [this check](https://github.com/opennextjs/opennextjs-cloudflare/blob/a780999895c8f89f3228a596cc4c96279ff913bb/packages/cloudflare/src/cli/build/bundle-server.ts#L57) in the cloudflare adapter

packages/open-next/src/build/patch/patches/patchEnvVar.ts
```diff
@@ -39,12 +55,12 @@ export const patchEnvVars: CodePatcher = {
       contentFilter: /process\.env\.NODE_ENV/,
       patchCode: createPatchCode(envVarRuleCreator("NODE_ENV", '"production"')),
     },
-    // This patch will set `TURBOPACK` env to false to avoid loading turbopack related deps at runtime
+    // Turbopack builds only include *-turbo.runtime.prod.js files, so we detect and set accordingly
```
vicb, line 58: nit: I don't think we really care about how we detect a turbopack build in this comment?

Maybe revert to the inlined version of the  code and remove the comment altogether when you add the helper?

vicb (commented):
It would be great to add an e2e test got that (i.e. identify something that is failing without this patch)

vicb:
This was addressed via https://github.com/opennextjs/opennextjs-aws/pull/1087

PR #1057, opened 2025-12-07
Implement PPR support in cache interceptor

conico974:
Enhance the cache interceptor to support PPR by introducing new types and modifying the request handling logic. This allows for immediate response delivery while processing requests in the background. Additionally, improve logging for better debugging.

One thing that doesn't work right now, is with an external middleware, but it might make sense to do it later. Especially if we plan to make the StreamCreator mandatory. Right now it would require to do it at the converter level.

**Only works with the Adapter API**

### DIFF
--- a/packages/open-next/src/core/routing/cacheInterceptor.ts
+++ b/packages/open-next/src/core/routing/cacheInterceptor.ts
@@ -135,6 +135,49 @@ function getBodyForAppRouter(
   }
 }
 
+function createPprPartialResult(
+  event: MiddlewareEvent,
+  localizedPath: string,
+  cachedValue: CacheValue<"cache">,
+  responseBody: string | (() => ReturnType<typeof toReadableStream>),
+  contentType: string,
+): PartialResult {
+  if (cachedValue.type !== "app") {
+    throw new Error("createPprPartialResult called with non-app cache value");
+  }
+
+  return {
+    resumeRequest: {
+      ...event,
+      method: "POST",
+      url: `http://${event.headers.host}${NextConfig.basePath || ""}${
+        localizedPath || "/"
+      }`,
+      headers: {
+        ...event.headers,
+        "next-resume": "1",
+      },
+      rawPath: localizedPath,
+      body: Buffer.from(cachedValue.meta?.postponed || "", "utf-8"),
+    },
+    result: {
+      type: "core",
+      statusCode: event.rewriteStatusCode ?? cachedValue.meta?.status ?? 200,
+      body:
+        typeof responseBody === "string"
+          ? toReadableStream(responseBody)
+          : responseBody(),
+      isBase64Encoded: false,
+      headers: {
+        "content-type": contentType,
+        "x-opennext-ppr": "1",
+        ...cachedValue.meta?.headers,
+        vary: VARY_HEADER,
+      },
+    },
+  };
+}
+
 async function generateResult(
   event: MiddlewareEvent,
   localizedPath: string,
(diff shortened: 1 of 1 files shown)


packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -139,32 +140,105 @@ async function generateResult(
...
     if (isDataRequest) {
       const { body: appRouterBody, additionalHeaders: appHeaders } =
         getBodyForAppRouter(event, cachedValue);
       body = appRouterBody;
       additionalHeaders = appHeaders;
+      if (cachedValue.meta?.postponed) {
+        debug("App router postponed request detected", localizedPath);
+
+        return {
+          resumeRequest: {
```
sommeeeer: Create a helper function that returns a `resumeRequest`

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -258,19 +339,35 @@ export async function cacheInterceptor(
...
   debug("Checking cache for", localizedPath, PrerenderManifest);
 
-  const isISR =
-    Object.keys(PrerenderManifest?.routes ?? {}).includes(
-      localizedPath ?? "/",
-    ) ||
-    Object.values(PrerenderManifest?.dynamicRoutes ?? {}).some((dr) =>
-      new RegExp(dr.routeRegex).test(localizedPath),
-    );
+  const isDynamicISR = Object.values(PrerenderManifest.dynamicRoutes).some(
```
sommeeeer: This one is possibly undefined, and the ones under it

.changeset/spicy-laws-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": minor
+---
+
+Make PPR work with the cache interceptor
```
vicb, line 5: It would be nice to make it clear that it "Only works with the Adapter API" ?

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -134,37 +135,107 @@ function getBodyForAppRouter(
...
+    } else {
+      body = cachedValue.html;
+    }
     type = isDataRequest ? "application/json" : "text/html; charset=utf-8";
   } else {
     throw new Error(
       "generateResult called with unsupported cache value type, only 'app' and 'page' are supported",
     );
   }
+  // close();
```
vicb, line 238: What to do with this?

packages/open-next/src/core/routingHandler.ts
```diff
@@ -235,26 +236,43 @@ export default async function routingHandler(
...
       debug("Cache interception enabled");
-      eventOrResult = await cacheInterceptor(eventOrResult);
-      if (isInternalResult(eventOrResult)) {
-        applyMiddlewareHeaders(eventOrResult, headers);
-        return eventOrResult;
+      const cacheInterceptionResult = await cacheInterceptor(eventOrResult);
+      if (isInternalResult(cacheInterceptionResult)) {
+        applyMiddlewareHeaders(cacheInterceptionResult, headers);
+        return cacheInterceptionResult;
+      } else if (isPartialResult(cacheInterceptionResult)) {
```
vicb, line 253: nit: no else after return?

packages/open-next/src/core/routingHandler.ts
```diff
@@ -235,26 +236,43 @@ export default async function routingHandler(
...
       !isInternalResult(eventOrResult)
     ) {
       debug("Cache interception enabled");
-      eventOrResult = await cacheInterceptor(eventOrResult);
-      if (isInternalResult(eventOrResult)) {
-        applyMiddlewareHeaders(eventOrResult, headers);
-        return eventOrResult;
+      const cacheInterceptionResult = await cacheInterceptor(eventOrResult);
+      if (isInternalResult(cacheInterceptionResult)) {
+        applyMiddlewareHeaders(cacheInterceptionResult, headers);
```
vicb, line 251: q: would it make sense for `applyMiddlewareHeaders()` to return the first param?

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -134,37 +135,107 @@ function getBodyForAppRouter(
...
+        );
+      } else {
+        body = cachedValue.html;
+      }
     }
     type = isDataRequest ? "text/x-component" : "text/html; charset=utf-8";
   } else if (cachedValue.type === "page") {
     isDataRequest = Boolean(event.query.__nextDataReq);
-    body = isDataRequest ? JSON.stringify(cachedValue.json) : cachedValue.html;
+    if (isDataRequest) {
```
vicb, line 227: if we want to change the ternary to an if then it would make sense to fold l 232 here as well?

.changeset/spicy-laws-fold.md
vicb: however I am not sure to understand from the code why/where the logic is scoped to the adapter handler, could you expand on that?

conico974: It's on the Next side, basically the resume endpoint is only available using the adapter (or in minimal mode or with some patch)

packages/open-next/src/core/routing/cacheInterceptor.ts
conico974: I'll remove it

packages/open-next/src/core/routingHandler.ts
conico974: It's an else if, there is a third case that none of these 2 are supposed to catch (when it is an InternalEvent)

packages/open-next/src/core/routingHandler.ts
conico974: Yeah, I guess. I'll have to check everywhere where it's used, but yeah

conico974:
Not in this repo anymore

PR #1061, opened 2025-12-11
Bump Next and React to fix vulnerabilities

vicb:


vicb:
Thanks for the review Nico!

PR #1070, opened 2025-12-19
Next 16 takes an array of tags in ComposableCacheHandler#getExpiration

vicb:
`...tags` in 15
https://github.com/vercel/next.js/blob/c5de33e93ccccaf3bee60cf50603e2152f9886e1/packages/next/src/server/lib/cache-handlers/types.ts#L109

`tags` in 16
https://github.com/vercel/next.js/blob/92f72373644fae3a2822386480e9e2b6fcf3df68/packages/next/src/server/lib/cache-handlers/types.ts#L73

Validated with the cloudflare adapter

### DIFF
--- a/.changeset/young-parents-take.md
+++ b/.changeset/young-parents-take.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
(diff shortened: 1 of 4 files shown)


vicb:
Thanks Nico!

PR #1072, opened 2025-12-31
Fix Next 16.1 not including skipTrailingSlashRedirect

conico974:
Update Next.js to version 16.1.1 and implement a patch to dynamically locate the `next.config` file, ensuring compatibility with the `skipTrailingSlashRedirect` requirement.

### DIFF
--- a/packages/open-next/src/build/patch/patches/patchNextConfig.ts
+++ b/packages/open-next/src/build/patch/patches/patchNextConfig.ts
@@ -11,7 +11,8 @@ export const patchNextConfig: CodePatcher = {
       versions: ">=16.1.0",
       patchCode: async ({ code, buildOptions }) => {
         // Find the next.config file with any supported extension
-        const possibleExtensions = [".ts", ".mjs", ".js", ".cjs"];
+        const tsExtensions = [".ts", ".mts", ".cts"];
+        const possibleExtensions = [...tsExtensions, ".mjs", ".js", ".cjs"];
         let configPath: string | undefined;
         let configExtension: string | undefined;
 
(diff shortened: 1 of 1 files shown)


packages/open-next/src/build/patch/patches/patchNextConfig.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const patchNextConfig: CodePatcher = {
+  name: "patch-next-config",
+  patches: [
+    {
+      pathFilter: /required\-server\-files\.json$/,
+      versions: ">=16.1.0",
+      patchCode: async ({ code, buildOptions }) => {
+        // Find the next.config file with any supported extension
+        const possibleExtensions = [".ts", ".mjs", ".js", ".cjs"];
```
vicb: looks like `.mts` is also supported from the Next repo.

line 34 should be updated below as well.

Maybe use `{extension: isTypeScript}` instead of an array here?

packages/open-next/src/build/patch/patches/patchNextConfig.ts
```diff
@@ -0,0 +1,62 @@
...
+          configToImport = path.join(
+            buildOptions.tempBuildDir,
+            "next.config.mjs",
+          );
+        } else {
+          // For .js, .mjs, .cjs, use the file directly
+          configToImport = configPath;
+        }
+
+        // In next 16.1+ we need to add `skipTrailingSlashRedirect` manually because next removes it from the config
```
vicb, line 52: Q: do you think this should be reported as an issue on the Next repo?

conico974: I guess we could, but I think they did it on purpose. Looks like they use define for that now, and I checked, it is defined in the adapter output

packages/open-next/src/build/patch/patches/patchNextConfig.ts
vicb: https://nextjs.org/docs/pages/api-reference/config/typescript#for-commonjs-projects-default also mentions `.cts` in a note, not sure if really needed?

Issue #1066, opened 2025-12-12
Issue with image optimization function bundling with NodeJS 24 that has symbolic links


ankur-jain-nhs:
When deploying with NodeJS 24, terraform archiver complains that it could not find the symlinked path, which did not exist anyway. In NodeJS 22, the path contained actual files.

Folder : `.open-next/image-optimization-function/node_modules/.bin`

Related issues: -
Open - https://github.com/nodejs/node/issues/59168
Closed - https://github.com/nodejs/node/issues/58939
Previous fix - https://github.com/opennextjs/opennextjs-aws/pull/934/files

andrewdavidcostello:
We've also encountered a similar issue, broken symlinks are being created and therefore at the moment we need to manually run a script to fix the symlinks. The one it actually complains about is .bin/semver.

cbravo:
@andrewdavidcostello can you tell me a bit about your script you are running to fix this? having the same issue about .bin/semver in our pipeline

andrewdavidcostello:
@cbravo Happy to given lost most of my afternoon on this one! We now have a custom build command

`buildCommand: 'npx --yes open-next@^3 build && node fix-symlinks.js',`

```
#!/usr/bin/env node
/**
 * Fix broken symlinks in .open-next directory
 * OpenNext creates symlinks to temp directories that get deleted
 * This script replaces symlinks with actual file copies
 */
const fs = require('fs');
const path = require('path');

function dereferenceSymlink(symlinkPath) {
  try {
    const stats = fs.lstatSync(symlinkPath);
    if (!stats.isSymbolicLink()) {
      return false;
    }
    
    const target = fs.readlinkSync(symlinkPath);
    const resolved = path.resolve(path.dirname(symlinkPath), target);
    
    if (!fs.existsSync(resolved)) {
      console.warn(`Broken symlink: ${symlinkPath} -> ${target}`);
      return false;
    }
    
    // Read the target file
    const content = fs.readFileSync(resolved);
    const targetStats = fs.statSync(resolved);
    
    // Remove the symlink
    fs.unlinkSync(symlinkPath);
    
    // Write the actual file content
    fs.writeFileSync(symlinkPath, content, { mode: targetStats.mode });
    
    return true;
  } catch (e) {
    console.error(`Error processing ${symlinkPath}:`, e.message);
    return false;
  }
}

function processDirectory(dirPath) {
  let fixed = 0;
  
  try {
[...]

PR #1074, opened 2026-01-06
fix(installDeps): Ensure symlinks are dereferenced on all Node versions

sommeeeer:
- [x] Test on an affected repo

I think we should always use the workaround to dereference symlinks manually. Keeping an up to date list of all the affected Node versions would become very tedious as it seems that this issue will be present in upcoming Node versions aswell.

(We could also skip copying the `.bin` folder entirely as its mostly used for CLI anyways. In a serverless function that doesn't make sense TBH)  

Closes https://github.com/opennextjs/opennextjs-aws/issues/1066

Upstream issue in Node: https://github.com/nodejs/node/issues/59168

### DIFF
--- a/packages/open-next/src/build/installDeps.ts
+++ b/packages/open-next/src/build/installDeps.ts
@@ -78,7 +78,7 @@ export function installDependencies(
       }
     }
     // End of Node Workaround
-    
+
     // Cleanup tempDir
     fs.rmSync(tempInstallDir, { recursive: true, force: true });
     logger.info(`Dependencies installed for ${name}`);


vicb:
~@sommeeeer could you please add a changeset too?~
Sorry it's here already

sommeeeer:
> Any reason why the PR is in draft state?

I haven't had time to test it out on a real repo yet. Can do it later tonight

sommeeeer:
This one is ready to ship now!

vicb:
Thanks @sommeeeer !

PR #1075, opened 2026-01-08
override and update deps with security issues

james-elicx:
This PR bumps / overrides certain deps to address the following security issues flagged in a Snyk scan.

```
  ✗ Allocation of Resources Without Limits or Throttling [Medium Severity][https://security.snyk.io/vuln/SNYK-JS-BODYPARSER-14105059] in body-parser@2.2.0
    introduced by express@5.1.0 > body-parser@2.2.0
  This issue was fixed in versions: 2.2.1
  ✗ Regular Expression Denial of Service (ReDoS) [Low Severity][https://security.snyk.io/vuln/SNYK-JS-BRACEEXPANSION-9789073] in brace-expansion@2.0.1
    introduced by @node-minify/core@8.0.6 > glob@9.3.5 > minimatch@8.0.4 > brace-expansion@2.0.1
  This issue was fixed in versions: 1.1.12, 2.0.2, 3.0.1, 4.0.1
  ✗ Improper Input Validation [Medium Severity][https://security.snyk.io/vuln/SNYK-JS-NANOID-8492085] in nanoid@3.3.7
    introduced by next@16.0.10 > postcss@8.4.31 > nanoid@3.3.7
  This issue was fixed in versions: 3.3.8, 5.0.9
  ✗ Allocation of Resources Without Limits or Throttling [High Severity][https://security.snyk.io/vuln/SNYK-JS-QS-14724253] in qs@6.14.0
    introduced by express@5.1.0 > qs@6.14.0 and 1 other path(s)
  This issue was fixed in versions: 6.14.1
```

It also bumps the pnpm version used in the pipelines so that catalog overrides are respected.

### DIFF
--- a/packages/open-next/package.json
+++ b/packages/open-next/package.json
@@ -50,7 +50,7 @@
     "chalk": "^5.6.2",
     "cookie": "^1.0.2",
     "esbuild": "catalog:",
-    "express": "^5.1.0",
+    "express": "^5.2.1",
     "path-to-regexp": "^6.3.0",
     "urlpattern-polyfill": "^10.1.0",
     "yaml": "^2.8.1"
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -14,3 +14,8 @@ catalog:
   tailwindcss: 3.3.3
   typescript: 5.9.3
   esbuild: 0.25.4
+
+overrides:
+  "qs@6.14.0": ^6.14.1
+  "brace-expansion@2.0.1": ^2.0.2
+  "nanoid@3.3.7": ^3.3.8
--- a/.changeset/crisp-dolls-go.md
+++ b/.changeset/crisp-dolls-go.md
@@ -1,0 +1,8 @@
+---
+"@opennextjs/aws": patch
+---
+
+override and update deps with security issues
+
+- bumps express version used for the express-dev wrapper.
+- overrides patch versions used for qs, brace-expansion, and nanoid.
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -27,7 +27,7 @@ runs:
     # Install pnpm. https://github.com/pnpm/action-setup
     - uses: pnpm/action-setup@v4
       with:
-        version: 9
+        version: 10
         # run_install: false
 
     # Get pnpm store path so we can cache it
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -90,7 +90,7 @@ jobs:
       - name: Setup pnpm
         uses: pnpm/action-setup@v4
         with:
-          version: 9
+          version: 10
 
       - name: Set up NodeJS
         uses: actions/setup-node@v4
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -40,6 +40,11 @@ catalogs:
       specifier: 5.9.3
       version: 5.9.3
 
+overrides:
+  qs@6.14.0: ^6.14.1
+  brace-expansion@2.0.1: ^2.0.2
+  nanoid@3.3.7: ^3.3.8
+
 importers:
 
   .:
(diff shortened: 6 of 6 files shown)


pnpm-workspace.yaml
```diff
@@ -14,3 +14,8 @@ catalog:
   tailwindcss: 3.3.3
   typescript: 5.9.3
   esbuild: 0.25.4
+
+overrides:
```
vicb, line 18: Not a big fan of overrides...
Can we figure out if we depend on them at runtime?

james-elicx: `qs` and `body-parser` are for express which is used by the `express-dev` wrapper and unlikely to ever be used by a consumer. `brace-expansion` is a deeply nested dependency of `@node-minify/core` which is used to minify JS in the server bundle.

These overrides are 1 patch version above the version that is currently depended on because those patches were released solely to resolve the security issue in their respective libraries.

packages/open-next/package.json
dacevedo12, line 41: BTW, is this exact pinning really necessary? https://github.com/advisories/GHSA-6475-r3vj-m8vf

conico974: Aws has proven to be not very good at respecting semver, they had some big breaking changes at times (that broke thing for a few versions and was fixed after)
I think we should keep pinning, but just update the SDK and ensure that everything works, maybe in another PR

packages/open-next/package.json
```diff
@@ -50,7 +50,7 @@
     "chalk": "^5.6.2",
     "cookie": "^1.0.2",
     "esbuild": "catalog:",
-    "express": "^5.1.0",
+    "express": "^5.2.1",
```
vicb, line 53: I think express should be moved to the dev dependencies.

And because it's only used in dev, we should not have to override any package for security reason (`qs` here) WDYT?

pnpm-workspace.yaml
```diff
@@ -14,3 +14,8 @@ catalog:
...
+  "qs@6.14.0": ^6.14.1
+  "brace-expansion@2.0.1": ^2.0.2
```
vicb, line 20: > brace-expansion is a deeply nested dependency of @node-minify/core which is used to minify JS in the server bundle.

IIUC correctly, this is used to [minify JSON only](https://github.com/opennextjs/opennextjs-aws/blob/4255fb71a2e513d1c69c16d4e8da396375dd4f6e/packages/open-next/src/build/createServerBundle.ts#L393). I don't think there is much to minimize in JSON files. 

So what we could do instead is to drop the dep and use `parse` followed by `stringify(o, null, 0)`.

@conico974 anything I miss here?

pnpm-workspace.yaml
```diff
@@ -14,3 +14,8 @@ catalog:
...
+  "nanoid@3.3.7": ^3.3.8
```
vicb, line 21:
```
-> % pnpm -F  @opennextjs/aws why nanoid         
Legend: production dependency, optional only, dev only

@opennextjs/aws@3.9.11 /Users/vberchet/code/work/aws/packages/open-next

dependencies:
next 16.0.10
└─┬ postcss 8.4.31
  └── nanoid 3.3.7
```

I would say that no override is necessary here.
`postcss` is build time.

IMO the best thing to do here is ask Next to update postcss if not done in their latest version?

james-elicx:
No worries, I'll close this if there isn't much appetite for it.

For context, at work we have a policy of removing issues flagged by security tools from our apps wherever possible (we don't use OpenNext, but if we did, these would be flagged by one of our tools and we'd be asked to do something about them), so it's common to override deeply nested dependencies if there's a patched non-breaking version.

> If we override 9 with 10 but the original package update to 11 then we will still be pinned at 10

I don't think this would be the case because the override is only for applying to a specific version of the dependency - if they upgrade, the override wouldn't apply anymore as the version is different to the one it's set to apply to.

vicb:
> I don't think this would be the case because the override is only for applying to a specific version of the dependency - if they upgrade, the override wouldn't apply anymore as the version is different to the one it's set to apply to.

Oh right, I missed that

PR #1077, opened 2026-01-17
fix: require.resolve throwing exception in next 16.1.0 config patch

james-elicx:
The compiled Next.js configs created in the patch preserves `require.resolve` calls as esbuild does not perform static analysis on them.

This creates a problem when we do a dynamic import on the compiled config as it is unable to resolve the import due to the `require` polyfill not providing a function to resolve imports.

This change inlines the `require.resolve` statements so that we are able to dynamically import the Next.js config again. This should not be problematic as we only compile the Next.js config so that we can import it to override the trailing slash option in our stringified version, and require.resolve calls being inlined should not impact the result.

This bug was found when trying to build next-app-router-playground on 16.1.x due to their [use of `@next/mdx`](https://github.com/vercel/next-app-router-playground/blob/main/next.config.ts).

**Before**
image

**After**
image

**Example problematic require.resolve**
image

### DIFF
--- a/packages/open-next/src/plugins/inline-require-resolve.ts
+++ b/packages/open-next/src/plugins/inline-require-resolve.ts
@@ -2,6 +2,14 @@ import fs from "node:fs";
 import { createRequire } from "node:module";
 import type { Plugin } from "esbuild";
 
+/**
+ * Inlines calls to `require.resolve` in JavaScript files.
+ *
+ * esbuild does not statically analyse `require.resolve` calls, and the polyfill
+ * does not include an implementation to handle them. This can be problematic
+ * if you attempt to dynamically import a file built by esbuild that unknowingly
+ * contains `require.resolve` calls, as they will throw an error during import.
+ */
 export const inlineRequireResolvePlugin: Plugin = {
   name: "inline-require-resolve",
   setup: (build) => {


packages/open-next/src/build/patch/patches/patchNextConfig.ts
```diff
@@ -33,12 +34,40 @@ export const patchNextConfig: CodePatcher = {
...
+              {
+                name: "inline-require-resolve",
+                setup: (build) => {
+                  build.onLoad(
+                    { filter: /\.(js|ts|mjs|cjs)$/ },
+                    async (args) => {
+                      const transformed = fs
+                        .readFileSync(args.path, "utf-8")
+                        .replace(
+                          /require\.resolve\(['"]([^'"]+)['"]\)/g,
```
vicb: What about `\((?<quote>['"])((?:\\.|.)*?)\k<quote>\)`

- look for a closing quote matching the opening quote
- make sure it is not esxpared

packages/open-next/src/build/patch/patches/patchNextConfig.ts
```diff
@@ -33,12 +34,40 @@ export const patchNextConfig: CodePatcher = {
...
         // Only compile if the extension is a TypeScript extension
         if (tsExtensions.includes(configExtension)) {
-          buildSync({
+          await build({
             entryPoints: [configPath],
             outfile: path.join(buildOptions.tempBuildDir, "next.config.mjs"),
             bundle: true,
             format: "esm",
             platform: "node",
+            plugins: [
```
vicb: Should it be a plugin in packages/open-next/src/plugins?

packages/open-next/src/build/patch/patches/patchNextConfig.ts
```diff
@@ -33,12 +34,40 @@ export const patchNextConfig: CodePatcher = {
...
+              {
+                name: "inline-require-resolve",
+                setup: (build) => {
+                  build.onLoad(
+                    { filter: /\.(js|ts|mjs|cjs)$/ },
+                    async (args) => {
+                      const transformed = fs
+                        .readFileSync(args.path, "utf-8")
```
vicb: I think ESBuild recommend using the async APIs

packages/open-next/src/build/patch/patches/patchNextConfig.ts
vicb: or ast grep if the parameter must be a string literral

james-elicx: switched to use your regex.

i don't think ast-grep can be used since this requires on running some js to figure out the fix, but happy to be corrected if it can

vicb: It could, the api gives you access to the matches so I think there should be no problem doing it.
But I agree that it's not because we can than we should. If you think the regexp is good enough, no need to overcomplicate.

james-elicx: Yes I think regex does a good enough job for the problem in this case

packages/open-next/src/plugins/inline-require-resolve.ts
```diff
@@ -0,0 +1,24 @@
+import fs from "node:fs";
+import { createRequire } from "node:module";
+import type { Plugin } from "esbuild";
+
+export const inlineRequireResolvePlugin: Plugin = {
```
vicb, line 13: Maybe add a comment about what this plugin is doing and when it is needed

PR #1080, opened 2026-01-19
refactor: lazily compile Next config

vicb:
The current code pro-actively compiles the Next Config.

The PR only compiles it when needed - that is when `skipTrailingSlashRedirect` is not present in the manifest.

Note that the behavior should be the same as of today because the patch is only executed for Next >= 16.1 where `skipTrailingSlashRedirect` is always missing. But hopefully it can be added back at some point.

/cc @ijjk


vicb:
Thanks Nico!

PR #1081, opened 2026-01-19
fix: make sure the edge adapter loads a patched Next config

vicb:
Before this PR, the Next Config was patched in `packages/open-next/src/build/createServerBundle.ts`.

This causes an issue when the middleware is external as the middleware would load the config from the `.next` folder and this config is not patched so it still does not have `skipTrailingSlashRedirect`.

This PR does patch the original manifest in the `.next` folders. Definitely not really clean by @ijjk thinks the regression has been introduced by https://github.com/vercel/next.js/pull/86830 and the patch should no more be needed after Next fixes that.

Validated with the cloudflare adapter

see: https://github.com/opennextjs/opennextjs-aws/pull/1080

### DIFF
--- a/packages/open-next/src/build/patch/patches/patchOriginalNextConfig.ts
+++ b/packages/open-next/src/build/patch/patches/patchOriginalNextConfig.ts
@@ -10,6 +10,7 @@ import * as buildHelper from "../../helper.js";
  * This patch adds it back in by compiling and importing the user's `next.config.js` file.
  *
  * It is a regression in https://github.com/vercel/next.js/pull/86830
+ * TODO(#1082): revisit when Next is fixed
  */
 export async function patchOriginalNextConfig(
   options: buildHelper.BuildOptions,


vicb:
> Wasn't it enough to just change nextDir here to point to the patched version
> 
> https://github.com/opennextjs/opennextjs-aws/blob/15f720951ecf1e888ab8f7196ba0fecfc3caf89b/packages/open-next/src/plugins/edge.ts#L159
> 
> 
> I'd prefer to not modify the original NextConfig if possible

I don't think because the middleware is compiled before `createServerBundle` is called and this where the patch was applied on copied traced files.

But let me know if I'm wrong here

packages/open-next/src/build/patch/patches/patchOriginalNextConfig.ts
```diff
@@ -0,0 +1,110 @@
...
+ * Next 16.1+ removes `skipTrailingSlashRedirect` from the config in `required-server-files.json`.
+ *
+ * This patch adds it back in by compiling and importing the user's `next.config.js` file.
+ *
+ * It is a regression in https://github.com/vercel/next.js/pull/86830
+ */
+export async function patchOriginalNextConfig(
+  options: buildHelper.BuildOptions,
+): Promise<void> {
+  if (buildHelper.compareSemver(options.nextVersion, "<", "16.1.0")) {
```
conico974, line 18: Could we add a TODO here that we want to remove this as soon as it is fixed in Next (maybe create an issue as well)

vicb: I have created #1082 and added a link in `patchOriginalNextConfig` comment.

vicb:
Thanks for the review Nico!

PR #1083, opened 2026-01-19
Add serverExternalPackages to required-server-files.json manifest

conico974:
Include serverExternalPackages in the manifest to enhance configuration options for server dependencies.


vicb (approved):
Did you check if anything else was missing?

conico974:
> Did you check if anything else was missing?

On our side I think that's it, there might be some other things that are missing that we don't use. Or maybe I missed something :wink: 
I won't make a release tonight anyway, so if something is missing we can add it later

PR #1086, opened 2026-01-20
export `findPackagerAndRoot` utility

dario-piotrowicz:
This PR exports the `findMonorepoRoot` utility so that it can be imported and used externally.

The use case is so that in the Cloudflare adapter we can use the utility to detect the package manager in use.

For example like so:
```ts
const packageManager = await import("@opennextjs/aws/build/helper.js").then(m =>
	findMonorepoRoot("./").packager
);
```

### DIFF
--- a/.changeset/good-flies-hear.md
+++ b/.changeset/good-flies-hear.md
@@ -2,4 +2,4 @@
 "@opennextjs/aws": patch
 ---
 
-export the `findPackagerAndRoot` utility externally
+export the `findPackagerAndRoot` utility


packages/open-next/src/build/helper.ts
```diff
@@ -65,7 +65,7 @@ export function normalizeOptions(
   };
 }
 
-function findMonorepoRoot(appPath: string) {
+export function findMonorepoRoot(appPath: string) {
```
vicb: maybe rename this to `findPmAndMonorepoRoot` now that it is exported?
Also add JSDoc and return type?

.changeset/good-flies-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": minor
```
vicb: we usually use patch for exporting utils to cf.

@conico974 what do you think?

packages/open-next/src/build/helper.ts
dario-piotrowicz: done 👍

.changeset/good-flies-hear.md
dario-piotrowicz: I have no idea/preference here, I'm happy to set this to whatever you prefer / is most consistent with other such changes 🙂

packages/open-next/src/build/helper.ts
```diff
@@ -64,8 +64,14 @@ export function normalizeOptions(
     tempBuildDir,
   };
 }
-
-function findMonorepoRoot(appPath: string) {
+/**
+ * Given the path to a project this function detects the project's repository root (wether the project is in a simple
```
khuezy:
```suggestion
 * Given the path to a project this function detects the project's repository root (whether the project is in a simple
```

dario-piotrowicz: Thanks 🙂  I _always_ get that wrong 😓

.changeset/good-flies-hear.md
conico974: Yeah these kind of thing are very "internal", so I think it's better to keep it as a patch for now.
Once we have a monorepo and a shared core, it will make sense to have these kind of change as minor for the core package itself.

vicb:
@dario-piotrowicz could you please update as @khuezy suggested above and run `pnpm lint:fix`
We'll most probably do a release later today with another fix

.changeset/good-flies-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+export the `findPackagerAndRoot` utility externally
```
vicb:
```suggestion
export the `findPackagerAndRoot` utility
```

PR #1087, opened 2026-01-21
fix: use the turbopack runtime

vicb:
`patchEnvVars` would always assume TURBOPACK == false which would result in using the webpack runtime in TURBOPACK builds

Before:

image

After:

image

### DIFF
--- a/.changeset/seven-moles-pull.md
+++ b/.changeset/seven-moles-pull.md
@@ -2,4 +2,4 @@
 "@opennextjs/aws": patch
 ---
 
-fix: use the turbopack runtime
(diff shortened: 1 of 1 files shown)


.changeset/seven-moles-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: use the turbopack runtime
```
conico974:
```suggestion
fix: use the correct runtime (i.e. webpack or turbopack)
```

conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review Nico!

PR #1089, opened 2026-01-21
perf: parallelize build bundles

Kyujenius:
## Summary
- Parallelize independent bundle creation operations in `build.ts` using `Promise.all`
- Each bundle function (`createServerBundle`, `createRevalidationBundle`, `createImageOptimizationBundle`, `createWarmerBundle`) writes to separate directories, making them safe to run concurrently

## Background
Initially attempted to parallelize async operations across multiple files:
- `build.ts` - bundle creation (build-time, heavy I/O)
- `middleware.ts` - resolver initialization (runtime)
- `createMainHandler.ts` - resolver initialization (runtime)
- `generateOutput.ts` - override extraction (build-time, minimal I/O)

After analysis, reverted parallelization for files with minimal I/O benefit:

| File | Execution | I/O Cost | Parallelization Benefit |
|------|-----------|----------|------------------------|
| `build.ts` | Build-time | **Heavy** (esbuild, fs operations) | ✅ Significant |
| `middleware.ts` | Runtime | Minimal (cached imports) | ❌ Negligible |
| `createMainHandler.ts` | Runtime | Minimal (cached imports) | ❌ Negligible |
| `generateOutput.ts` | Build-time | Minimal (string comparisons) | ❌ Negligible |

**Why not parallelize `generateOutput.ts`?**
Although it runs at build-time, its async functions (`canStream`, `extractOverrideName`, etc.) mostly perform string comparisons or call already-loaded functions - no real disk/network I/O. The added code complexity outweighs the minimal performance gain.

Kept only `build.ts` parallelization where real I/O operations (esbuild bundling, file system operations) provide measurable benefit.

## Future Consideration
If the team prefers consistency, I can add a follow-up commit to parallelize all build-time async operations (including `generateOutput.ts`) for uniformity. Let me know!

### `packages/open-next/src/build.ts`
```diff
-  await createServerBundle(options);
-  await createRevalidationBundle(options);
-  await createImageOptimizationBundle(options);
-  await createWarmerBundle(options);
+  await Promise.all([
+    createServerBundle(options),
+    createRevalidationBundle(options),
+    createImageOptimizationBundle(options),
+    createWarmerBundle(options),
+  ]);
```

## Test plan
- [x] Run `pnpm build` to verify build completes successfully
- [x] Verify all bundle directories are created correctly
- [x] Run existing test suite

---

### DIFF
--- a/.changeset/parallelize-build-bundles.md
+++ b/.changeset/parallelize-build-bundles.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/aws": patch
+---
+
+perf: parallelize build bundle creation
+
+Parallelize independent bundle creation operations (`createServerBundle`, `createRevalidationBundle`, `createImageOptimizationBundle`, `createWarmerBundle`) using `Promise.all`. Each bundle writes to separate directories, making them safe to run concurrently. This should improve build time by running esbuild bundling and file system operations in parallel.


vicb:
Thanks for the PR.
What perf gain are you seeing with that? (for a full build)

kyu-rong:
@vicb  Thanks for your comment!

Actually, I opened this PR because I noticed what looked like an anti-pattern while reading the code - four independent async functions being awaited sequentially. Intuitively, `Promise.all` seemed like the obvious fix.

But when I actually benchmarked it, something interesting happened.

### Benchmark Setup
- Next.js 15.1.0 app with 51 routes (I made it for a quick bench test)
- MacBook M1 Pro (Apple Silicon)

### Main branch (Sequential)

```bash
[0ms]    createServerBundle START
[2462ms] createServerBundle END          (2.5s)
[2462ms] createRevalidationBundle START
[2544ms] createRevalidationBundle END    (0.08s)
[2544ms] createImageOptimizationBundle START
[4759ms] createImageOptimizationBundle END (2.2s)
[4759ms] createWarmerBundle START
[4860ms] createWarmerBundle END          (0.1s)
[4860ms] All bundles complete            → Total: 4.86s
```


### PR branch (Parallel)
```bash
[0ms]    All 4 functions START simultaneously
[2308ms] createWarmerBundle END
[2314ms] createRevalidationBundle END
[5698ms] createImageOptimizationBundle END
[5699ms] createServerBundle END
[5699ms] All bundles complete            → Total: 5.70s
```


### Result

| Branch | Bundle Creation Time |
|--------|---------------------|
| main (sequential) | **4.86s** |
| PR (parallel) | **5.70s** |

Parallel is actually **~17% slower** 😅

### Why?

esbuild already utilizes multiple CPU cores internally for maximum performance. 
When we run one esbuild process at a time (sequential), it can fully utilize all available cores. But when we run 4 esbuild processes simultaneously (parallel), they end up competing for the same CPU resources.

The result: functions that took 0.08s~0.1s when run alone ended up taking 2+ seconds when fighting for CPU time with other esbuild processes.

### My take

This was a good learning experience! The "anti-pattern" I spotted isn't actually an anti-pattern here - sequential execution is the right choice when each task already maximizes CPU utilization internally.

Could you verify if you see similar results on your machine? If so, I think closing this PR would be the right call. 

Sometimes the obvious optimization isn't one 🙂

vicb:
I asked because parallel code could have tricky side effects (transient issues) so if there is no huge gain, I think it's not worth it.

kyu-rong:
You're right - learned that parallel isn't always better, Closing this PR now. Thanks for your time and the insightful feedback! 🙏

PR #1091, opened 2026-01-22
fix the detection of the runtime (webpack vs turbopack)

vicb:
The path to the runtime is different for SSR apps.

Validated that:
- it works for a SSR app
- webpack runtime is still at the same location for SSR apps


vicb:
Thanks Nico!

PR #1094, opened 2026-01-26
Check for supported Next version

vicb:
Note that this will error on unkown Next version.

I think it is a feature for when Next 17 is released (we have to explicitly marked is as supported)

Some of the code was generated by Claude/Opus


vicb:
Thanks James!

PR #1097, opened 2026-01-26
bump Next and React version for CVE-2026-23864

vicb:


vicb:
Thanks for the reviews!

PR #1101, opened 2026-01-29
bump `@ast-grep/napi`

### LINKED ISSUE #1067: Critical Regression: v3.9.x Build Failure on Older Linux CI Runners (glibc incompatibility) — Requesting Security Patch in v3.7.x (e.g., v3.7.9)
Dear OpenNext maintainers,

We are reporting a critical blocking issue that prevents us from adopting the latest security fixes for Next.js/React vulnerabilities because of a regression in compatibility in the `v3.9.x` branch.

We need to update to a secured version of OpenNext.

#### The Problem & Impact

1.  **Vulnerable Version:** We are currently on the version **v3.7.6**.
2.  **Latest Version Fails:** Upgrading to **v3.9.6** (which includes the security fixes) causes a build failure in our Continuous Integration (CI) environment.
3.  **Error Root Cause:** The failure is due to a change in the native binary dependency, `@ast-grep/napi` (v0.40.x), which now requires a modern C library (`glibc 2.28+`). Our widely-used Linux CI runners (Amazon Linux 2 / older Ubuntu, Arch: `arm64`) do not meet this requirement.

**Error Snippet:**

```
Error: /.../node_modules/@ast-grep/napi-linux-arm64-gnu/ast-grep-napi.linux-arm64-gnu.node: undefined symbol: le16toh
```

This regression prevents all users on older, stable Linux infrastructures from applying the necessary security updates.

We request a **security backport** of the patches for the Next.js/React vulnerabilities (e.g., CVE-2025-55182) to the **v3.7.x branch**.

We ask for a new patch release, such as **v3.7.9**, which contains the necessary security updates but retains the dependency versions and compilation compatibility of the `v3.7.x` branch.

This approach would allow us and other users facing this CI environment limitation to apply the critical security fixes without forcing a costly and immediate upgrade of our entire CI infrastructure.

Thank you for your prompt attention to this critical security matter.

### Environment Details

  * **OpenNext Version:** v3.9.6 (Failing) / v3.7.6 (Working)
  * **Next.js Version:** v15.4.x
  * **CI Runner OS:** Linux (Architecture: `arm64`) - (Old GLIBC / Amazon Linux 2 Equivalent)

sradevski:
This is related to https://github.com/opennextjs/opennextjs-cloudflare/issues/1050 
It potentially fixes https://github.com/opennextjs/opennextjs-aws/issues/1067 as well, but I didn't test that particular image/setup.
closes #1067

### DIFF
--- a/.changeset/true-teeth-sin.md
+++ b/.changeset/true-teeth-sin.md
@@ -1,5 +1,0 @@
----
-"@opennextjs/aws": patch
----
-
-Bump ast-grep/napi in order to fix an issue with certain linux distros


.changeset/true-teeth-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Bump ast-grep/napi in order to fix an issue with certain linux distros
```
vicb: Would this be accurate?
```suggestion
Bump ast-grep/napi in order to fix an issue with older linux distros
```

vicb (approved):
I have removed the changeset because IMO sometimes no info is better than vague info

.changeset/true-teeth-sin.md
vicb: Ended removing the changeset

sradevski:
Hey @vicb no worries, thanks for looking into it. I didn't realize `opennextjs-cloudflare` has ast-grep as a dependency as well, here's the PR: https://github.com/opennextjs/opennextjs-cloudflare/pull/1112

All good on the changeset, I thought it's needed to trigger a new release, that's all :)

PR #1104, opened 2026-02-05
fix: ensure copied files are writable for patching

anonrig:
copyFileSync preserves source file permissions. When source files are read-only (e.g., in Bazel's node_modules managed by aspect_rules_js), the copied files remain read-only. This causes EACCES errors when codePatcher tries to write patches to these files.

This fix adds a helper function that copies files and ensures they have write permissions, preventing permission denied errors in build environments with read-only source files.

Related to https://github.com/cloudflare/workerd/pull/5993

### DIFF
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -1,6 +1,7 @@
 import url from "node:url";
 
 import {
+  chmodSync,
   copyFileSync,
   cpSync,
   existsSync,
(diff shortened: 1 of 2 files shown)


packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -32,6 +33,21 @@ import {
 
 const __dirname = url.fileURLToPath(new URL(".", import.meta.url));
 
+/**
+ * Copies a file and ensures the destination is writable.
+ * This is necessary because copyFileSync preserves file permissions,
+ * and source files may be read-only (e.g., in Bazel's node_modules).
+ * Without this, subsequent patches would fail with EACCES errors.
+ */
+function copyFileAndMakeWritable(src: string, dest: string): void {
```
vicb:
```suggestion
function copyFileAndMakeOwnerWritable(src: string, dest: string): void {
```

vicb:
Thanks @anonrig

PR #1107, opened 2026-02-06
Write segment data on set

conico974:
Fix for https://github.com/opennextjs/opennextjs-cloudflare/issues/1115

### DIFF
--- a/packages/tests-unit/tests/adapters/cache.test.ts
+++ b/packages/tests-unit/tests/adapters/cache.test.ts
@@ -331,8 +331,8 @@ describe("CacheHandler", () => {
             html: "<html></html>",
             rsc: "rsc-data",
             segmentData: {
-              "segment1": "data1",
-              "segment2": "data2",
+              segment1: "data1",
+              segment2: "data2",
             },
             meta: {
               status: 200,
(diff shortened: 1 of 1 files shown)


conico974:
I haven't added test for it, not sure where or how yet. And I'm still AFK today 😉. But otherwise it works

vicb:
> And I'm still AFK today

Same here, let's look next week when we are closer to a keyboard ;)

packages/tests-unit/tests/adapters/cache.test.ts
```diff
@@ -323,6 +323,46 @@ describe("CacheHandler", () => {
...
+              segment1: "data1",
+              segment2: "data2",
+            },
+            meta: {
+              status: 200,
+              headers: { "x-custom": "value" },
+              postponed: "postponed-data",
+            },
+          },
+          lastModified: Date.now(),
```
vicb, line 343: Should you mock the timer/use a constant?

khuezy (approved):
LGTM!

I just hope that Meta and Vercel properly sanitize `postponed` and that we don't see another CVE soon.

packages/tests-unit/tests/adapters/cache.test.ts
conico974: Already done in this file https://github.com/opennextjs/opennextjs-aws/blob/bec7b1bf493974152d212334fd5e4083ab0beb3b/packages/tests-unit/tests/adapters/cache.test.ts#L15

vicb: Sorry, missed I missed it 🚀

PR #1109, opened 2026-02-06
fix(pages-router): Add patch for trustHostHeader using res.revalidate

### LINKED ISSUE #1102: Error: Failed to revalidate /path: Invariant: missing internal router-server-methods this is an internal bug
There is an error with latest OpenNext using pages router's `res.revalidate("/path")` on latest Next 16. It will throw with an error: `Error: Failed to revalidate /test/1: Invariant: missing internal router-server-methods this is an internal bug`.

I wont have time to look at this before later, but I made the issue to track it. If anyone wants to dig into this feel free!

I made a minimal reproduction here: https://github.com/sommeeeer/opennext-pages-revalidate-invariant

Steps are in the README

sommeeeer:
Closes https://github.com/opennextjs/opennextjs-aws/issues/1102

`res.revalidate("/path")` is broken on Next 15 and 16 for pages router. The reason is that [this](https://github.com/sommeeeer/next.js/blob/ea08bf273ce03ac2ba8c6b021f35c27a8facdac1/packages/next/src/server/api-utils/node/api-resolver.ts#L301) part of the revalidate code in Next will be undefined in our case. 

This code ends up bundled in `/next/dist/compiled/next-server/pages-api.runtime.prod.js`. I tried many other ways to solve this that **failed**:

- Setting [this](https://github.com/vercel/next.js/blob/7cd9c7071ea92d599b7415d0d1c195a17a2407f0/packages/next/src/build/templates/pages-api.ts#L97) env variable during build.
- I tried setting `experimental.trustHostHeader` in Next config.
- Also as a last try I tried passing `trustHostHeader` many places in `new NextServer.default({});`. Didn't seem to have any effect at all.

Not sure why none of these workarounds did the trick, but atleast patching the bundled code in the runtime module seems to do the trick.

### DIFF
--- a/examples/pages-router/src/pages/revalidate/[key].tsx
+++ b/examples/pages-router/src/pages/revalidate/[key].tsx
@@ -1,0 +1,69 @@
+import type { GetStaticPropsContext, GetStaticPropsResult } from "next";
+import Head from "next/head";
+
+type Params = { key: string };
+
+type FakeRecord = {
+  key: string;
+  title: string;
+  updatedAt: string;
+};
+
+type Props = {
+  record: FakeRecord;
+};
+
+const fakeDb: FakeRecord[] = [
+  { key: "1", title: "First record", updatedAt: new Date().toISOString() },
+  { key: "2", title: "Second record", updatedAt: new Date().toISOString() },
+  { key: "3", title: "Third record", updatedAt: new Date().toISOString() },
+];
+
+export default function TestKeyPage({ record }: Props) {
+  return (
+    <div style={{ padding: 24 }}>
+      <Head>
+        <title>SSG Test — {record.key}</title>
+      </Head>
+      <h1>SSG Test Page</h1>
+      <p>
+        <strong>Key:</strong> {record.key}
+      </p>
+      <p>
+        <strong>Title:</strong> {record.title}
+      </p>
+      <p>
+        <strong>Updated:</strong>{" "}
+        <span data-testid="updated-at">{record.updatedAt}</span>
+      </p>
+      <p>Revalidate set in getStaticProps.</p>
+    </div>
+  );
+}
+
+export async function getStaticProps({
+  params,
+}: GetStaticPropsContext<Params>): Promise<GetStaticPropsResult<Props>> {
+  const found = fakeDb.find((item) => item.key === params?.key);
+
+  if (!found) {
+    return { notFound: true };
+  }
+
+  return {
+    props: {
+      record: {
+        ...found,
+        updatedAt: new Date().toISOString(),
+      },
+    },
+  };
+}
+
+export async function getStaticPaths() {
+  const paths = fakeDb.map((item) => ({
+    params: { key: item.key },
+  }));
+
+  return { paths, fallback: "blocking" };
+}
--- a/examples/pages-router/src/pages/revalidate/[key].tsx
+++ b/examples/pages-router/src/pages/revalidate/[key].tsx
@@ -13,10 +13,19 @@ type Props = {
   record: FakeRecord;
 };
 
-const fakeDb: FakeRecord[] = [
-  { key: "1", title: "First record", updatedAt: new Date().toISOString() },
-  { key: "2", title: "Second record", updatedAt: new Date().toISOString() },
-  { key: "3", title: "Third record", updatedAt: new Date().toISOString() },
+const fakeDb: Omit<FakeRecord, "updatedAt">[] = [
+  {
+    key: "1",
+    title: "First record",
+  },
+  {
+    key: "2",
+    title: "Second record",
+  },
+  {
+    key: "3",
+    title: "Third record",
+  },
 ];
 
 export default function TestKeyPage({ record }: Props) {
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -229,6 +229,7 @@ export default class Cache {
     if (globalThis.openNextConfig.dangerous?.disableIncrementalCache) {
       return;
     }
+    console.log("SET CACHE", { key, data, ctx });
     // This one might not even be necessary anymore
     // Better be safe than sorry
     const detachedPromise = globalThis.__openNextAls
@@ -348,6 +349,7 @@ export default class Cache {
       await this.updateTagsOnSet(key, data, ctx);
       debug("Finished setting cache");
     } catch (e) {
+      console.log("ARE WE HERE");
       error("Failed to set cache", e);
     } finally {
       // We need to resolve the promise even if there was an error
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -349,7 +349,6 @@ export default class Cache {
       await this.updateTagsOnSet(key, data, ctx);
       debug("Finished setting cache");
     } catch (e) {
-      console.log("ARE WE HERE");
       error("Failed to set cache", e);
     } finally {
       // We need to resolve the promise even if there was an error
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -229,7 +229,6 @@ export default class Cache {
     if (globalThis.openNextConfig.dangerous?.disableIncrementalCache) {
       return;
     }
-    console.log("SET CACHE", { key, data, ctx });
     // This one might not even be necessary anymore
     // Better be safe than sorry
     const detachedPromise = globalThis.__openNextAls
--- a/packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
+++ b/packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
@@ -33,7 +33,7 @@ fix:
   'true'
 `;
 
-// Use correct protocol from `NextInternalRequestMeta` when doing HEAD fetch for revalidation
+// Use correct protocol when doing HEAD fetch for revalidation
 export const headFetchProtocolRule = `
 rule:
   kind: string_fragment
@@ -80,7 +80,7 @@ export const patchPagesApiRuntimeProd: CodePatcher = {
       patchCode: createPatchCode(trustHostHeaderRule),
       versions: ">=15.0.0",
     },
-    // Use correct protocol from `NextInternalRequestMeta` when doing HEAD fetch for revalidation
+    // Use correct protocol when doing HEAD fetch for revalidation
     {
       pathFilter,
       contentFilter: /https/,
--- a/packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
+++ b/packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
@@ -1,0 +1,91 @@
+import { getCrossPlatformPathRegex } from "utils/regex.js";
+import { createPatchCode } from "../astCodePatcher.js";
+import type { CodePatcher } from "../codePatcher.js";
+
+// `context.trustHostHeader` is undefined in our case
+// Trust the host header when invoking `res.revalidate("/path")` from pages router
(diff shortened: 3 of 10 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -314,6 +315,7 @@ export default class Cache {
       await this.updateTagsOnSet(key, data, ctx);
       debug("Finished setting cache");
     } catch (e) {
+      console.log("ARE WE HERE");
```
khuezy: dev logs?

sommeeeer: Yup, nice catch :) thanks

sommeeeer: Seems that while making this I found another bug. You can trigger it like this: 
```bash
cd examples/pages-router
pnpm openbuild:local && pnpm openbuild:local:start
```

Then in your browser open http://localhost:3002/revalidate/123 and see the error in the terminal.

sommeeeer: image

khuezy (commented):
LGTM but I'll defer to Conico/Victor since I haven't dealt w/ page router in years.

sommeeeer:
Thanks for the review @khuezy

examples/pages-router/src/pages/revalidate/[key].tsx
```diff
@@ -0,0 +1,69 @@
...
+};
+
+type Props = {
+  record: FakeRecord;
+};
+
+const fakeDb: FakeRecord[] = [
+  { key: "1", title: "First record", updatedAt: new Date().toISOString() },
+  { key: "2", title: "Second record", updatedAt: new Date().toISOString() },
+  { key: "3", title: "Third record", updatedAt: new Date().toISOString() },
```
Comment (conico974) [inline examples/pages-router/src/pages/revalidate/[key].tsx]: This updatedAt property here is a bit useless TBH

examples/pages-router/src/pages/revalidate/[key].tsx
```diff
@@ -0,0 +1,69 @@
...
+    },
+  };
+}
+
+export async function getStaticPaths() {
+  const paths = fakeDb.map((item) => ({
+    params: { key: item.key },
+  }));
+
+  return { paths, fallback: "blocking" };
```
Comment (conico974) [inline examples/pages-router/src/pages/revalidate/[key].tsx:77]: Is it limited to fallback blocking?

packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export const patchPagesApiRuntimeProd: CodePatcher = {
+  name: "patch-pages-api-runtime-prod",
+  patches: [
+    // Trust the host header when invoking `res.revalidate("") from pages router
+    {
+      pathFilter,
+      contentFilter: /trustHostHeader/,
+      patchCode: createPatchCode(trustHostHeaderRule),
+      versions: ">=15.0.0",
```
conico974, line 81: Is it right? I'm pretty sure that it worked in 15+

packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
```diff
@@ -0,0 +1,91 @@
...
+    },
+    // Use correct protocol when doing HEAD fetch for revalidation
+    {
+      pathFilter,
+      contentFilter: /https/,
+      patchCode: createPatchCode(headFetchProtocolRule),
+      versions: ">=15.0.0",
```
conico974, line 88: Same question?

packages/tests-e2e/tests/pagesRouter/revalidate.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  const json = await res.json();
+
+  expect(json).toEqual({ revalidated: true, path: "/revalidate/1" });
+
+  if (!res.ok()) {
+    throw new Error(`Failed to trigger revalidation: ${await res.text()}`);
+  }
+
+  // Wait for a short period to allow revalidation to complete
+  await page.waitForTimeout(1000);
```
conico974: Isn't `res.revalidate` done once the response is returned?

examples/pages-router/src/pages/revalidate/[key].tsx
```diff
@@ -0,0 +1,69 @@
...
+  record: FakeRecord;
+};
+
+const fakeDb: FakeRecord[] = [
+  { key: "1", title: "First record", updatedAt: new Date().toISOString() },
+  { key: "2", title: "Second record", updatedAt: new Date().toISOString() },
+  { key: "3", title: "Third record", updatedAt: new Date().toISOString() },
+];
+
+export default function TestKeyPage({ record }: Props) {
```
Comment (conico974) [inline examples/pages-router/src/pages/revalidate/[key].tsx:31]: Won't this page get cached in cloudfront?
We need to modify the cache control header in the middleware to ensure this won't happen

packages/tests-e2e/tests/pagesRouter/revalidate.test.ts
```diff
@@ -0,0 +1,25 @@
+import { expect, test } from "@playwright/test";
+
+test("res.revalidate should work", async ({ page }) => {
+  await page.goto("/revalidate/1");
+
+  const initialUpdatedAt = await page.getByTestId("updated-at").textContent();
```
conico974, line 6: We should also ensure that the page is indeed SSG, we cannot be sure of that in this test, and it will succeed improperly

examples/pages-router/src/pages/revalidate/[key].tsx
Comment (sommeeeer) [reply examples/pages-router/src/pages/revalidate/[key].tsx:77]: Nope, it doesn't matter what you set `fallback` to. The error will get triggered on all 3 possible values.

packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
sommeeeer: Yup. I thought so too, but I have double checked, and `>=15` is correct.

packages/tests-e2e/tests/pagesRouter/revalidate.test.ts
sommeeeer: Yeah, your right about that one. I'll remove it.

examples/pages-router/src/pages/revalidate/[key].tsx
Comment (sommeeeer) [reply examples/pages-router/src/pages/revalidate/[key].tsx:31]: > Won't this page get cached in cloudfront?
> We need to modify the cache control header in the middleware to ensure this won't happen

Good catch. Yeah we need to revalidate it in Cloudfront aswell. As a workaround I will modify it in the middleware.

sommeeeer:
Thanks for the review Nico! I have addressed your comments, and this PR should be ready to ship now.

When it comes to [this](https://github.com/opennextjs/opennextjs-aws/pull/1109#discussion_r2776452011) error its coming from [here](https://github.com/sommeeeer/opennextjs-aws/blob/43d31b798791de11f844eedec6c2e7943e92e2bb/packages/open-next/src/overrides/incrementalCache/fs-dev.ts#L33). It won't find that `path` since it doesn't really exist. Thus resulting in the error message from the image. It can probably be ignored? [This](https://github.com/sommeeeer/opennextjs-aws/blob/43d31b798791de11f844eedec6c2e7943e92e2bb/packages/open-next/src/adapters/cache.ts#L216-L218) is whats calling it. `data` is null here if you try to access a path that was not generated at build time.

packages/tests-unit/tests/build/patch/patches/patchPagesApiRuntimeProd.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          pagesResRevalidateCodeUnbundled,
+          headFetchProtocolRule,
+        ),
+      ).toMatchInlineSnapshot(`
+        "Index: pages-api.runtime.prod.js
+        ===================================================================
+        --- pages-api.runtime.prod.js
+        +++ pages-api.runtime.prod.js
+        @@ -1,5 +1,4 @@
+        -
```
vicb, line 178: what's the diff here?

packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
```diff
@@ -0,0 +1,91 @@
...
+import { createPatchCode } from "../astCodePatcher.js";
+import type { CodePatcher } from "../codePatcher.js";
+
+// `context.trustHostHeader` is undefined in our case
+// Trust the host header when invoking `res.revalidate("/path")` from pages router
+// https://github.com/vercel/next.js/blob/178a4c7/packages/next/src/server/api-utils/node/api-resolver.ts#L301
+export const trustHostHeaderRule = `
+rule:
+  kind: member_expression
+  pattern: $CONTEXT.trustHostHeader
```
vicb, line 11: Looks like there are 3 occurences of `context.tustHostHeader` in this function, what is the rationale for only updating one?

packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
```diff
@@ -0,0 +1,91 @@
...
+  inside:
+    kind: parenthesized_expression
+    inside:
+      kind: if_statement
```
vicb, line 15: We should look at making this simpler.
If not possible, it would be nice to add what code we are looking for in the JSDoc

packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
```diff
@@ -0,0 +1,91 @@
...
+                    kind: call_expression
+                    has:
+                      kind: identifier
+                      regex: ^fetch$
+fix:
+  'true'
+`;
+
+// Use correct protocol when doing HEAD fetch for revalidation
+export const headFetchProtocolRule = `
```
vicb, line 37: Same comment here, could you please add what snippet we are tryin to match. We might be able to come with a more readable rule too.

packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
```diff
@@ -0,0 +1,91 @@
...
+  String.raw`/next/dist/compiled/next-server/pages-api(-turbo)?\.runtime\.prod\.js$`,
+  {
+    escape: false,
+  },
+);
+
+export const patchPagesApiRuntimeProd: CodePatcher = {
+  name: "patch-pages-api-runtime-prod",
+  patches: [
+    // Trust the host header when invoking `res.revalidate("") from pages router
```
vicb, line 76: I think it's ok to only comment on `trustHostHeaderRule` ?
```suggestion
```

packages/open-next/src/build/patch/patches/patchPagesApiRuntimeProd.ts
```diff
@@ -0,0 +1,91 @@
...
+    {
+      pathFilter,
+      contentFilter: /trustHostHeader/,
+      patchCode: createPatchCode(trustHostHeaderRule),
+      versions: ">=15.0.0",
+    },
+    // Use correct protocol when doing HEAD fetch for revalidation
```
vicb, line 83: ditto
```suggestion
```

vicb (commented):
I think it used to work, trying to understand why not any more

it is no more [experimental](https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/core/util.ts#L48-L51)?

Oh maybe I was only considering cloudflare and we have [this](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/build/patches/plugins/res-revalidate.ts#L19-L20) and [this](https://github.com/opennextjs/opennextjs-cloudflare/blob/3acd272c97e2a405085ce0ef4e5a8daec32a529d/packages/cloudflare/src/cli/build/patches/plugins/route-module.ts#L67-L68)

Maybe we should dig a little and check if there is a way to factor code for both aws and cloudflare?

WDYT Magnus, you have look at it more deeply

PR #1114, opened 2026-02-09
Fix `findNextConfig` returning the incorrect result

dario-piotrowicz:
Sorry, I'm not sure how but the incorrect implementation for `findNextConfig` slipped through the cracks 🙇 

To be extra sure that the implementation is correct this time I've added unit tests for the utility 🙇

### DIFF
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -479,16 +479,16 @@ export function findNextConfig({
   | { path: string; isTypescript: boolean }
   | undefined {
   const extensions = [
-    { ext: ".ts", isTypescript: true },
-    { ext: ".mts", isTypescript: true },
-    { ext: ".cts", isTypescript: true },
-    { ext: ".js", isTypescript: false },
-    { ext: ".mjs", isTypescript: false },
-    { ext: ".cjs", isTypescript: false },
+    { ext: "ts", isTypescript: true },
+    { ext: "mts", isTypescript: true },
+    { ext: "cts", isTypescript: true },
+    { ext: "js", isTypescript: false },
+    { ext: "mjs", isTypescript: false },
+    { ext: "cjs", isTypescript: false },
   ];
 
   for (const { ext, isTypescript } of extensions) {
-    const configPath = path.join(appPath, `next.config${ext}`);
+    const configPath = path.join(appPath, `next.config.${ext}`);
     if (fs.existsSync(configPath)) {
       return {
         path: configPath,


khuezy (approved):
Good catch and no worries.

The original implementation was just returning the `.ext` huh... oooof.

I'll try to be more active in the PRs to help catch things 👀 

EDIT: it looks like it wasn't a big deal since the place where it checked didn't make use of the path.

packages/open-next/src/build/helper.ts
```diff
@@ -476,9 +476,15 @@ export function copyEnvFile(
 export function findNextConfig({
   appPath,
 }: Pick<BuildOptions, "appPath">): string | undefined {
-  return ["js", "cjs", "mjs", "ts"].find((ext) =>
+  const extension = ["js", "cjs", "mjs", "ts"].find((ext) =>
```
khuezy: Only in javascript, where you can have these many extensions. 😮‍💨 🤕

vicb: > Only in javascript, where you can have these many extensions. 😮‍💨 🤕

No, no... in TS too :)
This list is missing `mts` and `cts`

vicb (changes requested):
Good catch Dario.

As I commented inline, the function is also missing a couple of possible extensions.

It turns out we have an implementation of the same at https://github.com/opennextjs/opennextjs-aws/blob/cfbc4fbef5ee5d41497cda9cf7f12085e4789cd5/packages/open-next/src/build/patch/patches/patchOriginalNextConfig.ts#L79-L92

Could you please update the implementation here to return `{path, isTypescript}` and use the helper in `importNextConfigFromSource` too?

Thanks!

dario-piotrowicz:
> The original implementation was just returning the `.ext` huh... oooof.

Yeah... such a silly mistake! 🤦 

> EDIT: it looks like it wasn't a big deal since the place where it checked didn't make use of the path.

It wasn't for this package, but we use the function here: https://github.com/opennextjs/opennextjs-cloudflare/blob/3acd272c97e2a405085ce0ef4e5a8daec32a529d/packages/cloudflare/src/cli/commands/migrate.ts#L132 and here the path is indeed needed 😅

packages/open-next/src/build/helper.ts
```diff
@@ -471,14 +471,31 @@ export function copyEnvFile(
...
+  | { path: string; isTypescript: boolean }
+  | undefined {
+  const tsExtensions = [".ts", ".mts", ".cts"].map((ext) => ({
+    ext,
+    isTypescript: true,
+  }));
+  const jsExtensions = [".mjs", ".js", ".cjs"].map((ext) => ({
+    ext,
+    isTypescript: false,
+  }));
```
vicb: I guess at this point an explicit list would be more readable

```ts
const extensions: [ {ext: ".ts", isTypescript; true}, ...];
```

I suggested that to Nico on his inital PR ;)

packages/open-next/src/build/helper.ts
```diff
@@ -471,14 +471,31 @@ export function copyEnvFile(
...
+
+  for (const { ext, isTypescript } of [...tsExtensions, ...jsExtensions]) {
+    const testPath = path.join(appPath, `next.config${ext}`);
```
vicb:
```suggestion
    const configPath = path.join(appPath, `next.config${ext}`);
```

packages/open-next/src/build/helper.ts
```diff
@@ -471,14 +471,31 @@ export function copyEnvFile(
...
+    { ext: ".ts", isTypescript: true },
+    { ext: ".mts", isTypescript: true },
+    { ext: ".cts", isTypescript: true },
+    { ext: ".js", isTypescript: false },
+    { ext: ".mjs", isTypescript: false },
+    { ext: ".cjs", isTypescript: false },
+  ];
+
+  for (const { ext, isTypescript } of extensions) {
+    const configPath = path.join(appPath, `next.config${ext}`);
```
vicb: super nit, up to you to
```suggestion
    { ext: "ts", isTypescript: true },
    { ext: "mts", isTypescript: true },
    { ext: "cts", isTypescript: true },
    { ext: "js", isTypescript: false },
    { ext: "mjs", isTypescript: false },
    { ext: "cjs", isTypescript: false },
  ];

  for (const { ext, isTypescript } of extensions) {
    const configPath = path.join(appPath, `next.config.${ext}`);
```

PR #1116, opened 2026-02-10
feat(config): add `esbuild.external` option to `DefaultFunctionOptions`

### LINKED ISSUE #1115: Allow configuring custom esbuild externals for server function bundling
### Problem

When a Next.js plugin (e.g. `next-intl`) has a transitive dependency on a package with native `.node` binaries (e.g. `@swc/core`), the server function esbuild step fails because there is no loader configured for `.node` files:

```
ERROR: No loader is configured for ".node" files
  node_modules/.pnpm/@swc+core-linux-x64-musl@1.15.11/node_modules/@swc/core-linux-x64-musl/swc.linux-x64-musl.node
```

The esbuild externals in `helper.js` are currently hardcoded:

```js
external: [
  ...(esbuildOptions.external ?? []),
  "next",
  "./open-next.config.mjs",
],
```

There is no way to add custom entries to this list through configuration.

### What doesn't work

- **`serverExternalPackages`** in `next.config.ts`: Only affects Next.js webpack bundling, not OpenNext's esbuild step.
- **`install.packages`** in `open-next.config.ts`: Runs `installDependencies()` *after* esbuild, so the build still fails.

### Current workaround

Patching `@opennextjs/aws` (via `pnpm patch`) to add `@swc/core` and `@swc/wasm` to the esbuild externals list.

### Proposal

Allow users to specify additional esbuild externals through `open-next.config.ts`, for example:

```ts
// open-next.config.ts
const config = {
  default: {
    override: {
      // ...
    },
    esbuild: {
      external: ["@swc/core", "@swc/wasm"],
    },
  },
};

export default config;
```

### Environment

- `@opennextjs/aws`: 3.9.14
- `next`: 16.1.6
- `next-intl`: 4.8.2 (depends on `@swc/core`)
- Deployed via SST v3

clichedmoog:
## Summary

Closes #1115

When a Next.js plugin (e.g. `next-intl`) has a transitive dependency on packages with native `.node` binaries (e.g. `@swc/core`), the server function esbuild step fails with "No loader is configured for .node files".

The esbuild externals are currently hardcoded with no way to add custom entries via config.

This PR adds an `esbuild.external` option to `DefaultFunctionOptions`, allowing users to specify additional esbuild externals through `open-next.config.ts`.

### Changes
- Add `esbuild.external` option to `DefaultFunctionOptions` interface (`packages/open-next/src/types/open-next.ts`)
- Spread user-configured externals into esbuild calls across all bundle creation functions:
  - `createServerBundle.ts` — server function
  - `createImageOptimizationBundle.ts` — image optimization function (both async and sync passes)
  - `createRevalidationBundle.ts` — revalidation function
  - `createWarmerBundle.ts` — warmer function

### Usage

```ts
// open-next.config.ts
const config = {
  default: {
    esbuild: {
      external: ["@swc/core", "@swc/wasm"],
    },
  },
};
export default config;
```

Per-function configuration is also supported:

```ts
const config = {
  default: { /* ... */ },
  functions: {
    api: {
      routes: ["app/api/route"],
      patterns: ["/api/*"],
      esbuild: {
        external: ["@swc/core"],
      },
    },
  },
};
```

### DIFF
--- a/.changeset/custom-esbuild-externals.md
+++ b/.changeset/custom-esbuild-externals.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": minor
+---
+
+Add `esbuild.external` option to function config for custom esbuild externals
--- a/packages/open-next/src/build/createImageOptimizationBundle.ts
+++ b/packages/open-next/src/build/createImageOptimizationBundle.ts
@@ -66,7 +66,11 @@ export async function createImageOptimizationBundle(
           "image-optimization-adapter.js",
         ),
       ],
-      external: ["sharp", "next"],
+      external: [
+        "sharp",
+        "next",
+        ...(config.imageOptimization?.esbuild?.external ?? []),
+      ],
       outfile: path.join(outputPath, "index.mjs"),
       plugins,
     },
@@ -81,7 +85,11 @@ export async function createImageOptimizationBundle(
   buildHelper.esbuildSync(
     {
       entryPoints: [path.join(outputPath, "index.mjs")],
-      external: ["sharp", "@opentelemetry/api"],
+      external: [
+        "sharp",
+        "@opentelemetry/api",
+        ...(config.imageOptimization?.esbuild?.external ?? []),
+      ],
       allowOverwrite: true,
       outfile: path.join(outputPath, "index.mjs"),
       banner: {
--- a/packages/open-next/src/build/createRevalidationBundle.ts
+++ b/packages/open-next/src/build/createRevalidationBundle.ts
@@ -23,7 +23,12 @@ export async function createRevalidationBundle(
   // Build Lambda code
   await buildHelper.esbuildAsync(
     {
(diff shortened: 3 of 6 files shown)


clichedmoog:
@conico974 
You're right — esbuild here is only meant to bundle OpenNext's own adapter code.                                                                       

After investigating, the root cause is on `next-intl`'s side. Starting from v4.5.0, `next-intl` lists `@swc/core` as a regular dependency rather than a devDependency. However, `@swc/core` is only used at build time for the useExtracted feature (AST transformation via SWC plugin). The compiled output uses standard useTranslations and doesn't need @swc/core at runtime.

Because it's declared as a regular dependency, it ends up in the traced files and gets picked up by esbuild's module resolution during the server adapter bundling step.

I'll raise this upstream with `next-intl` to have `@swc/core` moved to devDependencies or optionalDependencies. That should be the proper fix rather than exposing esbuild internals through config.

Issue #166, opened 2023-07-28
bundled node_modules bloat


khuezy:
This issue is a PSA and to track future improvements to minimize lambda size.

These are dev/build deps that gets bundled, we should remove these during bundling... but for now, if you want to reduce the lambda size, manually remove them prior to deployment.
```
    '@esbuild',
    '@swc/core-darwin-arm64', // 33MB!
    'esbuild',
    'webpack',
    'uglify-js',
    '@webassemblyjs'
```

There be more, but these are the ones that take up the majority of the bloat.

NOTE: after attempting to programmatically setting this, it was too brittle to correctly handle due to the config having potential functions, which is not possible to stringify/parse the user file. And since `next build` runs on a child thread, we can't set the next.config at runtime.

chorobin:
I've noticed this issue for a while. I tried to resolve it with `experimental.outputFileTracingExcludes` but I could never get it to work

khuezy:
I was able to get outputFileTracingExcludes to work, I'll put up a PR soon.

chorobin:
Nice. Maybe I had something wrong with the globs

khuezy:
Trying to update the user's `next.config.js` file to include the `outputFileTracingExcludes` is not trivial, especially when they have a function in the object - it's hard to stringify and parse the function :/

We can't import the config and update the object b/c `next build` is executed on a different thread...

I think we'll just document the config and let users configure what they want to exclude.

chorobin:
The extra docs would be useful for this case

sladg:
Same as `'@swc/core-darwin-arm64', // 33MB!` applies to Prisma. If I specify 
```
generator client {
    provider        = "prisma-client-js"
    binaryTargets   = ["native", "rhel-openssl-1.0.x"]
}
```
in `schema.prisma` I get both, native `libquery_engine-darwin-arm64.dylib.node` as well as Lambda's `libquery_engine-rhel-openssl-1.0.x.so.node`. 

OS-specified libraries are pain and Next does not have mechanism to omit them for target infra while bundling.

khuezy:
thanks @sladg I'll put up a PR next week to remove these at build time.

PR #1117, opened 2026-02-13
fix(build): auto-filter non-Linux platform-specific native binaries

clichedmoog:
## Motivation

When building on macOS, npm/pnpm installs platform-matching native binaries (e.g. `@swc/core-darwin-arm64`, `@esbuild/darwin-arm64`) via `optionalDependencies`. Next.js output file tracing picks these up because they exist on disk, causing two problems:

1. **esbuild build crash** — esbuild has no loader for `.node` files and fails with "No loader is configured for .node files" ([next-intl#2255](https://github.com/amannn/next-intl/issues/2255))
2. **Bundle bloat** — `copyTracedFiles` copies darwin/win32 binaries into the Lambda bundle where they can never work (~65MB in a production app)

The current user-side workaround is `outputFileTracingExcludes` in `next.config.ts`, but this requires manually listing every platform-specific package.

This mainly affects local macOS/Windows builds — CI environments on Linux wouldn't install these binaries in the first place.

**Local development is not affected** — these changes only run during the OpenNext build step. `next dev`, `next build`, and the local OpenNext server (`openbuild:local:start`) all work normally on macOS.

### 1. `helper.ts` — Prevent esbuild crash on `.node` files
Added `loader: { ".node": "empty" }` to both `esbuildSync` and `esbuildAsync`, so esbuild skips native binaries instead of crashing.

### 2. `copyTracedFiles.ts` — Filter non-Linux platform binaries
Added `isNonLinuxPlatformPackage()` that auto-detects and excludes `darwin`/`win32`/`freebsd`/`android` packages matching the `{pkg}-{platform}-{arch}` naming convention. Operates independently from existing `EXCLUDED_PACKAGES`.

Both `linux-arm64` and `linux-x64` packages are kept since the target architecture is unknown at build time.

This filter complements the existing `install` feature (`InstallOptions` with `--os`/`--arch`): `install` provides the correct platform binaries for runtime dependencies, while this filter removes the wrong ones that came from the host machine.

### Related issues
- #166 (bundled node_modules bloat)
- #1115 (esbuild externals)
- [next-intl#2255](https://github.com/amannn/next-intl/issues/2255) (esbuild .node crash)

## Verification
Tested on a production Next.js app (next-intl + Prisma + esbuild) by **completely removing `outputFileTracingExcludes`** — both Next.js and OpenNext builds succeed, and no darwin binaries end up in the Lambda bundle.

### DIFF
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -73,7 +73,7 @@ export function isExcluded(srcPath: string): boolean {
   );
 }
 
-const NON_LINUX_PLATFORMS = ["darwin", "win32", "freebsd", "android"];
+const NON_LINUX_PLATFORMS = ["darwin", "win32", "freebsd"];
 
 const platformPattern = NON_LINUX_PLATFORMS.join("|");
 const nonLinuxPlatformRegex = getCrossPlatformPathRegex(
@@ -297,7 +297,11 @@ File ${serverPath} does not exist
       return;
     }
     // Skip non-Linux platform-specific native binaries (e.g. @swc/core-darwin-arm64)
-    if (isNonLinuxPlatformPackage(from)) {
+    // Set OPEN_NEXT_SKIP_PLATFORM_FILTER=true to bypass this check
+    if (
+      !process.env.OPEN_NEXT_SKIP_PLATFORM_FILTER &&
+      isNonLinuxPlatformPackage(from)
+    ) {
       const match = from.match(
         /node_modules\/(?:\.pnpm\/.*\/node_modules\/)?((?:@[^/]+\/)?[^/]+)/,
       );
--- a/packages/open-next/src/build/helper.ts
+++ b/packages/open-next/src/build/helper.ts
@@ -124,8 +124,6 @@ export function esbuildSync(
     sourcemap: debug ? "inline" : false,
     sourcesContent: false,
     ...esbuildOptions,
-    // Native .node binaries cannot be bundled by esbuild
-    loader: { ".node": "empty", ...esbuildOptions.loader },
     external: ["./open-next.config.mjs", ...(esbuildOptions.external ?? [])],
     banner: {
       ...esbuildOptions.banner,
@@ -165,8 +163,6 @@ export async function esbuildAsync(
     sourcemap: debug ? "inline" : false,
     sourcesContent: false,
     ...esbuildOptions,
-    // Native .node binaries cannot be bundled by esbuild
-    loader: { ".node": "empty", ...esbuildOptions.loader },
     external: [
       ...(esbuildOptions.external ?? []),
       "next",
--- a/packages/tests-unit/tests/build/copyTracedFiles.test.ts
+++ b/packages/tests-unit/tests/build/copyTracedFiles.test.ts
@@ -66,15 +66,10 @@ describe("isNonLinuxPlatformPackage", () => {
     ).toBe(true);
   });
 
-  test("should exclude freebsd and android packages", () => {
+  test("should exclude freebsd packages", () => {
     expect(
       isNonLinuxPlatformPackage(
         "/project/node_modules/@rollup/rollup-freebsd-x64/rollup.freebsd-x64.node",
-      ),
-    ).toBe(true);
-    expect(
-      isNonLinuxPlatformPackage(
-        "/project/node_modules/@rollup/rollup-android-arm64/rollup.android-arm64.node",
       ),
     ).toBe(true);
   });


vicb:
@clichedmoog Thanks for the PR.

Could you please expand about the rationale for this PR?

Would I still be able to dev on my local Mac Book if we merge this?

clichedmoog:
@vicb Thanks for the review!

**Rationale:**

When building on macOS, npm/pnpm installs platform-matching native binaries (e.g. `@swc/core-darwin-arm64`, `@esbuild/darwin-arm64`) via `optionalDependencies`. Next.js output file tracing picks these up because they exist on disk, and `copyTracedFiles` copies them into the Lambda bundle
 — where they won't work due to the binary format mismatch with Linux.

I ran into this in a production project using `next-intl` (which depends on `@swc/core`). The workaround I ended up with was adding `outputFileTracingExcludes` to `next.config.ts` to manually list each platform-specific package. It works, but needs to be updated whenever a new native dependency is added.

In that project, I measured ~65MB of darwin-only binaries in a 209MB standalone output - `@swc/core-darwin-arm64` (22MB), `@esbuild/darwin-arm64`(10MB), Prisma darwin engine (18MB), etc.

This PR tries to automate that filtering by matching the `{pkg}-{platform}-{arch}` naming convention that most native packages follow.

I should note that this mainly affects local macOS builds — CI environments on Linux wouldn't install darwin binaries in the first place (same reason Vercel doesn't hit this issue).

**Local Mac development:**

Yes — this filter only runs inside `copyTracedFiles()`, which is part of the OpenNext build step for creating the Lambda bundle. `next dev` and `next build` are not affected since they don't go through OpenNext.

I tested on macOS and confirmed `pnpm dev` works normally, including `next-intl` i18n which uses `@swc/core` at dev time.

vicb:
> **Local Mac development:**
> 
> Yes — this filter only runs inside `copyTracedFiles()`, which is part of the OpenNext build step for creating the Lambda bundle. `next dev` and `next build` are not affected since they don't go through OpenNext.
> 
> I tested on macOS and confirmed `pnpm dev` works normally, including `next-intl` i18n which uses `@swc/core` at dev time.

I was referring the launching a local Open Next server, i.e. what we do for local e2e tests.

Maybe you can try if [this](https://github.com/opennextjs/opennextjs-aws/blob/main/README.md#testing) still works.

Edit: also I think we have a feature to install external packages and specify what arch to install - I don't have time to lookup for the details but you should be able to find something in the code or looking at the Open Next config.

clichedmoog:
@vicb Thanks for the clarification! I haven't tested the full local OpenNext e2e flow yet — I'll set up the examples and try `openbuild:local` + `openbuild:local:start` on macOS.

That said, I think it should be fine: the packages typically matched by this filter (SWC, esbuild, Rollup, etc.) are build-time native binaries that aren't needed by the server at runtime — Next.js pre-compiles everything in standalone mode. The `.node` empty loader similarly only affects the esbuild bundling step, not runtime loading.

Good point about the `install` feature — I found `installDeps.ts` with `--os` defaulting to `"linux"`. For packages that genuinely need native binaries at runtime (like Prisma engines or sharp), users would use `install` to get the correct platform binaries. So the two features seem complementary: `install` provides the right binaries, and this filter removes the wrong ones that came from the host machine.

I'll follow up with local e2e test results.

clichedmoog:
**Update:** Tested the local OpenNext e2e flow on macOS (M1):

1. `pnpm build` — all packages compiled
2. `pnpm -r openbuild:local` — all 4 examples (app-router, pages-router, app-pages-router, experimental) built successfully with OpenNext
3. `openbuild:local:start` — all servers started on ports 3001-3004
4. Verified SSR pages serve correctly via curl on all 4 servers

Local OpenNext development on macOS is not affected by this change.

vicb:
> Local OpenNext development on macOS is not affected by this change.

Which as you touch base in your last message tend to confirm that they are not used.
Would make them external be a better way to solve your initial issue?
_(Only asking the question here, I have not looked deeply into the issue)_

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -73,6 +73,18 @@ export function isExcluded(srcPath: string): boolean {
   );
 }
 
+const NON_LINUX_PLATFORMS = ["darwin", "win32", "freebsd", "android"];
```
conico974: Pretty sure you can drop android 😄

packages/open-next/src/build/helper.ts
```diff
@@ -124,6 +124,8 @@ export function esbuildSync(
     sourcemap: debug ? "inline" : false,
     sourcesContent: false,
     ...esbuildOptions,
+    // Native .node binaries cannot be bundled by esbuild
+    loader: { ".node": "empty", ...esbuildOptions.loader },
```
conico974: This is supposed to be called only for OpenNext code, if not then we have a way bigger issue.

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -278,11 +290,22 @@ File ${serverPath} does not exist
...
   //Actually copy the files
+  const skippedPlatformPackages = new Set<string>();
   filesToCopy.forEach((to, from) => {
     // We don't want to copy excluded packages (e.g. sharp)
     if (isExcluded(from)) {
       return;
     }
+    // Skip non-Linux platform-specific native binaries (e.g. @swc/core-darwin-arm64)
+    if (isNonLinuxPlatformPackage(from)) {
+      const match = from.match(
```
conico974, line 305: This is always a bit dangerous to skip things like that. Could you add a bypass for that (like an env variable to bypass this check)

conico974 (changes requested):
I don't think the esbuild thing is doing anything, or if it is we have a bigger problem that we cannot just dismiss like that.

The esbuild issue is more likely coming from here https://github.com/opennextjs/opennextjs-aws/blob/8bd68097426a300029c2788d3e79f3d7679e9c53/packages/open-next/src/build/patch/patches/patchOriginalNextConfig.ts#L91-L98, we don't need to bundle anything here, just build the config file.

I'd prefer this esbuild fix to be in another PR as well, these are 2 different issue

packages/open-next/src/build/copyTracedFiles.ts
clichedmoog: Good idea. Added `OPEN_NEXT_SKIP_PLATFORM_FILTER` env variable for that.

packages/open-next/src/build/helper.ts
clichedmoog: That makes sense — moved to #1120 so we can discuss it separately there.

packages/open-next/src/build/copyTracedFiles.ts
clichedmoog: I've dropped the android

PR #1118, opened 2026-02-13
fix: preserve query parameters during i18n redirects to a localized path

### LINKED ISSUE #1060: Query parameters are lost during locale redirection
## Description
I am deploying a Next.js (Pages Router) application with i18n enabled using OpenNext.  
When a user accesses the root path (`/`) with a browser language setting that triggers a redirect to a non-default locale (e.g., `/en`), any query parameters present in the original URL are lost during the redirection.  

This behavior differs from the local Next.js environment (`next dev` / `next start`), where query parameters are correctly preserved.

## Environment
OpenNext version: 3.9.4
Next.js version: 14.2.33

## To Reproduce
1. Initialize a Next.js project
    ```sh
    npx create-next-app@14 . 
    # Select "Pages Router"
    npm install next-i18next react-i18next i18next
    ```

2. Configure i18n
    ```javascript
    // next-i18next.config.mjs
    export const i18n = {
        defaultLocale: 'ja',
        locales: ['en', 'ja'],
    }
    ```

    ```javascript
    // next.config.mjs
    import { i18n } from './next-i18next.config.mjs';

    const nextConfig = {
        i18n,
        reactStrictMode: true,
    };

    export default nextConfig;
    ```

3. Deploy using SST
    ```sh
    npx sst@latest init
    npx sst deploy --stage production
    ```

4. Change your browser's preferred language to English.

5. Access the deployed root URL with query parameters: `https://example.com/?xxx=yyy`

## Expected behavior
The user should be redirected to the English locale path with query parameters preserved: `https://example.com/en?xxx=yyy`

## Actual behavior
The user is redirected to the English locale path, but query parameters are missing: `https://example.com/en`

tsurumeso:
## Summary
This PR fixes a bug where query parameters are lost during i18n redirects to a localized path.

## Proposed Changes
Reflecting [the official Next.js implementation logic](https://github.com/vercel/next.js/blob/51d97c8bdcb2ccdf30207eac1331665a1bc2756f/packages/next/src/shared/lib/i18n/get-locale-redirect.ts#L70-L117):
- Domain-based locale redirects remain **unchanged** (query parameters are not preserved).
- Updated path-based locale redirects to preserve query parameters.

## Related Issue
Fix #1060

### DIFF
--- a/.changeset/i18n_redirects_with_query_parameters.md
+++ b/.changeset/i18n_redirects_with_query_parameters.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Preserve query parameters during i18n redirects to a localized path


tsurumeso:
> Would you mind adding a changeset?

Sure, I'll add a changeset.

> There is another section above that also returns the Location object, does that also need the queryString applied?

[That section](https://github.com/tsurumeso/opennextjs-aws/blob/c6bccc91ed5b5818e1e77f8d42306edbfa50f1fd/packages/open-next/src/core/routing/i18n/index.ts#L156) handles domain-based locale redirects. Since [the official Next.js implementation](https://github.com/vercel/next.js/blob/51d97c8bdcb2ccdf30207eac1331665a1bc2756f/packages/next/src/shared/lib/i18n/get-locale-redirect.ts#L104) does not append query parameters in that scenario, it is not necessary for OpenNext either.

> Also, what about trailingSlash: true configurations? Does that need to be handled here?

I think that likely needs to be handled as well. I'll update the code.

PR #1120, opened 2026-02-19
fix(build): add empty loader for .node files in esbuild

clichedmoog:
## Motivation

When a project has native binary dependencies (e.g. `@swc/core`, Prisma), esbuild can encounter `.node` files during bundling and crash with:

```
No loader is configured for ".node" files
```

This was split from #1117 per review feedback.

The root cause is that esbuild resolves package imports (e.g. `@swc/core-darwin-arm64`) through `package.json` main fields, which can point to `.node` native binaries. Since esbuild has no built-in handler for this format, it fails.

## Changes

Added `loader: { ".node": "empty" }` to both `esbuildSync` and `esbuildAsync` in `helper.ts`, so esbuild produces an empty module instead of crashing.

### Related issues
- [next-intl#2255](https://github.com/amannn/next-intl/issues/2255)
- #1115
- #1117

### DIFF
--- a/.changeset/esbuild-node-loader.md
+++ b/.changeset/esbuild-node-loader.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix(build): add empty loader for .node files in esbuild to prevent build crashes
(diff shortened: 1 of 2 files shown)


conico974 (changes requested):
Not the right change as previously said in https://github.com/opennextjs/opennextjs-aws/pull/1117#pullrequestreview-3801966195

PR #1121, opened 2026-02-20
fix(open-next): allow more status codes for ISR revalidation

ntltd:
### Problem

The revalidation handler was checking `res.statusCode !== 200` to determine success. This had two issues:

- **Too strict**: valid ISR outcomes like redirects (`301`, `302`, `307`, `308`), Not Found (`404`), and Gone (`410`) were incorrectly treated as failures and pushed to `failedRecords`.
- **Too implicit**: 5xx errors were excluded by accident rather than by intent, making the behavior hard to understand and easy to break.

### Solution

Replace the `!== 200` check with an explicit `ALLOWED_STATUSES` allowlist:

```js
const ALLOWED_STATUS_CODES = [200, 301, 302, 307, 308, 404, 410]
```

A revalidation is now only considered successful if both conditions are met:
- the status code is in `ALLOWED_STATUS_CODES`
- the `x-nextjs-cache` header equals `REVALIDATED`

### Why these status codes?

- `200` — standard successful page response
- `301, 302, 307, 308` — redirects are valid ISR outcomes and should not be treated as failures
- `404` — intentional Not Found pages are a valid cached state in Next.js ISR
- `410` — Gone, used for pages that have been intentionally removed

Any other status code (notably 5xx errors) will now explicitly mark the record as failed and trigger the existing retry/error reporting logic.

### DIFF
--- a/packages/open-next/src/adapters/revalidate.ts
+++ b/packages/open-next/src/adapters/revalidate.ts
@@ -25,7 +25,7 @@ export interface RevalidateEvent {
   }[];
 }
 
-const ALLOWED_STATUS_CODES = [200, 301, 302, 307, 308, 404, 410]
+const ALLOWED_STATUS_CODES = [200, 307, 308, 404]
 
 const defaultHandler = async (event: RevalidateEvent) => {
   const failedRecords: RevalidateEvent["records"] = [];
--- a/.changeset/breezy-owls-stand.md
+++ b/.changeset/breezy-owls-stand.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Fix ISR revalidation to also accept valid Next.js non-200 status codes (307, 308 and 404)


packages/open-next/src/adapters/revalidate.ts
```diff
@@ -25,6 +25,8 @@ export interface RevalidateEvent {
   }[];
 }
 
+const ALLOWED_STATUS_CODES = [200, 301, 302, 307, 308, 404, 410]
```
conico974: Is Next using the 410 status code? Similar question for 301/302, from what I remember, Next only uses 307/308

ntltd: Using `getServerSideProps`, you are able to set a custom status code. For example, on our website, we return a `410` status when content is deactivated.
While it's true that Next.js internally relies only on `307` and `308` status codes, I have included `301` and `302` to ensure full support for all HTTP redirects.

conico974: Sorry forgot to respond here, page using `getServerSideProps` does not use ISR revalidation , useless to add anything for them.

ntltd: You're right. I’ve updated the logic to include `200`, `404`, and `307` / `308` status codes. I’ve actually been maintaining a local patch for about a year to handle edge cases where 'bad' `404`s or redirects weren't revalidating properly in my app.
I can remove the `410` and `301` / `302` codes if you prefer. My initial goal was to be as exhaustive as possible regarding "allowed" codes, but your points are valid.

conico974: Don't hesitate to ping me if I don't respond.
If you could drop the 410 and 301/302 that would be great. I don't want to include things that are not Next.js behavior here.

ntltd: Done @conico974! Thanks for feedback.

conico974:
Thanks for the PR @ntltd

PR #1122, opened 2026-03-15
Add support for swr with revalidateTag

conico974:
Add support for swr with revalidateTag.

It introduces a new methods in the tag cache (`hasBeenStale`, optional), both for the original and the next mode tag cache.

It also introduces a RequestCache utils that can be used to cache things scoped to a request (stored in our internal AsyncLocalStorage context)

### DIFF
--- a/.changeset/slick-crabs-hear.md
+++ b/.changeset/slick-crabs-hear.md
@@ -1,0 +1,9 @@
+---
+"@opennextjs/aws": minor
+---
+
+Add support for stale-while-revalidate in revalidateTag
+
+Introduces a new methods in the tag cache (hasBeenStale, optional), both for the original and the next mode tag cache, mandatory to make swr work.
+
+It also introduces a RequestCache utils that can be used to cache things scoped to a request (stored in the OpenNext internal AsyncLocalStorage context)
--- a/.changeset/slick-crabs-hear.md
+++ b/.changeset/slick-crabs-hear.md
@@ -7,3 +7,7 @@ Add support for stale-while-revalidate in revalidateTag
 Introduces a new methods in the tag cache (hasBeenStale, optional), both for the original and the next mode tag cache, mandatory to make swr work.
 
 It also introduces a RequestCache utils that can be used to cache things scoped to a request (stored in the OpenNext internal AsyncLocalStorage context)
+
+### BREAKING CHANGE
+
+`writeTags` for the tag cache signature has changed to `writeTags(tags: NextModeTagCacheWriteInput[]): Promise<void>;` for Next mode, and `writeTags(tags: OriginalTagCacheWriteInput[]): Promise<void>;` for the original mode.
--- a/.changeset/slick-crabs-hear.md
+++ b/.changeset/slick-crabs-hear.md
@@ -4,11 +4,13 @@
 
 Add support for SWR (stale-while-revalidate) in `revalidateTag`
 
-Introduces a new methods in the tag cache (hasBeenStale, optional), both for the original and the next mode tag cache, mandatory for SWR.
+Introduces a new optional method `isStale` in the tag cache for both the original and the next modes. The implementation is mandatory for SWR to work.
 
-It also introduces a RequestCache utils that can be used to cache things scoped to a request (stored in the OpenNext internal AsyncLocalStorage context)
+It also introduces a `RequestCache` utility that can be used to cache things scoped to a request (stored in the OpenNext internal AsyncLocalStorage context)
 
 ### BREAKING CHANGE
 
 `writeTags` for the tag cache signature has changed to `writeTags(tags: NextModeTagCacheWriteInput[]): Promise<void>` for Next mode, and `writeTags(tags: OriginalTagCacheWriteInput[]): Promise<void>` for the original mode.
 This is breaking only for custom tag cache implementations, if you are using the default one provided by OpenNext, you don't need to do anything.
+
+`globalThis.isNextAfter15` is no longer available in the cache.
--- a/.changeset/slick-crabs-hear.md
+++ b/.changeset/slick-crabs-hear.md
@@ -2,12 +2,13 @@
 "@opennextjs/aws": minor
 ---
 
-Add support for stale-while-revalidate in revalidateTag
+Add support for SWR (stale-while-revalidate) in `revalidateTag`
 
-Introduces a new methods in the tag cache (hasBeenStale, optional), both for the original and the next mode tag cache, mandatory to make swr work.
+Introduces a new methods in the tag cache (hasBeenStale, optional), both for the original and the next mode tag cache, mandatory for SWR.
 
 It also introduces a RequestCache utils that can be used to cache things scoped to a request (stored in the OpenNext internal AsyncLocalStorage context)
 
 ### BREAKING CHANGE
 
-`writeTags` for the tag cache signature has changed to `writeTags(tags: NextModeTagCacheWriteInput[]): Promise<void>;` for Next mode, and `writeTags(tags: OriginalTagCacheWriteInput[]): Promise<void>;` for the original mode.
+`writeTags` for the tag cache signature has changed to `writeTags(tags: NextModeTagCacheWriteInput[]): Promise<void>` for Next mode, and `writeTags(tags: OriginalTagCacheWriteInput[]): Promise<void>` for the original mode.
+This is breaking only for custom tag cache implementations, if you are using the default one provided by OpenNext, you don't need to do anything.
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -466,8 +466,6 @@ export default class Cache {
           }
         }
 
-        console.log("Tags to revalidate", toInsert);
-
         // Update all keys with the given tag with revalidatedAt set to now
         await writeTags(toInsert);
 
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -6,7 +6,7 @@ import type {
 import {
   getTagsFromValue,
   hasBeenRevalidated,
-  hasBeenStale,
+  isStale as isStale,
   writeTags,
 } from "utils/cache";
 import { isBinaryContentType } from "../utils/binary";
@@ -103,7 +103,7 @@ export default class Cache {
 
       const _isStale = cachedEntry.shouldBypassTagCache
         ? false
-        : await hasBeenStale(key, _tags, _lastModified);
+        : await isStale(key, _tags, _lastModified);
 
       return {
         lastModified: _isStale ? 1 : _lastModified,
@@ -136,7 +136,7 @@ export default class Cache {
 
       const _isStale = cachedEntry.shouldBypassTagCache
         ? false
-        : await hasBeenStale(key, tags, _lastModified);
+        : await isStale(key, tags, _lastModified);
 
       const store = globalThis.__openNextAls.getStore();
       if (store) {
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -333,7 +333,10 @@ export default class Cache {
     }
   }
 
-  public async revalidateTag(tags: string | string[]) {
+  public async revalidateTag(
+    tags: string | string[],
+    durations?: { expire?: number }
+  ) {
     const config = globalThis.openNextConfig.dangerous;
     if (config?.disableTagCache || config?.disableIncrementalCache) {
       return;
@@ -347,7 +350,25 @@ export default class Cache {
       if (globalThis.tagCache.mode === "nextMode") {
         const paths = (await globalThis.tagCache.getPathsByTags?.(_tags)) ?? [];
 
-        await writeTags(_tags);
+        const now = Date.now();
+        const tagsToWrite = _tags.map((tag) => {
+          if (durations) {
+            // Use provided durations
+            return {
+              tag,
+              stale: now,
+              expiry: durations.expire !== undefined ? now + durations.expire * 1000 : undefined,
+            };
+          } else {
+            // Default behavior: immediate expiration
+            return {
+              tag,
+              expiry: now,
+            };
+          }
+        });
+        
+        await writeTags(tagsToWrite);
         if (paths.length > 0) {
           // TODO: we should introduce a new method in cdnInvalidationHandler to invalidate paths by tags for cdn that supports it
           // It also means that we'll need to provide the tags used in every request to the wrapper or converter.
@@ -373,10 +394,24 @@ export default class Cache {
         // Find all keys with the given tag
         const paths = await globalThis.tagCache.getByTag(tag);
         debug("Items", paths);
-        const toInsert = paths.map((path) => ({
-          path,
-          tag,
-        }));
+        const now = Date.now();
+        const toInsert = paths.map((path) => {
+          const baseEntry = { path, tag };
+          if (durations) {
+            // Use provided durations
+            return {
+              ...baseEntry,
+              stale: now,
+              expiry: durations.expire !== undefined ? now + durations.expire * 1000 : undefined,
+            };
+          } else {
+            // Default behavior: immediate expiration
+            return {
+              ...baseEntry,
+              expiry: now,
+            };
+          }
+        });
 
         // If the tag is a soft tag, we should also revalidate the hard tags
         if (tag.startsWith(SOFT_TAG_PREFIX)) {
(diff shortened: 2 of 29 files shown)


.changeset/slick-crabs-hear.md
```diff
@@ -0,0 +1,13 @@
+---
+"@opennextjs/aws": minor
+---
+
+Add support for stale-while-revalidate in revalidateTag
```
vicb:
```suggestion
Add support for SWR (stale-while-revalidate) in `revalidateTag`
```

.changeset/slick-crabs-hear.md
```diff
@@ -0,0 +1,13 @@
...
+
+Introduces a new methods in the tag cache (hasBeenStale, optional), both for the original and the next mode tag cache, mandatory to make swr work.
```
vicb:
```suggestion
Introduces a new methods in the tag cache (hasBeenStale, optional), both for the original and the next mode tag cache, mandatory for SWR.
```

.changeset/slick-crabs-hear.md
```diff
@@ -0,0 +1,13 @@
...
+
+It also introduces a RequestCache utils that can be used to cache things scoped to a request (stored in the OpenNext internal AsyncLocalStorage context)
+
+### BREAKING CHANGE
+
+`writeTags` for the tag cache signature has changed to `writeTags(tags: NextModeTagCacheWriteInput[]): Promise<void>;` for Next mode, and `writeTags(tags: OriginalTagCacheWriteInput[]): Promise<void>;` for the original mode.
```
vicb:
```suggestion
`writeTags` for the tag cache signature has changed to `writeTags(tags: NextModeTagCacheWriteInput[]): Promise<void>` for Next mode, and `writeTags(tags: OriginalTagCacheWriteInput[]): Promise<void>` for the original mode.
```

vicb: Maybe we can clarify that it is a breaking change for people implementing their own cache but it doesn't affect ppl using one of the builtin cache?

packages/open-next/src/types/overrides.ts
```diff
@@ -156,21 +156,38 @@ Cons :
 - One page request (i.e. GET request) could require to check a lot of tags (And some of them multiple time when used with the fetch cache)
 - Almost impossible to do automatic cdn revalidation by itself
 */
+
+export type NextModeTagCacheWriteInput =
+  | string
+  | {
+      tag: string;
+      stale?: number;
+      expiry?: number;
```
vicb: Are `stale` and `expiry` duration in seconds as `stale` and `expire` in [Next.js](https://github.com/vercel/next.js/blob/be3ee02b83756b9b24c07464b94e5fb687e8b1c8/packages/next/src/server/lib/cache-handlers/types.ts#L6)?

We should probably add comments and maybe make naming consistent with Next. I like when the name as a unit suffix (i.e. `S` or `Ms`)

packages/open-next/src/utils/cache.ts
```diff
@@ -1,10 +1,31 @@
...
   WithLastModified,
 } from "types/overrides";
 import { debug } from "../adapters/logger";
 
+export async function hasBeenStale(
+  key: string,
+  tags: string[],
+  lastModified?: number,
+): Promise<boolean> {
+  if (!globalThis.isNextAfter16) {
```
vicb: nit: add a comment saying that SWR had been implemented in Next 16

packages/open-next/src/utils/requestCache.ts
```diff
@@ -0,0 +1,28 @@
...
+   * If no Map exists yet for that key, a new empty Map is created, stored, and returned.
+   * Repeated calls with the same key always return the **same** Map instance.
+   */
+  getOrCreate<K = unknown, V = unknown>(key: string): Map<K, V> {
+    let cache = this._caches.get(key);
+    if (!cache) {
+      cache = new Map<unknown, unknown>();
+      this._caches.set(key, cache);
+    }
+    return cache as Map<K, V>;
```
vicb: type nits:

```suggestion
    let cache = this._caches.get(key) as Map<K, V> | undefined;
    if (!cache) {
      cache = new Map<K, V>();
      this._caches.set(key, cache);
    }
    return cache;
```

I still have to review of this is used but do we need to cleanup at some point / use a `WeakMap` ?

packages/open-next/src/overrides/tagCache/fs-dev.ts
```diff
@@ -51,11 +78,26 @@ const tagCache: TagCache = {
...
+          tag: { S: string };
+          path: { S: string };
+          revalidatedAt: { N: string };
+          stale?: { N: string };
+          expiry?: { N: string };
+        } = {
+          tag: { S: buildKey(item.tag) },
+          path: { S: buildKey(item.path) },
+          revalidatedAt: { N: `${item.revalidatedAt ?? Date.now()}` },
+        };
```
vicb:
```suggestion
        const tagEntry: TagCacheMetaFile = {
          tag: { S: buildKey(item.tag) },
          path: { S: buildKey(item.path) },
          revalidatedAt: { N: `${item.revalidatedAt ?? Date.now()}` },
        };
```

packages/open-next/src/overrides/tagCache/fs-dev.ts
```diff
@@ -51,11 +78,26 @@ const tagCache: TagCache = {
...
+        if (item.stale !== undefined) {
+          tagEntry.stale = { N: `${item.stale}` };
+        }
+        if (item.expiry !== undefined) {
+          tagEntry.expiry = { N: `${item.expiry}` };
+        }
```
vicb: Maybe this can be inline in the object above?

```ts
...(item.stale !== undefined ? {stale: { N: `${item.stale}` }} : {})
```

Similarly for expiry

packages/open-next/src/types/global.ts
```diff
@@ -112,6 +115,13 @@ declare global {
    */
   var isNextAfter15: boolean;
 
+  /**
```
vicb: nit: maybe we should pass the Next version and check at runtime to avoid having to add more `isNextAfter...` in the future ?

packages/open-next/src/utils/cache.ts
```diff
@@ -1,10 +1,31 @@
 import type {
   CacheValue,
+  NextModeTagCacheWriteInput,
   OriginalTagCacheWriteInput,
   WithLastModified,
 } from "types/overrides";
 import { debug } from "../adapters/logger";
 
+export async function hasBeenStale(
```
vicb: nit: add JSDoc for this function?

packages/open-next/src/adapters/cache.ts
```diff
@@ -96,8 +101,12 @@ export default class Cache {
         }
       }
 
+      const _isStale = cachedEntry.shouldBypassTagCache
+        ? false
+        : await hasBeenStale(key, _tags, _lastModified);
```
vicb: Question: should the name rather be `isStale` ?

packages/open-next/src/build/patch/patches/patchNextServer.ts
```diff
@@ -52,6 +52,20 @@ fix:
   this.handleNextImageRequest = async (req, res, parsedUrl) => false
 `;
 
+export const provideInternalWaitUntil = `
+rule:
```
vicb, line 56: Nit: would it be more robust to just replace the body of `getInternalWaitUntil` ?

We do something similar for `createEmptyBodyRule`

vicb (commented):
@conico974 I have started to look at the PR and added a few minor comments.

I'll do a second pass early afternoon.

packages/open-next/src/types/cache.ts
```diff
@@ -99,6 +99,8 @@ export type TagCacheMetaFile = {
   tag: { S: string };
   path: { S: string };
   revalidatedAt: { N: string };
+  stale?: { N: string };
+  expiry?: { N: string };
```
vicb: nits: 
- add JSDoc
- should we use `expire` for consistency with `ComposableCacheEntry` below?

packages/open-next/src/utils/requestCache.ts
conico974: This is only used inside of AsyncLocalStorage, the one scoped to the request. We create a new RequestCache every time, so we should be fine

packages/open-next/src/adapters/cache.ts
conico974: Good point, I wanted to have something similar to `hasBeenRevalidated`, but I don't think it makes much sense here

packages/open-next/src/types/overrides.ts
```diff
@@ -156,21 +156,38 @@ Cons :
...
-  writeTags(tags: string[]): Promise<void>;
+  writeTags(tags: NextModeTagCacheWriteInput[]): Promise<void>;
   // Optional method to get paths by tags
   // It is used to automatically invalidate paths in the CDN
   getPathsByTags?: (tags: string[]) => Promise<string[]>;
+  /**
+   * Optional method to check if any tag has become stale (but not yet expired).
+   * When tags are stale, the cache entry is still returned but revalidate is set to -1
+   * to trigger background revalidation.
+   */
```
vicb, line 192: nits:
- "check if any tag has become stale (but not yet expired)" -> this does not fully describes what happen when an entry is fresh, stale, or expired. Maybe we should list those 3 cases and the corresponding return value.
- Maybe doc the behavior when an entry is expired (dynamically regenerated).

packages/open-next/src/types/overrides.ts
```diff
@@ -198,6 +215,12 @@ export type OriginalTagCache = BaseTagCache & {
   getByPath(path: string): Promise<string[]>;
   getLastModified(path: string, lastModified?: number): Promise<number>;
   writeTags(tags: OriginalTagCacheWriteInput[]): Promise<void>;
+  /**
```
vicb, line 235: Same comments as above to clarify the behavior.

packages/open-next/src/utils/cache.ts
```diff
@@ -50,18 +71,25 @@ export function getTagsFromValue(value?: CacheValue<"cache">) {
...
+): string {
   if (typeof tag === "string") {
     return tag;
   }
-  return JSON.stringify({
-    tag: tag.tag,
-    path: tag.path,
-  });
+  // For OriginalTagCacheWriteInput, include path in the key
+  if ("path" in tag) {
```
vicb, line 100: question: Could we use `globalThis.tagCache.mode === "nextMode"` here for consistency?

packages/open-next/src/adapters/cache.ts
```diff
@@ -96,8 +101,12 @@ export default class Cache {
         }
       }
 
+      const _isStale = cachedEntry.shouldBypassTagCache
```
vicb, line 105: We have `if (_hasBeenRevalidated) return null;` above.

Does that means that `hasBeenRevalidated` would only return true with the max profile?

Maybe we could add a comment (and/or rename the function) to clarify?

packages/open-next/src/adapters/cache.ts
```diff
@@ -347,7 +364,27 @@ export default class Cache {
...
+            return {
+              tag,
+              stale: now,
+              expiry:
+                durations.expire !== undefined
+                  ? now + durations.expire * 1000
+                  : undefined,
+            };
+          }
+          // Default behavior: immediate expiration
```
vicb: IIUC this is not the default in Next 16.
It should only happen if `revalidateTag` has no secondf arg (deprecated) or with the max profile

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -34,6 +34,7 @@ export default {
       debug("composable cache result", result);
 
       // We need to check if the tags associated with this entry has been revalidated
```
vicb: Move that comment to line 42?

packages/tests-unit/tests/adapters/cache.test.ts
```diff
@@ -703,6 +860,168 @@ describe("CacheHandler", () => {
         },
       ]);
     });
+
+    describe("durations parameter", () => {
+      it("Should set expiry immediately when no durations provided (default behavior) - original mode", async () => {
```
vicb: Do we need to have those empty tests?
Maybe remove them and a a comment on the target test about all that it tests?

vicb (commented):
A few more comments and questions

packages/open-next/src/adapters/cache.ts
conico974: > Does that means that hasBeenRevalidated would only return true with the max profile?

It returns true only with {expire: 0} or if it has expired (i.e. revalidatedAt + expire)

packages/open-next/src/build/patch/patches/patchNextServer.ts
conico974: Not sure, it all depends on what they decide to change.
And this won't be needed with the adapters API

packages/open-next/src/utils/cache.ts
conico974: We could, but then typescript complains because it doesn't know if path exists in tag

packages/open-next/src/build/patch/patches/patchNextServer.ts
vicb: At least looking at [the test](https://github.com/opennextjs/opennextjs-aws/pull/1122/changes#diff-fd88a4c8f6ff97224daa5dc1e13d1f7c6dfe3d06c5cefc19bef63fe4e07fd1f7R424) it would prevent the lazy init of `internalWaitUntil`

.changeset/slick-crabs-hear.md
```diff
@@ -0,0 +1,14 @@
+---
+"@opennextjs/aws": minor
+---
+
+Add support for SWR (stale-while-revalidate) in `revalidateTag`
+
+Introduces a new methods in the tag cache (hasBeenStale, optional), both for the original and the next mode tag cache, mandatory for SWR.
```
vicb:
```suggestion
Introduces a new optional method `isStale` in the tag cache for both the original and the next modes. The implementation is mandatory for SWR to work.
```

.changeset/slick-crabs-hear.md
```diff
@@ -0,0 +1,14 @@
...
+
+It also introduces a RequestCache utils that can be used to cache things scoped to a request (stored in the OpenNext internal AsyncLocalStorage context)
```
vicb:
```suggestion
It also introduces a `RequestCache` utility that can be used to cache things scoped to a request (stored in the OpenNext internal AsyncLocalStorage context)
```

packages/open-next/src/adapters/cache.ts
vicb: What about adding JSDoc on the method in `packages/open-next/src/types/overrides.ts`?

packages/open-next/src/adapters/composable-cache.ts
```diff
@@ -149,4 +167,70 @@ export default {
     // This function does absolutely nothing
     return;
   },
+
+  /**
+   * Added in Next.js 16. Updates tags with optional stale/expire durations.
+   * Mirrors the logic in `Cache.revalidateTag` but without CDN invalidation
+   * since composable cache keys are not URL paths.
+   *
+   * When `durations` is provided, marks tags as stale immediately and optionally
```
vicb, line 176: nit: add details about the unit for `durations.expire` - it is a duration in ms

packages/open-next/src/adapters/dynamo-provider.ts
```diff
@@ -16,6 +16,12 @@ type DataType = {
   revalidatedAt: {
     N: string;
   };
+  stale?: {
+    N: string;
+  };
+  expire?: {
+    N: string;
+  };
```
vicb, line 30: nit: add JSDoc with 
- type: duration vs timestamp
- unit (ms)

packages/open-next/src/build/compileCache.ts
```diff
@@ -39,7 +33,7 @@ export function compileCache(
           `globalThis.disableDynamoDBCache = ${
             config.dangerous?.disableTagCache ?? false
           };`,
-          `globalThis.isNextAfter15 = ${isAfter15};`,
```
vicb, line 42: The removal of `isNextAfter15` should probably be listed in the breaking change? (alternatively, we could keep it as deprecated and remove it in the next major)

packages/open-next/src/overrides/tagCache/dummy.ts
```diff
@@ -16,6 +16,9 @@ const dummyTagCache: TagCache = {
   writeTags: async () => {
     return;
   },
+  hasBeenStale: async (_path: string) => {
```
vicb: Should this be `isStale` ?

packages/open-next/src/types/overrides.ts
```diff
@@ -198,6 +232,17 @@ export type OriginalTagCache = BaseTagCache & {
...
+   * Optional method to check if any tag entry for the given path has become stale (but not yet expired).
+   * There are three possible states for a cache entry:
+   * - **Fresh**: no tag/path combination has been revalidated since `lastModified` → returns `false`.
+   * - **Stale**: at least one tag/path combination was revalidated after `lastModified` but has not
+   *   yet expired → returns `true`. The cache entry is still served but `revalidate` is set to `1`
+   *   to trigger background revalidation.
+   * - **Expired**: at least one tag/path combination has fully expired → handled by `getLastModified`
+   *   returning `-1`. This method is only called when the entry is not fully expired.
+   */
+  hasBeenStale?(path: string, lastModified?: number): Promise<boolean>;
```
vicb: Should this be `isStale` ?

vicb (approved):
Thanks for the updates @conico974 

I have added a few minor comments but LGTM otherwise

PR #1135, opened 2026-04-07
fix: Handle multiple same-name headers as array in fetch proxyExternalRequest

### LINKED ISSUE #1134: BUG: proxyExternalRequest(fetch) response overwrites multiple headers with the same name.
#### Description
In proxyExternalRequest (fetch), multiple response headers with the same name are overwritten during the header filtering process. Only the last occurrence is preserved and sent to the client.

https://github.com/opennextjs/opennextjs-aws/blob/8210035cc30720702a6052e22bbf9275ef54c636/packages/open-next/src/overrides/proxyExternalRequest/fetch.ts#L16-L25

This behavior violates RFC 7230 and causes critical data loss, most notably for Set-Cookie headers.

#### Steps to Reproduce
- Use Next.js rewrites to proxy a request to an external service.
- The external service returns multiple Set-Cookie headers.
- Observe that the client only receives the final cookie.

#### Current Behavior
```
Server Send:
Set-Cookie: foo=foo;
Set-Cookie: bar=bar;

Client Received:
Set-Cookie: bar=bar;
```
#### Expect
```
Server Send:
Set-Cookie: foo=foo;
Set-Cookie: bar=bar;

Client Received:
Set-Cookie: foo=foo;
Set-Cookie: bar=bar;
```

ktKongTong:
resolve: https://github.com/opennextjs/opennextjs-aws/issues/1134

### DIFF
--- a/packages/tests-unit/tests/overrides/proxyExternalRequest/fetch.test.ts
+++ b/packages/tests-unit/tests/overrides/proxyExternalRequest/fetch.test.ts
@@ -1,7 +1,16 @@
 import fetchProxy from "@opennextjs/aws/overrides/proxyExternalRequest/fetch.js";
-import { vi } from "vitest";
+import { afterEach, beforeEach, vi } from "vitest";
 
 describe("proxyExternalRequest/fetch", () => {
+  let originalFetch: typeof global.fetch;
+
+  beforeEach(() => {
+    originalFetch = globalThis.fetch;
+  });
+
+  afterEach(() => {
+    globalThis.fetch = originalFetch;
+  });
   // Note: if the url is hosted on the Cloudflare network we want to make sure that a `cf-connecting-ip` header is not being sent as that causes a DNS error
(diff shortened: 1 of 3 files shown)


khuezy (approved):
LGTM thanks!

packages/tests-unit/tests/overrides/proxyExternalRequest/fetch.test.ts
```diff
@@ -36,4 +36,28 @@ describe("proxyExternalRequest/fetch", () => {
...
+    responseHeaders.append("set-cookie", "foo=value1");
+    responseHeaders.append("set-cookie", "bar=value2");
+    responseHeaders.append("Set-Cookie", "cookie=value3");
+    const fetchMock = vi.fn<typeof global.fetch>(
+      async () =>
+        new Response(null, {
+          headers: responseHeaders,
+        }),
+    );
+    globalThis.fetch = fetchMock;
```
conico974, line 59: We should reset the global fetch here after this test is done

conico974:
Thanks for your PR

PR #1139, opened 2026-04-10
`applyRule` does not require a `fix` field

vicb:
A few changes in the ASTGrep utils:

- The `fix` is not more required in `applyRule`. That's because you might only be interested in the returned `matches`
- The `fix` can be an empty string (to delete code)
- Added a `parseCode` function to parse source code - the existing `parse` takes a path.


khuezy (approved):

james-elicx (approved):

vicb:
Thanks for the reviews!

PR #1142, opened 2026-04-11
Enhance cache stale checking logic

conico974:
Fix cache interceptor not using swr stale logic.
Fix next mode not using stale properly
Update e2e test to reflect that

### DIFF
--- a/.changeset/big-turtles-move.md
+++ b/.changeset/big-turtles-move.md
@@ -3,3 +3,4 @@
 ---
 
 Fix cache interceptor isStale handling for Next 16+
+Fix next mode tag cache not retrieving stale data properly
--- a/.changeset/big-turtles-move.md
+++ b/.changeset/big-turtles-move.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Fix cache interceptor isStale handling for Next 16+
--- a/packages/open-next/src/core/routing/cacheInterceptor.ts
+++ b/packages/open-next/src/core/routing/cacheInterceptor.ts
@@ -68,11 +68,15 @@ async function computeCacheControl(
       etag,
     };
   }
-  const isTimeStale = finalRevalidate !== CACHE_ONE_YEAR && Math.max(finalRevalidate - age, 1) === 1;
+  const isTimeStale =
+    finalRevalidate !== CACHE_ONE_YEAR &&
+    Math.max(finalRevalidate - age, 1) === 1;
   const isStale = isTimeStale || isStaleFromTagCache;
 
   if (finalRevalidate !== CACHE_ONE_YEAR || isStaleFromTagCache) {
-    const sMaxAge = isStaleFromTagCache ? 1 : Math.max(finalRevalidate - age, 1);
+    const sMaxAge = isStaleFromTagCache
+      ? 1
+      : Math.max(finalRevalidate - age, 1);
     debug("sMaxAge", {
       finalRevalidate,
       age,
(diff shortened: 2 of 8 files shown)


packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -67,19 +68,25 @@ async function computeCacheControl(
...
-    const sMaxAge = Math.max(finalRevalidate - age, 1);
+  const isTimeStale =
+    finalRevalidate !== CACHE_ONE_YEAR &&
+    Math.max(finalRevalidate - age, 1) === 1;
+  const isStale = isTimeStale || isStaleFromTagCache;
+
+  if (finalRevalidate !== CACHE_ONE_YEAR || isStaleFromTagCache) {
+    const sMaxAge = isStaleFromTagCache
+      ? 1
+      : Math.max(finalRevalidate - age, 1);
```
khuezy: Nit: redundant ` Math.max(finalRevalidate - age, 1)`

.changeset/big-turtles-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Fix cache interceptor isStale handling for Next 16+
```
vicb, line 5: The PR description lists 2 fixes.
They should probably both be listed here.

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -67,19 +68,25 @@ async function computeCacheControl(
       etag,
     };
   }
-  if (finalRevalidate !== CACHE_ONE_YEAR) {
-    const sMaxAge = Math.max(finalRevalidate - age, 1);
+  const isTimeStale =
```
vicb: nits: 
- should the name be `isStaleFromTime` for consistency ?
- add a comment why `CACHE_ONE_YEAR` is special cased

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -67,19 +68,25 @@ async function computeCacheControl(
...
+    finalRevalidate !== CACHE_ONE_YEAR &&
+    Math.max(finalRevalidate - age, 1) === 1;
+  const isStale = isTimeStale || isStaleFromTagCache;
+
+  if (finalRevalidate !== CACHE_ONE_YEAR || isStaleFromTagCache) {
```
vicb: nit

`finalRevalidate !== CACHE_ONE_YEAR` is computed twice too.

Extract to a const with a comment?

vicb (approved):
LGTM with a few nits

packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -292,6 +301,17 @@ export async function cacheInterceptor(
           return event;
         }
       }
+
+      // Check if the cache entry is stale (valid but needs background revalidation)
+      const tags = getTagsFromValue(cachedData.value);
```
edmundhung: It looks like `getTagsFromValue()` mutates `cachedData.value` by deleting `x-next-cache-tags` after reading it. Since this path already calls it a few lines above for `hasBeenRevalidated()`, this second call will likely miss the tags.

Could we extract tags once and reuse them for both `hasBeenRevalidated()` and `isStale()`?

conico974: Good catch, thanks

PR #1144, opened 2026-04-16
Replace NEXT_BUILD_ID with OPEN_NEXT_BUILD_ID for consistency

conico974:
Deprecate the NEXT_BUILD_ID environment variable in favor of OPEN_NEXT_BUILD_ID to enhance consistency and maintain backward compatibility.

### DIFF
--- a/packages/open-next/src/adapters/config/index.ts
+++ b/packages/open-next/src/adapters/config/index.ts
@@ -41,5 +41,10 @@ export const AppPathRoutesManifest =
 export const FunctionsConfigManifest =
   /* @__PURE__ */ loadFunctionsConfigManifest(NEXT_DIR);
 
+/**
+ * @deprecated Do not use NEXT_BUILD_ID above Next 16.2, as it could be a fixed value if used in conjonction with deploymentId
+ * We keep it for backward compatibility
+ */
 process.env.NEXT_BUILD_ID = BuildId;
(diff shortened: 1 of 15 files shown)


packages/open-next/src/adapters/config/index.ts
```diff
@@ -41,5 +41,10 @@ export const AppPathRoutesManifest =
 export const FunctionsConfigManifest =
   /* @__PURE__ */ loadFunctionsConfigManifest(NEXT_DIR);
 
+/**
+ * @deprecated Do not use NEXT_BUILD_ID above Next 16.2, as it could be a fixed value if used in conjonction with deploymentId
```
sommeeeer, line 45: conjunction

sommeeeer (approved):

packages/open-next/src/plugins/edge.ts
```diff
@@ -195,6 +195,7 @@ ${contents}
   export const PagesManifest = ${JSON.stringify(PagesManifest)};
 
   process.env.NEXT_BUILD_ID = BuildId;
```
khuezy, line 197: Does this need to be deleted?

conico974: I prefer to keep it for now, in case people were using it. Especially for custom overrides

PR #1146, opened 2026-04-19
chore(deps): bump node-minify

oliverphardman:
- Upgraded `@node-minify/core` and `@node-minify/terser` from version `^8.0.6` to `^10.5.0` in `package.json`
- Updated import statements for `minify` and `terser` to use named imports, as required by the new version of `@node-minify`.
- Added a file size check in `minimize-js.ts` to skip minification for empty files as terser no longer handles them gracefully.

This should reduce the security warnings shown in the console.


ladykirsah:
Gentle bump on this — the `@node-minify/core@8.0.6` chain it fixes is now a flagged security advisory for downstream consumers.

`@opennextjs/aws@4.1.0` still pins `@node-minify/core@^8.0.6`, whose transitive `glob@9 → minimatch@8 → brace-expansion@2.1.2` falls under [GHSA-mh99-v99m-4gvg / CVE-2026-14257](https://github.com/advisories/GHSA-mh99-v99m-4gvg) (brace-expansion DoS; flat range `<= 5.0.7`, patched in `5.0.8`). This PR's bump to `@node-minify/core@^10.5.0` pulls `minimatch@10`, which is the only compatible way to reach a patched brace-expansion on that path.

The security impact — and why consumers can't safely work around it themselves — is written up in #1204: a naive `brace-expansion` override is a false green (`npm audit` reports `0 vulnerabilities` while the build breaks with `expand is not a function`, since `minimatch@8` imports the default export that `5.x` dropped).

Would it be possible to get a review here? Happy to help test against a Next.js + `@opennextjs/cloudflare` build if that's useful. Thanks for maintaining OpenNext!

PR #1148, opened 2026-04-22
Define `globalThis.nextVersion` globally

vicb:
Validated with the Cloudflare adapter


conico974 (approved):

vicb:
Thanks for the review Nico!

PR #1149, opened 2026-04-23
fix(cache): use the Open Next Build Id for the cache assets folder

vicb:
The build id should be the Open Next Build Id introduced in #1144

The current code is causing troubles in the Cloudflare adapter which retrieves the build id from the path.
I can confirm that the fix here solve the issue in the Cloudflare adapter.


conico974 (approved):

vicb:
Thanks for the review Nico!

PR #1151, opened 2026-04-24
fix: include `package.json#imports` re-maps in traced files

james-elicx:
Next's NFT tracer does not appear to follow the `imports` field in `package.json`, so packages only reachable via subpath remaps (e.g. `@mathjax/src`'s `"#mhchem/*"` -> `"mhchemparser/esm/*"`) are missing from the traced server bundle, and built-time and runtime fails to resolve them in the Cloudflare adapter.

`copyTracedFiles` now scans every traced `package.json` for an `imports` field and, for any bare-specifier remap target, resolves the target package from the consumer's real location and pulls its source files in.

This is tested via https://github.com/opennextjs/opennextjs-cloudflare/pull/1231 with a reproduction for https://github.com/opennextjs/opennextjs-cloudflare/issues/1229 (would be a good candidate for monorepo!)

The root cause of the issue seems to be a bug in Next.js, see https://github.com/vercel/next.js/issues/93295

### DIFF
--- a/.changeset/copytracedfiles-imports-remap.md
+++ b/.changeset/copytracedfiles-imports-remap.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: include `package.json#imports` subpath remap targets in traced files.
+
+Next's NFT tracer does not follow the `imports` field, so packages only reachable via remaps (e.g. `@mathjax/src`'s `#mhchem/*` → `mhchemparser/esm/*`) were missing from the server bundle. `copyTracedFiles` now scans traced package.json files for an `imports` field and pulls in any bare-specifier remap targets from source.
--- a/.changeset/copytracedfiles-imports-remap.md
+++ b/.changeset/copytracedfiles-imports-remap.md
@@ -5,3 +5,5 @@
 fix: include `package.json#imports` subpath remap targets in traced files.
 
 Next's NFT tracer does not follow the `imports` field, so packages only reachable via remaps (e.g. `@mathjax/src`'s `#mhchem/*` → `mhchemparser/esm/*`) were missing from the server bundle. `copyTracedFiles` now scans traced package.json files for an `imports` field and pulls in any bare-specifier remap targets from source.
+
+Workaround for upstream issue https://github.com/vercel/next.js/issues/93295.
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -1,6 +1,7 @@
 import url from "node:url";
 
 import {
+  type Dirent,
   chmodSync,
   copyFileSync,
   cpSync,
@@ -13,6 +14,7 @@ import {
   symlinkSync,
   writeFileSync,
 } from "node:fs";
+import { createRequire } from "node:module";
 import path from "node:path";
 
 import {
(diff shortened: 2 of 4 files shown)


vicb (commented):
Thanks for looking that quickly @james-elicx 

Let's double check the bundle size, a user is reporting a big delta in https://github.com/opennextjs/opennextjs-cloudflare/pull/1231. It might be specific to his app that was not working before though.

We should double check any potential impact on other apps (i.e. the examples).

Thanks!

james-elicx:
Replied on the other PR with all the before / after sizes for the apps.

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -85,6 +87,152 @@ export function isNonLinuxPlatformPackage(srcPath: string): boolean {
   return nonLinuxPlatformRegex.test(srcPath);
 }
 
+const packageJsonPathRegex = getCrossPlatformPathRegex(
+  String.raw`/package\.json$`,
+  { escape: false },
+);
+
+function collectImportTargets(value: unknown, out: string[]): void {
```
vicb: Could you please add JSDoc (comment, `@param` and `@return`) and tests for the added functions?

vicb (commented):
There is a simple repro in https://github.com/opennextjs/opennextjs-cloudflare/issues/1229#issuecomment-4323120434

IIUC the root cause of the issue is because the Next tracing doesn't consider the `imports` field of the package.json.

It would be great to investigate the root cause and check if there is an exisiting issue on the Next repo and if not create a new one.

Are both turbopack and webpack affected by the issue ?

Do standalone builds work? I guess they should have the same issue - you might need to re-root the standalone folder outside of the app folder when testing that to make sure the top level node_modules is not used.

If we confirm the bug is on the Next side, it will take some time to get this fix so we should merge this PR in the meantime (and for versions before the fix). Could you please add comments/tests?

Thanks

james-elicx:
I can take a look this evening - my OpenNext stuff is on my other laptop.

james-elicx:
> Are both turbopack and webpack affected by the issue ?
>
> Do standalone builds work?

Tested the standalone build - it's failing for both Webpack and Turbopack, so yep, it impacts Next.js as well. Thanks for the suggestion.

Now that GitHub issues has started working again, I was looking at Next's issue tracker and it appears that one of the Node.js team posted an issue today about import aliases not working as well - https://github.com/vercel/next.js/issues/93295

.changeset/copytracedfiles-imports-remap.md
```diff
@@ -0,0 +1,7 @@
+---
```
vicb, line 1: Could you please mention https://github.com/vercel/next.js/issues/93295 in there?

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -85,6 +87,197 @@ export function isNonLinuxPlatformPackage(srcPath: string): boolean {
   return nonLinuxPlatformRegex.test(srcPath);
 }
 
+const packageJsonPathRegex = getCrossPlatformPathRegex(
```
vicb: What about moving all the code in a separate file and mention that it is a workaround for https://github.com/vercel/next.js/issues/93295 in this file?

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -85,6 +87,197 @@ export function isNonLinuxPlatformPackage(srcPath: string): boolean {
...
+      if (packageJsonPathRegex.test(srcEntry))
+        pending.push([srcEntry, dstEntry]);
+    }
+  }
+}
+
+/**
+ * Augment `filesToCopy` with targets of `package.json#imports` subpath remaps.
+ *
+ * Next's NFT tracer does not follow the `imports` field, so packages that
```
vicb: Please link https://github.com/vercel/next.js/issues/93295 here too

vicb (commented):
Thanks a lot for digging into the issue and finding the root cause in Next.js

I have added a few comments on the PR.

The idea would be to isolate the changes in a separate file and make it clear we do this to workaround a Next.js tracing issue.

It will be easier to remove the code that way when it is not needed any more.

🙏

vicb (approved):
LGTM,

Thanks for the fix and the issue James.

Feel free to merge anytime after the fix has been validated on the original Cloudflare issue.

PR #1166, opened 2026-05-12
fix(build): use basename to detect package.json in copyTracedFiles

ash1day:
## What

`copyTracedFiles` decides whether a traced file is a node-package manifest with

```ts
if (module.endsWith("package.json")) {
  nodePackages.set(path.dirname(module), path.dirname(dst));
}
```

`endsWith` matches any filename whose tail is `package.json`, so files like `care-package.json` or `my-package.json` get registered as if their parent directory were a node package. The map is then handed to the Cloudflare adapter's `copyWorkerdPackages`, which tries to read `<parent-dir>/package.json` and surfaces:

```
ERROR Failed to copy /path/to/data/tables
```

The actual file copy is unaffected — that goes through `filesToCopy`, which is keyed by full paths — so the error is purely cosmetic. But it's misleading enough that someone hitting it usually has to dig into OpenNext's source to convince themselves the deploy is fine, which is what motivated this patch.

## Repro

In a Next.js app, statically import a JSON data file whose name ends in `-package.json`:

```ts
// src/lib/foo.ts
import careData from '@@/data/tables/care-package.json' assert { type: 'json' };
```

Next traces that file, OpenNext walks the `.nft.json`, and the inline `endsWith` check fires on `care-package.json`. Cloudflare adapter's deploy log then prints `Failed to copy <dir>` for each affected directory.

## Fix

Switch the inline check to a real basename comparison. The helper is exported so the regression is covered by a unit test alongside `isExcluded` / `isNonLinuxPlatformPackage`.

```ts
export function isPackageJson(modulePath: string): boolean {
  const lastSep = Math.max(
    modulePath.lastIndexOf("/"),
    modulePath.lastIndexOf("\\"),
  );
  return modulePath.slice(lastSep + 1) === "package.json";
}
```

`path.basename` was avoided because traced paths can come from either platform after `path.join`, and `path.basename` only splits on the OS-native separator. The manual scan handles both.

## Tests

Added three cases in `packages/tests-unit/tests/build/copyTracedFiles.test.ts`:

- positive: `…/next/package.json` and bare `package.json` → `true`
- regression: `…/care-package.json`, `…/my-package.json` → `false`
- cross-platform: Windows-style `\` separators behave the same

```
✓ packages/tests-unit/tests/build/copyTracedFiles.test.ts (12 tests) 3ms
```

## Risk

Behavior only changes for filenames that share the `package.json` suffix but aren't actually package manifests. Real `package.json` traces continue to be picked up identically.

## Changeset

Included a `patch` changeset under `.changeset/fix-copytraced-package-json-suffix.md`.

### DIFF
--- a/packages/open-next/src/build/copyTracedFiles.ts
+++ b/packages/open-next/src/build/copyTracedFiles.ts
@@ -85,6 +85,16 @@ export function isNonLinuxPlatformPackage(srcPath: string): boolean {
   return nonLinuxPlatformRegex.test(srcPath);
 }
 
+// `path.basename` only splits on the OS-native separator, so we explicitly
+// look at both `/` and `\` to handle paths produced under either platform.
+export function isPackageJson(modulePath: string): boolean {
+  const lastSep = Math.max(
+    modulePath.lastIndexOf("/"),
+    modulePath.lastIndexOf("\\"),
+  );
+  return modulePath.slice(lastSep + 1) === "package.json";
+}
+
 function copyPatchFile(outputDir: string) {
   const patchFile = path.join(__dirname, "patch", "patchedAsyncStorage.js");
   const outputPatchFile = path.join(outputDir, "patchedAsyncStorage.cjs");
@@ -153,7 +163,7 @@ export async function copyTracedFiles({
       filesToCopy.set(src, dst);
 
       const module = path.join(dotNextDir, subDir, tracedPath);
-      if (module.endsWith("package.json")) {
+      if (isPackageJson(module)) {
         nodePackages.set(path.dirname(module), path.dirname(dst));
       }
     });
--- a/.changeset/fix-copytraced-package-json-suffix.md
+++ b/.changeset/fix-copytraced-package-json-suffix.md
@@ -1,0 +1,10 @@
+---
+"@opennextjs/aws": patch
+---
+
+`copyTracedFiles`: only register a traced file as a node package when its
+basename is exactly `package.json`, not when its filename merely ends with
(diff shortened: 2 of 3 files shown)


sommeeeer (approved):
LGTM!

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -85,6 +85,16 @@ export function isNonLinuxPlatformPackage(srcPath: string): boolean {
...
+// `path.basename` only splits on the OS-native separator, so we explicitly
+// look at both `/` and `\` to handle paths produced under either platform.
+export function isPackageJson(modulePath: string): boolean {
+  const lastSep = Math.max(
+    modulePath.lastIndexOf("/"),
+    modulePath.lastIndexOf("\\"),
+  );
+  return modulePath.slice(lastSep + 1) === "package.json";
+}
+
```
vicb, line 97: We should avoid using OS specific constructs as much as possible.
I don't think this is needed here.

packages/open-next/src/build/copyTracedFiles.ts
```diff
@@ -153,7 +163,7 @@ export async function copyTracedFiles({
       filesToCopy.set(src, dst);
 
       const module = path.join(dotNextDir, subDir, tracedPath);
-      if (module.endsWith("package.json")) {
+      if (isPackageJson(module)) {
```
vicb, line 166: Would that work instead?

```suggestion
      if (path.basename(module) === "package.json") {
```

vicb (changes requested):
I guess using `path.basename` could be a proper fix?

PR #1176, opened 2026-05-30
fix(security): 2 improvements across 1 files

andoan16:
## Summary

fix(security): 2 improvements across 1 files

## Problem

**Severity**: `Medium` | **File**: `packages/open-next/src/overrides/imageLoader/s3.ts:L10`

The S3 image loader has a hardcoded region 'us-east-1' which could lead to issues if the bucket is in a different region. This should be configurable via environment variables.

## Solution

Replace the hardcoded region with an environment variable like process.env.S3_REGION || 'us-east-1'

## Changes

- `packages/open-next/src/overrides/imageLoader/s3.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/packages/open-next/src/overrides/imageLoader/s3.ts
+++ b/packages/open-next/src/overrides/imageLoader/s3.ts
@@ -6,7 +6,7 @@ import { FatalError } from "utils/error";
 
 import { awsLogger } from "../../adapters/logger";
 
-const { BUCKET_NAME, BUCKET_KEY_PREFIX } = process.env;
+const { BUCKET_NAME, BUCKET_KEY_PREFIX, S3_REGION } = process.env;
 
 function ensureBucketExists() {
   if (!BUCKET_NAME) {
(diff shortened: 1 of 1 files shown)


packages/open-next/src/overrides/imageLoader/s3.ts
```diff
@@ -17,7 +17,7 @@ function ensureBucketExists() {
 const s3Loader: ImageLoader = {
   name: "s3",
   load: async (key: string) => {
-    const s3Client = new S3Client({ logger: awsLogger });
+    const s3Client = new S3Client({ region: S3_REGION || 'us-east-1', logger: awsLogger });
```
conico974, line 20: The function is deployed on lambda, the region is set there. We should absolutely not set it to `us-east-1`

conico974 (changes requested):

PR #1186, opened 2026-06-16
fix: Cloudflare streaming response backpressure

Cooperrr:
Prevents excessive memory usage when streaming large responses. Previously, writes did not account for consumer backpressure, so a fast producer could buffer large amounts of data when the consumer was slow. This PR pauses pulling from the producer until the consumer is ready for more data.

~~Also skips `ReadableStream.tee()` when chunks do not need to be retained, avoiding an unused duplicate response body during direct streaming.~~

### DIFF
--- a/packages/open-next/src/core/requestHandler.ts
+++ b/packages/open-next/src/core/requestHandler.ts
@@ -162,6 +162,8 @@ export async function openNextHandler(
           response.statusCode = routingResult.statusCode;
           response.flushHeaders();
           const [bodyToConsume, bodyToReturn] = routingResult.body.tee();
+          // Use pipeline so the streamCreator Writable can apply backpressure
+          // instead of eagerly consuming the whole Web stream.
           await pipeline(Readable.fromWeb(bodyToConsume), response);
           routingResult.body = bodyToReturn;
         }
--- a/packages/open-next/src/core/requestHandler.ts
+++ b/packages/open-next/src/core/requestHandler.ts
@@ -161,13 +161,9 @@ export async function openNextHandler(
           );
           response.statusCode = routingResult.statusCode;
           response.flushHeaders();
-          let bodyToConsume = routingResult.body;
-          if (options.streamCreator.retainChunks !== false) {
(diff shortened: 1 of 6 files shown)


khuezy:
Hi thanks for the PR! LGTM but I'll defer to the Cloudflare folks.

packages/open-next/src/core/requestHandler.ts
```diff
@@ -159,12 +161,13 @@ export async function openNextHandler(
           );
           response.statusCode = routingResult.statusCode;
           response.flushHeaders();
-          const [bodyToConsume, bodyToReturn] = routingResult.body.tee();
-          for await (const chunk of bodyToConsume) {
-            response.write(chunk);
+          let bodyToConsume = routingResult.body;
+          if (options.streamCreator.retainChunks !== false) {
```
conico974: This is a breaking change, this PR cannot just be a patch. 
It will also likely break some converters that relies on the body being readable 
We could gate that under an env variable that we'll remove on a future major

packages/open-next/src/overrides/wrappers/cloudflare-node.ts
conico974, line 1: Add more comment on your change please, this is quite hard to review and follow what's going on

conico974 (changes requested):
Could you fix the tests as well please.

packages/open-next/src/core/requestHandler.ts
Cooperrr: Thanks for taking the time to review. 

I've reverted back to keeping a copy of the body to avoid breaking changes (would be nice to remove down the line as you mentioned in a future major) but not essential for this PR.

packages/open-next/src/overrides/wrappers/cloudflare-node.ts
Cooperrr: I've tried to better explain what's going on with https://github.com/opennextjs/opennextjs-aws/pull/1186/commits/961a03b5fb320e6918bf23b2061e8c68359df6a2

PR #1189, opened 2026-06-23
fix(cacheInterceptor): handle index route cache key

gaauwe:
We noticed that our index page was not being served by the cache interceptor even though `enableCacheInterception` was enabled.

This is similar to #860, but the existing fix still uses `??` fallbacks. After trailing slash normalization, `/` becomes an empty string, so `localizedPath ?? "/"` still evaluates to `""`.

This normalizes the route path back to `/` for ISR detection, while keeping `/index` as the cache lookup key to match the generated cache asset.

Verified against a Cloudflare/OpenNext build:
- prerender manifest route key: `/`
- generated cache asset: `index.cache`
- populated cache key: `/index`

### DIFF
--- a/packages/open-next/src/core/routing/cacheInterceptor.ts
+++ b/packages/open-next/src/core/routing/cacheInterceptor.ts
@@ -295,7 +295,7 @@ export async function cacheInterceptor(
       ) {
         const _hasBeenRevalidated = cachedData.shouldBypassTagCache
           ? false
-          : await hasBeenRevalidated(localizedPath, tags, cachedData);
+          : await hasBeenRevalidated(cacheKey, tags, cachedData);
 
         if (_hasBeenRevalidated) {
           return event;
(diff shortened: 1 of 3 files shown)


packages/open-next/src/core/routing/cacheInterceptor.ts
```diff
@@ -265,23 +265,23 @@ export async function cacheInterceptor(
   localizedPath = localizedPath.replace(/\/$/, "");
 
   // Then we decode the path params
-  localizedPath = decodePathParams(localizedPath);
+  localizedPath = decodePathParams(localizedPath) || "/";
```
conico974, line 268: localizedPath is used in the rest of the files in a couple of places (i.e. `hasBeenRevalidated`, `isStale`).
Did you check if it has the right value there ?

conico974 (changes requested):
Could you add some unit test ?

packages/open-next/src/core/routing/cacheInterceptor.ts
gaauwe: Thanks for the review! It seems that I indeed missed the usage in `hasBeenRevalidated` and `isStale`, which should use the `cacheKey`.
 
I've added some tests to validate the behavior:
* `/albums` still uses `/albums` for the cache lookup and 'original-mode' cache checks, so existing non-index behavior is unchanged
* The new index cache key test fails without the route normalization, and now passes.
* The same index cache key test failed after my initial fix, because `hasBeenRevalidated` and `isStale` still received `/`
    * It now verifies that 'original-mode' tag cache checks use `/index`
* The new stale index route path test verifies that background revalidation still uses the public route `/`, not the cache key `/index`

As you can imagine this code is all new to me, so let me know if there is anything I can improve (:

conico974 (approved):
LGTM, Thanks for the PR

PR #1194, opened 2026-07-05
Fix cache handling for optional rscData

conico974:
Improve fallback behavior for optional rscData in the cache, ensuring it handles undefined values gracefully. Additionally, apply linting changes and include a changeset for better tracking.


packages/open-next/src/adapters/cache.ts
```diff
@@ -169,6 +169,12 @@ export default class Cache {
         } as CacheHandlerValue;
       }
       if (cacheData?.type === "page" || cacheData?.type === "app") {
+        // Next.js only writes the `.html` file for prerendered pages, without it there is
+        // no valid value to return so we report a miss and let the server render the page.
+        if (cacheData.html === undefined) {
```
conico974, line 178: Need to verify that, but I think that some earlier version of PPR were not setting any html (not sure which one)

vicb (approved):

PR #1197, opened 2026-07-12
fix(converters): preserve query param encoding in getQueryFromSearchParams

azuma317:
## What

`getQueryFromSearchParams` parsed query values by iterating over `URLSearchParams.entries()`, which **decodes** each value. However, since #817, `convertToQueryString` intentionally does **not** re-encode values (it concatenates them raw to preserve their original encoding). This leaves the two sides of the round-trip inconsistent:

- **parse** (`getQueryFromSearchParams`) → decoded values
- **serialize** (`convertToQueryString`) → assumes already-encoded values, no re-encoding

As a result, any query value containing reserved characters (`&`, `=`, `+`, spaces, …) is corrupted when `req.url` is rebuilt from the parsed query in `requestHandler.ts`:

```ts
req.url =
  initialURL.pathname +
  convertToQueryString(routingResult.internalEvent.query);
```

A decoded value such as `https://example.com/auth?a=1&b=2` gets concatenated as-is, so the embedded `&`/`=` split the outer query string and everything after the first reserved char is lost or reinterpreted as separate params.

## Reproduction

This is the root cause of opennextjs/opennextjs-cloudflare#1134. It reliably breaks, for example, Sign in with Apple via `better-auth`'s Expo flow: the client hits `…/expo-authorization-proxy?authorizationURL=<percent-encoded URL>`, the encoded `%26`/`%3D` inside `authorizationURL` are decoded to `&`/`=`, the URL is truncated at the first `&`, the `state` param is lost, and the endpoint returns `400 "Unexpected error"`. It only manifests in production (Workers / any deployment that goes through this `req.url` reconstruction), not in `next dev`.

## Fix

Keep the parsed values **percent-encoded** so they round-trip correctly through `convertToQueryString`. This completes the model established in #817 (query records hold encoded values end-to-end) rather than reverting `convertToQueryString` — which would reintroduce the double-decode problem #817 fixed and break the existing "should respect existing query encoding" test.

Multi-value grouping and the empty-query case are preserved.

## Tests

Added unit tests for `getQueryFromSearchParams` covering the empty case, single value, repeated-key grouping, and the reserved-character round-trip from #1134. Full unit suite (`604 passed`), `biome check`, and `tsc --noEmit` all pass.

### DIFF
--- a/.changeset/preserve-query-param-encoding.md
+++ b/.changeset/preserve-query-param-encoding.md
@@ -2,8 +2,10 @@
 "@opennextjs/aws": patch
 ---
 
-Preserve percent-encoding of query parameters in `getQueryFromSearchParams`
+Preserve percent-encoding of query parameters in `getQueryFromSearchParams` and `convertToQuery`
 
-`getQueryFromSearchParams` iterated over `URLSearchParams.entries()`, which decodes the values. Since `convertToQueryString` does not re-encode them (by design, see #817), any query value containing reserved characters (`&`, `=`, `+`, spaces, ...) was corrupted when the query string was rebuilt for `req.url`. The parser now keeps the values encoded so they round-trip correctly through `convertToQueryString`.
+Both query parsers read values through `URLSearchParams`, which decodes them. Since `convertToQueryString` does not re-encode them (by design, see #817), any query value containing reserved characters (`&`, `=`, `+`, spaces, ...) was corrupted when the query string was rebuilt — for `req.url` in the request handler, and for the origin `querystring` in the CloudFront converter. The parsers now keep the values encoded so they round-trip correctly through `convertToQueryString`.
+
+`getQueryFromSearchParams` covers the `node` and `edge` converters; `convertToQuery` covers the `aws-apigw-v2` and `aws-cloudfront` converters.
 
 Fixes opennextjs/opennextjs-cloudflare#1134.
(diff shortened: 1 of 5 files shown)


clichedmoog:
Thanks for picking this up. One scope note: this covers `getQueryFromSearchParams`, which the `node` and `edge` converters use, but the default AWS converter builds its query differently:

```ts
// overrides/converters/aws-apigw-v2.ts
query: removeUndefinedFromQuery(convertToQuery(rawQueryString)),
```

`convertToQuery` in `core/routing/util.ts` still decodes via `new URLSearchParams(...)`, so on the `aws-lambda` wrapper values still reach `convertToQueryString` already decoded. Verified on 4.0.3 behind CloudFront: a `returnUrl` value containing `&` loses everything after the first inner param.

Would it make sense to give `convertToQuery` the same treatment here, so the AWS path is covered too? Happy to test the branch against our deployment if you do.

azuma317:
Thanks for the detailed report and for verifying this on a real deployment — you're right, my fix didn't cover the default AWS converters.

I've pushed 1bf3729 which applies the same treatment to `convertToQuery`: values are kept percent-encoded so they round-trip unchanged through `convertToQueryString`. This covers both `aws-apigw-v2` and `aws-cloudfront`, and I added a unit test mirroring your `returnUrl` scenario.

If you could test the branch against your CloudFront deployment, that would be much appreciated!

PR #1209, opened 2026-08-07
fix: persist runtime tags for APP_PAGE cache entries

### LINKED ISSUE #1180: revalidateTag() does not invalidate composable cache entries ('use cache' + cacheTag)
# `revalidateTag()` does not invalidate composable cache entries (`'use cache'` + `cacheTag`)

## Description

When using Next.js 16's composable cache (`cacheComponents: true` + `'use cache'` directive + `cacheTag()`), calling `revalidateTag()` does not cause subsequent requests to serve fresh content. The page continues to return stale cached data indefinitely.

The **write side works correctly** — DynamoDB tag invalidation records are created with the proper `revalidatedAt` timestamp. The **read side is broken** — when the Lambda serves a request, the composable cache `get()` reads the S3 entry, checks DynamoDB, but incorrectly returns the entry as fresh (`x-nextjs-cache: HIT`).

Bypassing the cache with the `x-prerender-revalidate` header returns updated content (`x-nextjs-cache: MISS`), confirming the data source is fresh and only the cache read path is broken.

This reproduces on both `@opennextjs/aws` 3.10.4 and 4.0.3. It works correctly on `next start` locally.

## Reproduction

Minimal repro repo: https://github.com/mattalco/opennext-repro

### Setup

```bash
pnpm install
npx open-next build
sst deploy
```

### `next.config.ts`

```ts
import type { NextConfig } from 'next';

  output: 'standalone',
  cacheComponents: true,
};

export default nextConfig;
```

### `open-next.config.ts`

```ts

const config = {
  default: {},
} satisfies OpenNextConfig;

export default config;
```

### `src/app/page.tsx`

```tsx
import { cacheLife, cacheTag } from 'next/cache';

async function getData() {
  'use cache';
  cacheLife('max');
  cacheTag('pages');

  const res = await fetch('https://jsonplaceholder.typicode.com/posts/1', {
    cache: 'no-store',
  });
  const post = await res.json();

  return { title: post.title, renderedAt: Date.now() };
}

export default async function Page() {
  const data = await getData();

  return (
    <main>
      <h1>OpenNext Composable Cache Repro</h1>
      <p>Title: {data.title}</p>
      <p>Rendered at: {new Date(data.renderedAt).toISOString()}</p>
    </main>
  );
}
```

### `src/app/api/revalidate/route.ts`

```ts
import { revalidateTag } from 'next/cache';
import { NextResponse } from 'next/server';

export async function GET() {
  revalidateTag('pages', { expire: 0 });

  return NextResponse.json({
    revalidated: true,
    tag: 'pages',
    now: Date.now(),
  });
}
```

## Steps to reproduce

1. Deploy the app with `sst deploy`
2. Visit the page — note the "Rendered at" timestamp
3. Hit `/api/revalidate` — response confirms `{ revalidated: true }`
4. Refresh the page — **timestamp does not change** (bug)
5. Add `x-prerender-revalidate: <token>` header — timestamp updates (proves data is fresh)

## Expected behavior

After calling `revalidateTag('pages')`, the next request to the page should serve fresh content with an updated timestamp.

## Actual behavior

The page continues to serve the stale cached entry. The `x-nextjs-cache: HIT` header confirms the composable cache is returning the old entry without honoring the tag invalidation.

## Root cause analysis

The composable cache write path correctly records tag invalidation in DynamoDB:
- `revalidateTag('pages')` writes `revalidatedAt` / `expire` / `stale` timestamps for composable cache entries tagged with `pages`

The composable cache read path does **not** correctly check these records:
- On a subsequent request, `composable-cache.get()` reads the cached entry from S3
- It checks DynamoDB for tag invalidation records
- It incorrectly determines the entry is still valid and returns it as a cache HIT

## Environment

- `next`: 16.2.7
- `@opennextjs/aws`: 4.0.3 (also reproduced on 3.10.4)
- `sst`: 4.15.2
- Deployment: AWS Lambda + S3 + DynamoDB (default SST Nextjs component)

cgcompassion:
Fixes #1180

## Summary

- persist runtime `x-next-cache-tags` associations for legacy `PAGE` and modern `APP_PAGE` entries when using an original-mode tag cache
- normalize both string and string-array cache-tag headers
- add deployed E2E coverage for App Router on-demand pages and experimental `use cache` pages
- add focused unit coverage and a patch changeset for `@opennextjs/aws`

## Scope

This affects original-mode tag caches such as `dynamodb-lite`. Next-mode tag caches do not maintain path/tag associations on set. Build-time page associations are already populated from generated metadata, so this change covers pages generated on demand and tags added during runtime regeneration.

Cache Components exposed the reported symptom, but the missing association is in incremental page-cache handling rather than the composable-cache adapter.

## Root cause

`revalidateTag()` correctly wrote the revalidation marker, but `Cache.updateTagsOnSet()` only persisted page-cache tag associations for the legacy `PAGE` kind. Runtime `APP_PAGE` entries therefore had no path/tag association for original-mode tag-cache lookup and CDN invalidation.

## Validation

- Biome: 440 files passed
- Unit tests: 611 passed, 1 todo
- `@opennextjs/aws` TypeScript build passed
- App Router and experimental example production builds passed; both new fixtures were classified as on-demand SSG routes
- Playwright discovered all affected App Router and experimental tests

### DIFF
--- a/.changeset/fix-page-cache-tag-associations.md
+++ b/.changeset/fix-page-cache-tag-associations.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Persist cache tag associations for `PAGES` and `APP_PAGE` entries so tag revalidation works with modern Next.js page cache kinds.
--- a/.changeset/fix-page-cache-tag-associations.md
+++ b/.changeset/fix-page-cache-tag-associations.md
@@ -2,4 +2,4 @@
 "@opennextjs/aws": patch
 ---
 
-Persist cache tag associations for `PAGES` and `APP_PAGE` entries so tag revalidation works with modern Next.js page cache kinds.
+Persist runtime cache tag associations for App Router `APP_PAGE` entries when using original-mode tag caches, including pages generated on demand and tags added after the build.
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -531,9 +531,7 @@ export default class Cache {
     // Write derivedTags to the tag cache
     // If we use an in house version of getDerivedTags in build we should use it here instead of next's one
     const pageCacheTags =
-      data?.kind === "PAGE" ||
-      data?.kind === "PAGES" ||
-      data?.kind === "APP_PAGE"
+      data?.kind === "PAGE" || data?.kind === "APP_PAGE"
         ? data.headers?.["x-next-cache-tags"]
         : undefined;
     const derivedTags: string[] =
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -530,13 +530,19 @@ export default class Cache {
     }
     // Write derivedTags to the tag cache
     // If we use an in house version of getDerivedTags in build we should use it here instead of next's one
+    const pageCacheTags =
+      data?.kind === "PAGE" ||
+      data?.kind === "PAGES" ||
+      data?.kind === "APP_PAGE"
+        ? data.headers?.["x-next-cache-tags"]
+        : undefined;
     const derivedTags: string[] =
       data?.kind === "FETCH"
         ? //@ts-expect-error - On older versions of next, ctx was a number, but for these cases we use data?.data?.tags
           (ctx?.tags ?? data?.data?.tags ?? []) // before version 14 next.js used data?.data?.tags so we keep it for backward compatibility
-        : data?.kind === "PAGE"
-          ? (data.headers?.["x-next-cache-tags"]?.split(",") ?? [])
-          : [];
+        : Array.isArray(pageCacheTags)
+          ? pageCacheTags
+          : (pageCacheTags?.split(",") ?? []);
     debug("derivedTags", derivedTags);
 
     // Get all tags stored in dynamodb for the given key
--- a/packages/open-next/src/adapters/cache.ts
+++ b/packages/open-next/src/adapters/cache.ts
@@ -554,6 +554,10 @@ export default class Cache {
     }
     // Write derivedTags to the tag cache
     // If we use an in house version of getDerivedTags in build we should use it here instead of next's one
+    // `x-next-cache-tags` is always a comma separated string: Next.js sets it from
+    // `metadata.fetchTags`, which is a string. The header type allows an array because it
+    // is shared with headers that can legitimately repeat, so we narrow on `typeof` here.
+    // Next.js guards the same way when it reads the header back.
     const pageCacheTags =
       data?.kind === "PAGE" || data?.kind === "APP_PAGE"
         ? data.headers?.["x-next-cache-tags"]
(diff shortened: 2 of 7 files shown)


packages/open-next/src/adapters/cache.ts
```diff
@@ -530,13 +530,19 @@ export default class Cache {
     }
     // Write derivedTags to the tag cache
     // If we use an in house version of getDerivedTags in build we should use it here instead of next's one
+    const pageCacheTags =
+      data?.kind === "PAGE" ||
+      data?.kind === "PAGES" ||
```
conico974: You cannot use `revalidateTag` with `PAGES` (aka page router pages). You can drop it

.changeset/fix-page-cache-tag-associations.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+Persist cache tag associations for `PAGES` and `APP_PAGE` entries so tag revalidation works with modern Next.js page cache kinds.
```
conico974: This is limited to the original tag cache mode. It also only applies to page that were not present at build time or to tags that were added afterwards.

conico974 (commented):
Could you add e2e tests (both in app-router and in experimental for `use cache`) , this has nothing to do with the composable cache actually.

packages/open-next/src/adapters/cache.ts
cgcompassion: Agreed. I removed `PAGES` from the implementation and unit matrix. `PAGE` remains for older App Router entries, while `APP_PAGE` covers modern App Router runtime entries.

.changeset/fix-page-cache-tag-associations.md
cgcompassion: Thanks for clarifying the scope. I narrowed the changeset and PR description to original-mode tag caches and runtime associations for pages generated on demand or tags added after the build. The description now also notes that build-time associations are populated separately.

cgcompassion:
Added the requested E2E regressions in 510ca8c. Both use dynamic paths with empty `generateStaticParams()` so the `APP_PAGE` entries and their tag associations are created at runtime rather than prepopulated from build metadata: one covers App Router `revalidateTag`, and one covers experimental `use cache`. Both fixture apps build successfully as on-demand SSG routes, and Playwright discovers the new cases. I also removed `PAGES` and clarified that the defect is in incremental page-cache association handling, not the composable-cache adapter.

packages/open-next/src/adapters/cache.ts
```diff
@@ -530,13 +530,17 @@ export default class Cache {
...
+        ? data.headers?.["x-next-cache-tags"]
+        : undefined;
     const derivedTags: string[] =
       data?.kind === "FETCH"
         ? //@ts-expect-error - On older versions of next, ctx was a number, but for these cases we use data?.data?.tags
           (ctx?.tags ?? data?.data?.tags ?? []) // before version 14 next.js used data?.data?.tags so we keep it for backward compatibility
-        : data?.kind === "PAGE"
-          ? (data.headers?.["x-next-cache-tags"]?.split(",") ?? [])
-          : [];
+        : Array.isArray(pageCacheTags)
```
vicb: Could `pageCacheTags` ever be an array at this point?

vicb (commented):
I added one inline comments

I also merged the main branch (to resolve a conflict) and suggest changes in unit tests and e2e as additional commit (with the help of Claude Opus).

@cgcompassion and @conico974 could you please take a look and let me know what you think.

Thanks!

packages/open-next/src/adapters/cache.ts
conico974: I don't think it can be no, you're right. It's not a "real" headers here

vicb: Adding a last commit to address this

vicb (approved):
Changes LGTM with a few additions.

I'll wait on @conico974 to also approve before merging

Thanks @cgcompassion and @conico974 🙏

conico974 (approved):

PR #1220, opened 2026-08-24
fix: fix `getUrlParts` when no pathname

### LINKED ISSUE #1217: getUrlParts returns the whole URL as `pathname` when a destination is `/` plus a query
Ran into this after moving a Next app onto Workers (`@opennextjs/cloudflare`, which pulls this package in). A redirect that had been fine on our previous host started returning 500 on every request to that path.

## Repro

```js
// next.config.js
module.exports = {
  async redirects() {
    return [{ source: "/promo/:path*", destination: "/?ref=promo", permanent: true }];
  },
};
```

`GET /promo/anything` → 500, with `Unexpected MODIFIER at 1, expected END` out of path-to-regexp:

```
Error in routingHandler
  at mustConsume
  at parse
  at compile
  at handleRewrites
  at handleRedirects
  at routingHandler
```

## Cause

https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/core/routing/util.ts#L62-L72

```ts
const regex = /\/([^?]*)\??(.*)/;
const match = url.match(regex);
return {
  hostname: "",
  pathname: match?.[1] ? `/${match[1]}` : url,
  ...
```

For `/?ref=promo` the first capture group is an empty string, which is falsy, so `pathname` falls back to the full url instead of `/`. `handleRewrites` then calls `compile(escapeRegex(pathname, { isPath: true }))` and path-to-regexp chokes on the `?` because there's no preceding token for it to modify. Since it throws inside the routing handler, the whole path 500s rather than just skipping the rule.

Only destinations whose path is exactly `/` hit this. `/terms?type=x` is fine because the group captures `terms`. Out of 57 destinations in our config the only one that broke was the single one pointing at the root with a query. `handleRewrites` is shared, so a rewrite with the same destination shape should hit it too, though I only saw it via redirects.

## Expected

Next's own `prepare-destination` runs the destination through `parseUrl()` and compiles just the pathname, so `/?ref=promo` comes out as pathname `/` + query `ref=promo`. The identical config returns a 308 on a plain Next server. A query in the destination is a documented shape too — the redirects docs have `destination: '/home?authorized=:authorized'` — so this doesn't look like an unsupported case, just the manual parse diverging from Next's.

## Suggested fix

```diff
-      pathname: match?.[1] ? `/${match[1]}` : url,
+      pathname: match ? `/${match[1]}` : url,
```

`match[1]` is always a string when `match` is non-null, so the empty case becomes `/` and every other shape is untouched. I've been running this as a yarn patch and it clears the 500 without changing behaviour for any of the other destinations we have.

Happy to open a PR if that diff looks reasonable.

## Versions

- `@opennextjs/aws` 4.1.0 (via `@opennextjs/cloudflare` 1.20.2)
- `next` 16.2.12

vicb:
fixes https://github.com/opennextjs/opennextjs-aws/issues/1217

### DIFF
--- a/.changeset/url-parts-query-only-destination.md
+++ b/.changeset/url-parts-query-only-destination.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix `getgetUrlParts()` when the path name is empty (root)
--- a/packages/open-next/src/core/routing/util.ts
+++ b/packages/open-next/src/core/routing/util.ts
@@ -56,6 +56,16 @@ export function convertFromQueryString(query: string) {
 }
 
 /**
+ * Splits an URL into its parts
+ *
+ * The URL is a route destination, i.e. a `path-to-regexp` pattern rather than a
+ * well formed URL, so it can not be parsed with `new URL()` which would percent
+ * encode some of the pattern characters.
+ *
+ * @param url The URL to split
+ * @param isExternal Whether the URL points to an external host
+ * @returns The protocol, hostname, pathname and query string of the URL
+ * @throws When `isExternal` is true and the URL is not an absolute HTTP(S) URL
  *
  * @__PURE__
  */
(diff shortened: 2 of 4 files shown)


conico974 (approved):
Thanks @vicb

vicb:
Thanks for the review Nico, I pushed a little fixup...

PR #1226, opened 2026-08-26
fix: forward images.dangerouslyAllowLocalIP to fetchExternalImage on Next 16

### LINKED ISSUE #1225: images.dangerouslyAllowLocalIP is not forwarded to fetchExternalImage on Next 16 — private-IP upstreams always rejected
### Bug

Next.js 16's image optimizer rejects upstream image hosts that resolve to private IPs unless the user opts out via [`images.dangerouslyAllowLocalIP`](https://nextjs.org/docs/app/api-reference/components/image#dangerouslyallowlocalip). `fetchExternalImage` takes that value as its second argument on Next 16 (see the signature table in `image-optimization.replacement.ts`).

The `optimizeImage` wrapper hardcodes that argument to `false`:

```ts
// v16.1.5+
if (compareSemver(version, ">=", "16.1.5")) {
  return fetchExternalImage(href, false, maximumResponseBody);
}

// v16.0.0-v16.1.4
if (compareSemver(version, ">=", "16")) {
  return fetchExternalImage(href, false);
}
```

so the user's `images.dangerouslyAllowLocalIP: true` never reaches the check, and every external image served from a private-IP host fails with:

```
ERROR ⨯ upstream image http://proxy.<internal-host>/media/....jpg hostname resolved to private IP ["10.0.x.x"] If this is expected and you understand SSRF risk, use images.dangerouslyAllowLocalIP = true to continue.
ERROR Failed to optimize image [Error]: "url" parameter is not allowed { statusCode: 400 }
```

The value is correctly present in `.next/required-server-files.json` (verified) and in the merged `nextConfig` the wrapper already receives — it is just not forwarded.

### Setup where this bites

- `@opennextjs/aws` 4.1.1, Next.js 16.3.3, deployed on Lambda
- Product/media images served through an internal reverse proxy (`http://proxy.<zone>.internal/media/...`, PrivateLink/VPC endpoint to the commerce backend) — the hostname legitimately resolves to a private IP
- `images.remotePatterns` allows the internal host and `images.dangerouslyAllowLocalIP: true` is set

### Expected

`fetchExternalImage` should receive `nextConfig.images?.dangerouslyAllowLocalIP ?? false` so the documented Next.js opt-out keeps working when deployed through OpenNext.

PR with the one-line forward incoming.

lcordier42:
Fixes #1225

Next.js 16's image optimizer rejects upstream hosts resolving to private IPs unless the user opts out via `images.dangerouslyAllowLocalIP`. The `optimizeImage` wrapper hardcoded that `fetchExternalImage` argument to `false`, so the user's config never reached the check and external images served from private-IP hosts (e.g. an internal media proxy behind a PrivateLink/VPC endpoint) always failed with a 400, even with the opt-out set.

The value is already present in the merged `nextConfig` the wrapper receives (loaded from `required-server-files.json`) — this PR forwards it:

- v16.1.5+: `fetchExternalImage(href, dangerouslyAllowLocalIP, maximumResponseBody)`
- v16.0.0–v16.1.4: `fetchExternalImage(href, dangerouslyAllowLocalIP)`

Pre-16 signatures are untouched (the parameter does not exist there). Default stays `false` (`?? false`), so behavior only changes for users who explicitly set the flag.

Verified on a real deployment (Next 16.3.3, `@opennextjs/aws` 4.1.1 on Lambda, media through an internal proxy resolving to 10.x): with the value forwarded, the previously failing `/_next/image?url=http%3A%2F%2Fproxy...` requests return `200 image/jpeg`; without it, `400 "url" parameter is not allowed`.

Includes a patch changeset.

### DIFF
--- a/packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
+++ b/packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
@@ -44,13 +44,13 @@ export async function optimizeImage(
   // Next 16 rejects upstream hosts resolving to private IPs unless the user
   // opts out with `images.dangerouslyAllowLocalIP` — forward their setting
   // instead of always disabling the opt-out.
-  const dangerouslyAllowLocalIP =
-    nextConfig.images?.dangerouslyAllowLocalIP ?? false;
+  const dangerouslyAllowLocalIP = Boolean(
+    nextConfig.images?.dangerouslyAllowLocalIP,
+  );
 
   const callFetchExternalImage = () => {
     const version = globalThis.nextVersion;
 
-    // v16.1.5+
     if (compareSemver(version, ">=", "16.1.5")) {
       return fetchExternalImage(
         href,
@@ -59,13 +59,11 @@ export async function optimizeImage(
       );
     }
 
-    // v16.0.0–v16.1.4
     if (compareSemver(version, ">=", "16")) {
       // @ts-expect-error - fetchExternalImage signature varies across Next.js versions
       return fetchExternalImage(href, dangerouslyAllowLocalIP);
     }
 
-    // v15.5.10–v15.x
     if (compareSemver(version, ">=", "15.5.10")) {
       // @ts-expect-error - fetchExternalImage signature varies across Next.js versions
       return fetchExternalImage(href, maximumResponseBody);
--- a/packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
+++ b/packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
@@ -41,18 +41,28 @@ export async function optimizeImage(
    *
    * https://github.com/vercel/next.js/blob/bfe2ab4/packages/next/src/server/image-optimizer.ts#L711
    */
+  // Next 16 rejects upstream hosts resolving to private IPs unless the user
+  // opts out with `images.dangerouslyAllowLocalIP` — forward their setting
+  // instead of always disabling the opt-out.
+  const dangerouslyAllowLocalIP =
+    nextConfig.images?.dangerouslyAllowLocalIP ?? false;
+
   const callFetchExternalImage = () => {
     const version = globalThis.nextVersion;
 
     // v16.1.5+
     if (compareSemver(version, ">=", "16.1.5")) {
-      return fetchExternalImage(href, false, maximumResponseBody);
+      return fetchExternalImage(
+        href,
+        dangerouslyAllowLocalIP,
+        maximumResponseBody,
+      );
     }
 
     // v16.0.0–v16.1.4
     if (compareSemver(version, ">=", "16")) {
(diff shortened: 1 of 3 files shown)


packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
```diff
@@ -41,18 +41,28 @@ export async function optimizeImage(
...
+        dangerouslyAllowLocalIP,
+        maximumResponseBody,
+      );
     }
 
     // v16.0.0–v16.1.4
     if (compareSemver(version, ">=", "16")) {
       // @ts-expect-error - fetchExternalImage signature varies across Next.js versions
-      return fetchExternalImage(href, false);
+      return fetchExternalImage(href, dangerouslyAllowLocalIP);
```
vicb, line 64: nit: could you please remove comments on l53, 62, 68 - they are redundant with the code and easily get out of sync

packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
```diff
@@ -41,18 +41,28 @@ export async function optimizeImage(
    *
    * https://github.com/vercel/next.js/blob/bfe2ab4/packages/next/src/server/image-optimizer.ts#L711
    */
+  // Next 16 rejects upstream hosts resolving to private IPs unless the user
+  // opts out with `images.dangerouslyAllowLocalIP` — forward their setting
+  // instead of always disabling the opt-out.
+  const dangerouslyAllowLocalIP =
+    nextConfig.images?.dangerouslyAllowLocalIP ?? false;
```
vicb:
```suggestion
  const dangerouslyAllowLocalIP =
    Boolean(nextConfig.images?.dangerouslyAllowLocalIP);
```

vicb (commented):
Thanks for the PR!

I have 2 inline nits.

It would be nice to add a test if possible.

packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
lcordier42: Done in c88505c.

packages/open-next/src/adapters/plugins/image-optimization/image-optimization.replacement.ts
lcordier42: Done — removed the three version comments (kept the `<=v15.5.9` one on the fallback since it has no adjacent condition). Also added unit tests in `tests-unit/tests/adapters/image-optimization.test.ts` covering the forwarding on both v16 branches (19 tests pass locally).

vicb (commented):
Claude added a test for 15.x (see my last commit).

Let me know if this looks ok to you.

Your changes look good so that's the only question to answer before being able to merge.

Thx

sommeeeer (approved):
LGTM!

PR #1227, opened 2026-08-26
fix: do not cache 404 and 500 responses served by the cache interceptor

vicb:
Next.js writes notFound() results for ISR routes to the incremental cache, so
the cache interceptor can serve them. It derived cache-control from the entry's
revalidate value alone and never looked at the status, so a route declaring no
revalidate was treated as SSG and its cached 404 was sent to the CDN with
s-maxage=31536000. A transient 404 - a row not yet replicated, a slug created a
moment later - was then sticky at the edge for a year.

The server path already forbids this in OpenNextNodeResponse.fixHeadersForError,
but the interceptor short-circuits routingHandler and returns an InternalResult
directly, so that override never runs. Apply it in the interceptor too, after the
response headers are assembled so that it also wins over a cache-control stored
in the cached entry's own headers.

Only 404 and 500 are overridden, as on the server path: other error statuses come
from the application, which owns their cache headers.
OPEN_NEXT_DANGEROUSLY_SET_ERROR_HEADERS=true still opts out. The no-store literal
and that env gate were duplicated in four places and now live in
utils/cacheControl.ts.

Note: Claude caught the issue

### DIFF
--- a/.changeset/cache-interception-error-cache-control.md
+++ b/.changeset/cache-interception-error-cache-control.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/aws": patch
+---
+
+Stop the cache interceptor from serving cached 404 and 500 responses with a cacheable `cache-control`.
+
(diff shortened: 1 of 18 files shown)


conico974 (approved):
LGTM Thanks 🎉

sommeeeer (approved):
LGTM

vicb:
Thanks for the reviews!

PR #1243, opened 2026-09-10
fix: mark sharp as external when bundling next-server

### LINKED ISSUE #1238: bundleNextServer.js esbuild externals list doesn't include sharp, breaking builds when it's a resolvable dependency
### Version
@opennextjs/aws@4.1.4 (used via @opennextjs/cloudflare@1.20.6)

### Description
`dist/build/bundleNextServer.js` bundles `next/dist/esm/server/next-server.js` with esbuild using a hardcoded `externals` array. That array does not include `sharp`, but Next core statically references `sharp` from `next-server.js`'s module graph regardless of whether image optimization is actually used at runtime (e.g. with `images.unoptimized: true`).

When `sharp` is resolvable in node_modules (which it usually is, since `next` lists it as an optional/transitive dependency, or it gets pulled in another way — in our case via a dependency chain), the build fails because esbuild tries to bundle sharp's native `.node` binaries:

    X [ERROR] Could not resolve require("../src/build/Release/sharp-*-*.node")
    X [ERROR] No loader is configured for ".node" files: .../sharp-win32-x64/lib/sharp-win32-x64-0.35.4.node

There's no `open-next.config.ts` or `next.config.ts` hook (`serverExternalPackages` doesn't apply — that only affects Next's own bundler, not this separate esbuild pass) to extend this externals list from userland.

### Reproduction
1. Have `sharp` resolvable in `node_modules` (any reason — direct dep, transitive, whatever).
2. `opennextjs-cloudflare build` (or the equivalent aws-adapter build).
3. Build fails during "Building server function: default" with the errors above.

### Suggested fix
Add `"sharp"` to the hardcoded `externals` array in `bundleNextServer.js`, or (better) expose a way for adapters/users to extend it, since Cloudflare's own OpenNext image handling (`env.IMAGES` binding) never actually calls sharp at runtime — it's genuinely dead code there but still needs to not blow up the build.

### Workaround
We're currently carrying a local `pnpm patch` on `@opennextjs/aws@4.1.4` that adds `"sharp"` to this array.

cpruijsen:
Add `"sharp"` to the hardcoded `externals` array in `bundleNextServer` so esbuild leaves `sharp` as a runtime `require`.

`bundleNextServer` bundles `next/dist/esm/server/next-server.js`. Next's compiled `image-optimizer` still does `require("sharp")` from that graph, even when image optimization is unused. If `sharp` is resolvable (including as Next's optional dependency), the build fails with `Could not resolve require("../src/build/Release/sharp-*-*.node")` and `No loader is configured for ".node" files`. The same exclusion already exists for traced server files and for the image-optimization bundle.

Fixes #1238

## Decision

The other option is a user/adapter hook to extend esbuild externals, or a generic `.node` plugin. On #1116 a maintainer said esbuild here is meant for OpenNext's own code, not a new config surface for Next/user packages. The reporter's working workaround is this list entry. Can switch to a `.node` plugin or a narrow extra-externals option.

## Test plan

- [x] Unit test `bundleNextServer` with a fixture `next-server` that `require("sharp")`s a dummy `.node` file: build succeeds, output keeps `require("sharp")`, native bytes are not inlined.
- [ ] With `experimentalBundledNextServer: true` and `sharp` resolvable in `node_modules`, `open-next build` (or the Cloudflare adapter build that calls `bundleNextServer`) completes the "Building server function: default" step without `.node` loader errors.
- [ ] Default path (`experimentalBundledNextServer` unset) still traces/excludes `sharp` as before; image-optimization function still installs sharp separately.

The last two were not run.

### DIFF
--- a/.changeset/externalize-sharp-next-server.md
+++ b/.changeset/externalize-sharp-next-server.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: mark `sharp` as external when bundling `next-server`
--- a/packages/open-next/src/build/bundleNextServer.ts
+++ b/packages/open-next/src/build/bundleNextServer.ts
@@ -40,6 +40,10 @@ const externals = [
   // We need to remove this since this is what webpack is building
   // Adding it cause to add a lot of unnecessary deps
   "next/dist/compiled/next-server",
+
+  // Next's image-optimizer does `require("sharp")`. When sharp is installed,
+  // esbuild tries to bundle its native `.node` binaries and the build fails.
+  "sharp",
 ];
 
 export async function bundleNextServer(
--- a/packages/tests-unit/tests/build/bundleNextServer.test.ts
+++ b/packages/tests-unit/tests/build/bundleNextServer.test.ts
@@ -1,0 +1,91 @@
+import {
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+
+import { bundleNextServer } from "@opennextjs/aws/build/bundleNextServer.js";
+import { afterEach } from "vitest";
+
+function write(file: string, content: string) {
+  mkdirSync(path.dirname(file), { recursive: true });
+  writeFileSync(file, content);
+}
+
+/**
+ * Next's compiled `next-server` graph includes a bare `require("sharp")` in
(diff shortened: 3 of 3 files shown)


sommeeeer (approved):
LGTM

conico974 (commented):
I'm not sure if I want to merge this one TBH, `experimentalBundledNextServer` is deprecated and very likely doesn't work anymore. It's a patch over something that likely doesn't work and that we will definitely not support (adapters-api is the way to go for that)

PR #1244, opened 2026-09-10
fix: quote cache interceptor ETag values as HTTP entity-tags

### LINKED ISSUE #1183: [BUG] Cache interceptor returns etag with incorrect syntax
The response header for the etag in `packages/open-next/src/core/routing/cacheInterceptor.ts` returns an etag of an incorrect syntax.
`cacheInterceptor.ts` is used when the following config is enabled: 
```js
const config = {
  default: {},
  dangerous: {
    enableCacheInterception: true,
  },
} satisfies OpenNextConfig;
```

Currently it returns `etag: <etag_value>`, while the correct syntax requires `<etag_value>` to be surrounded by double quotes: i.e. `etag: "<etag_value>"`.
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag#syntax

This causes issues with Cloudfront CDN compression, as it will remove invalid etags after the compression.
[See CloudFront Serve compressed files](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#compressed-content-cloudfront-etag-header:~:text=When%20the%20object%20from%20the%20origin%20includes%20an%20invalid%20ETag%20header%20value%20(the%20value%20does%20not%20begin%20with%20%22%20or%20with%20W/)%2C%20CloudFront%20removes%20the%20ETag%20header%20and%20returns%20the%20object%20to%20the%20viewer%20without%20the%20ETag%20response%20header.)

**Suggested fix**
[`packages/open-next/src/core/routing/cacheInterceptor.ts#L61`](https://github.com/opennextjs/opennextjs-aws/blob/d35f7cc1e1acfa7fba5130e63495cfa1cda80132/packages/open-next/src/core/routing/cacheInterceptor.ts#L61) creates the etag value as: 
```js
const etag = hash(body);
```
Instead it should be replaced with the correct etag syntax:
```js
const etag = `"${hash(body)}"`
```

The suggested fix has been tested through patching locally and deployment.

Environment: sst@4.14.3, @opennextjs/aws@4.0.3, next@16.2.6

cpruijsen:
Quote the interceptor-computed value at the existing assignment in `computeCacheControl`: `etag: "<md5>"`. With `dangerous.enableCacheInterception`, the interceptor served `etag: <md5>` (raw hex). RFC 9110 requires a quoted entity-tag, and CloudFront removes ETags that do not begin with `"` or `W/` when it compresses the object, so clients never see a validator. The reporter already tested the patch from #1183 via a local patch and a deployment (sst@4.14.3, @opennextjs/aws@4.0.3, next@16.2.6).

Fixes #1183

## Decision

The other option is to quote only the HTTP header and leave the SQS `eTag` / deduplication id as the raw hash. One assignment feeds the response header, the revalidation queue body, and the dedup id. Queue implementations only stringify / hash that field; the revalidate worker does not read `eTag`. Quoting at the assignment is the smallest change. Dedup ids will not collide with in-flight pre-upgrade messages, so a rolling deploy can fire one extra revalidation in the 5-minute SQS window.

Can switch to header-only quoting if you would rather leave the queue payload unchanged.

## Test plan

- [x] Unit test fails without the quotes (`6cd3556deb0da54bca060b4c39479839`) and passes with them (`"6cd3556deb0da54bca060b4c39479839"`).
- [x] Full `cacheInterceptor` unit file: 54 passed.
- [ ] Optional: deploy with `enableCacheInterception: true` in front of CloudFront compression and confirm the viewer still receives `etag: "<hex>"`.

The CloudFront deploy check was not run.

### DIFF
--- a/.changeset/quote-cache-interceptor-etag.md
+++ b/.changeset/quote-cache-interceptor-etag.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: quote cache interceptor ETag values so they are valid HTTP entity-tags
--- a/packages/open-next/src/core/routing/cacheInterceptor.ts
+++ b/packages/open-next/src/core/routing/cacheInterceptor.ts
@@ -65,7 +65,9 @@ async function computeCacheControl(
   // calculate age
   const age = Math.round((Date.now() - (lastModified ?? 0)) / 1000);
   const hash = (str: string) => createHash("md5").update(str).digest("hex");
-  const etag = hash(body);
+  // RFC 9110 entity-tag is a quoted opaque-tag. CloudFront drops unquoted ETags
+  // when serving compressed objects.
+  const etag = `"${hash(body)}"`;
   if (revalidate === 0) {
     // This one should never happen
     return {
--- a/packages/tests-unit/tests/core/routing/cacheInterceptor.test.ts
+++ b/packages/tests-unit/tests/core/routing/cacheInterceptor.test.ts
@@ -1,4 +1,6 @@
 /* eslint-disable sonarjs/no-duplicate-string */
+import { createHash } from "node:crypto";
+
 import { NextConfig } from "@opennextjs/aws/adapters/config/index.js";
 import { cacheInterceptor } from "@opennextjs/aws/core/routing/cacheInterceptor.js";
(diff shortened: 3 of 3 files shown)


conico974 (approved):
LGTM Thanks for the PR

PR #1245, opened 2026-09-11
fix: scope composable cache pending writes to the request, not the isolate

### LINKED ISSUE #1218: [BUG] composable cache shares pending write promises across requests (module-scope map + multi-request isolates)
### Summary

`adapters/composable-cache.ts` keeps in-flight write promises in a **module-scope** `Map`:

```ts
const pendingWritePromiseMap = new Map<string, Promise<CacheValue<"composable">>>();

async get(cacheKey: string) {
  // We first check if we have a pending write for this cache key
  // If we do, we return the pending promise instead of fetching the cache
  if (pendingWritePromiseMap.has(cacheKey)) {
    const stored = pendingWritePromiseMap.get(cacheKey);
    if (stored) {
      return stored.then((entry) => ({ ...entry, value: toReadableStream(entry.value) }));
    }
  }
  ...
```

Module scope is per-**isolate**, not per-request. On Lambda those are the same thing — one request per sandbox — so this is safe there. On `workerd` (`@opennextjs/cloudflare`) one isolate serves many concurrent requests, so **request A's pending write promise is handed to request B**, and B awaits a promise created in A's request context.

Cloudflare's runtime says as much when it happens: *"A promise was resolved or rejected from a different request context than the one it was created in."*

### Why it matters

When A's request context is torn down before its write settles — a client disconnect is enough — that promise can never settle. Three things follow:

1. **B hangs with no upper bound.** The invocation is held until the runtime kills it at the 128 MB memory limit. The signature is minutes of wall time against milliseconds of CPU.
2. **`.finally()` never runs**, so `pendingWritePromiseMap.delete(cacheKey)` never happens.
3. **The key stays poisoned for the isolate's lifetime.** Every subsequent request in that isolate asking for the same key attaches to the same dead promise.

Item 3 is the reason this is worth fixing even though it is rare: the failure is not self-limiting.

This is reachable from every `'use cache'`, since `@opennextjs/cloudflare` patches Next's `loadCustomCacheHandlers` to register this adapter as both the `default` and `remote` handler.

It also cannot be worked around downstream. The handler is injected by a build-time AST patch, so there is no override seam — unlike `incrementalCache`, a user-supplied wrapper has nowhere to attach. And a wrapper would not help anyway: the pending-write fast path in `get` returns *before* `globalThis.incrementalCache` is ever called, so a timeout on the incremental cache sits below the hazard.

### Reproduction

Runs against the published package with `vitest`. No AWS or Cloudflare account needed.

```ts
import { AsyncLocalStorage } from "node:async_hooks";
import composableCache from "@opennextjs/aws/adapters/composable-cache.js";
import { runWithOpenNextRequestContext } from "@opennextjs/aws/utils/promise.js";
import { beforeEach, expect, it, vi } from "vitest";

const KEY = "some-cache-key";
/** A write whose request context is torn down: it never settles. */
const NEVER = new Promise<never>(() => {});

const incrementalCache = { get: vi.fn(), set: vi.fn() };

const inRequest = <T>(requestId: string, fn: () => Promise<T>) =>
  runWithOpenNextRequestContext({ isISRRevalidation: false, requestId }, fn);

beforeEach(() => {
  const g = globalThis as any;
  g.__openNextAls = new AsyncLocalStorage();
  g.incrementalCache = incrementalCache;
  g.tagCache = { mode: "nextMode" };
  g.openNextConfig = { dangerous: {} };
});

it("does not hand one request a pending write promise from another", async () => {
  // Request A starts a write that never completes, then goes away.
  void inRequest("A", () => composableCache.set(KEY, NEVER));
  await new Promise((r) => setTimeout(r, 10));

  incrementalCache.get.mockResolvedValue({
    value: { value: "from-the-store", tags: [], revalidate: 60 },
    lastModified: Date.now(),
  });

  // Request B asks for the same key. It must reach the store, not attach to A.
  const result = await Promise.race([
    inRequest("B", () => composableCache.get(KEY)),
    new Promise((r) => setTimeout(() => r("HUNG"), 250)),
  ]);

  expect(result).not.toBe("HUNG"); // fails on 4.1.0
});
```

On `4.1.0` this reports `HUNG`; with the fix below it passes. A third context with no request at all also hangs on that key afterwards, which is item 3.

One setup note if you run it against the published package rather than in-repo: the dist ESM uses extensionless relative imports (`../utils/cache`), which Node's ESM resolver won't follow, so vitest needs

```ts
test: { server: { deps: { inline: [/@opennextjs[\\/]aws/] } } }
```

### Suggested fix

The repo already ships the right primitive — `utils/requestCache.ts`, whose own doc comment says *"Overrides can use this to store and share data within the scope of a single request without polluting global state"* — and the DynamoDB, D1, KV and DO-sharded tag caches already use it. The composable cache just isn't.

```diff
-const pendingWritePromiseMap = new Map<string, Promise<CacheValue<"composable">>>();
+const PENDING_WRITES_CACHE_KEY = "composable:pendingWrites";
+const getPendingWritePromiseMap = () =>
+  globalThis.__openNextAls?.getStore()?.requestCache?.getOrCreate<
+    string,
+    Promise<CacheValue<"composable">>
+  >(PENDING_WRITES_CACHE_KEY);

 async get(cacheKey: string) {
   try {
+    const pendingWritePromiseMap = getPendingWritePromiseMap();
-    if (pendingWritePromiseMap.has(cacheKey)) {
+    if (pendingWritePromiseMap?.has(cacheKey)) {

 async set(cacheKey, pendingEntry) {
   const promiseEntry = pendingEntry.then(...);
+  const pendingWritePromiseMap = getPendingWritePromiseMap();
-  pendingWritePromiseMap.set(cacheKey, promiseEntry);
+  pendingWritePromiseMap?.set(cacheKey, promiseEntry);
   const entry = await promiseEntry.finally(() => {
-    pendingWritePromiseMap.delete(cacheKey);
+    pendingWritePromiseMap?.delete(cacheKey);
   });
```

This keeps the read-coherence the map exists for — a `get` still sees a write in flight from the same render — while making it impossible for two request contexts to share a promise. With no ALS store (build time) the dedupe is skipped and `get` falls through to the incremental cache.

Happy to open a PR.

### Severity — measured

We shipped this fix with a counter recording what the module-scope map *would* have done. Over 52,994 production requests on Cloudflare (`workerd`):

- **253** isolates confirmed running the instrumented path — the positive control, so a zero below is meaningful rather than dark.
- **0** cross-request shares.
- `owners=1` on every stats line: one composable-cache write in flight per key per isolate, so the overlap this needs rarely opens.

Upper bound on exposure ≈ 1 in 17,700 requests. Re-checked four days later: still zero, control still firing.

On our workload this is therefore a **latent correctness bug, not a hot one**, and it does not explain memory kills — ours were application-side, and fixing them moved us from 4.60 to 0.08 `exceededMemory` per 10k with this patch already in place and doing nothing. Anyone arriving here from a hang is probably looking at a different cause.

The case for fixing it is item 3 above: the failure is not self-limiting. One occurrence poisons that cache key for the isolate's lifetime, because `.finally()` never runs on a promise whose request context was torn down. A hotter or more concentrated key distribution than ours would hit it far more often.

### Environment

- `@opennextjs/aws` 4.1.0 (current `main` is unchanged)
- `@opennextjs/cloudflare` 1.20.2
- Next.js 16.3.1
- Cloudflare Workers (`workerd`), 128 MB isolate limit

Related: opennextjs/opennextjs-cloudflare#1350 (three unbounded awaits in the Cloudflare overrides — a different defect in the same request path).

isaacrowntree:
Fixes #1218.

`adapters/composable-cache.ts` kept in-flight `'use cache'` write promises in a module-scope `Map`. Module scope is per isolate, so on a runtime that serves many concurrent requests from one isolate (Cloudflare Workers via `@opennextjs/cloudflare`) request A's pending write promise could be handed to request B. If A's context was torn down before the write settled, B awaited a promise that never resolved, `.finally()` never ran, and the key stayed poisoned for the life of the isolate. Full analysis and measurements are in the issue.

### Change

The map now lives in the per-request `RequestCache` (`utils/requestCache.ts`), which the DynamoDB tag caches already use for the same purpose. A `get` still sees a write in flight from the same render; two request contexts can no longer share a promise. With no ALS store (build time) the dedupe is skipped and `get` falls through to the incremental cache.

### Tests

- The existing "should return pending write promise if available" test now runs against a real `RequestCache` in the mock store.
- New: a write started in one request context is not returned to a `get` from another (the pre-fix behaviour hangs the test).
- New: with no request context, `get` falls through to the incremental cache instead of hanging.

Changeset included (patch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Tbi1NmauUu87Lp6m3dsPT1

### DIFF
--- a/.changeset/composable-cache-request-scope.md
+++ b/.changeset/composable-cache-request-scope.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: scope the composable cache's pending writes to the request, not the isolate
+
+The map of in-flight `'use cache'` writes was module-scoped. On a runtime that serves many concurrent requests from one isolate (e.g. Cloudflare Workers) a pending write promise from one request could be handed to another; if the first request's context was torn down before the write settled, the second hung and the key stayed poisoned for the life of the isolate. The map now lives in the per-request `RequestCache`.
--- a/packages/open-next/src/adapters/composable-cache.ts
+++ b/packages/open-next/src/adapters/composable-cache.ts
@@ -4,17 +4,39 @@ import { isStale, writeTags } from "utils/cache";
 import { fromReadableStream, toReadableStream } from "utils/stream";
 import { debug } from "./logger";
 
-const pendingWritePromiseMap = new Map<
-  string,
-  Promise<CacheValue<"composable">>
->();
+/**
+ * In-flight writes are tracked per request, not per module.
+ *
+ * A module-scope map is per isolate. On a runtime that serves one request per
+ * sandbox that is the same as per request, but on a runtime that serves many
+ * concurrent requests from one isolate (e.g. Cloudflare Workers) a pending
+ * write promise created in request A would be handed to request B. If A's
+ * context is torn down before the write settles, B awaits a promise that never
+ * resolves, `.finally()` never runs, and the key stays poisoned for the life of
+ * the isolate.
+ *
+ * `RequestCache` is the per-request store the tag caches already use for this.
+ * Scoping the map to it keeps the read-coherence a `get` needs for a write in
+ * flight from the same render while making a cross-request share impossible.
+ * With no ALS store (e.g. at build time) the dedupe is skipped and `get` falls
+ * through to the incremental cache.
+ */
+const PENDING_WRITES_CACHE_KEY = "composable-cache:pending-writes";
+
+const getPendingWritePromiseMap = () =>
+  globalThis.__openNextAls
+    ?.getStore()
+    ?.requestCache?.getOrCreate<string, Promise<CacheValue<"composable">>>(
+      PENDING_WRITES_CACHE_KEY,
+    );
 
 export default {
   async get(cacheKey: string) {
     try {
       // We first check if we have a pending write for this cache key
       // If we do, we return the pending promise instead of fetching the cache
-      if (pendingWritePromiseMap.has(cacheKey)) {
+      const pendingWritePromiseMap = getPendingWritePromiseMap();
+      if (pendingWritePromiseMap?.has(cacheKey)) {
         const stored = pendingWritePromiseMap.get(cacheKey);
         if (stored) {
           return stored.then((entry) => ({
@@ -91,10 +113,11 @@ export default {
       ...entry,
       value: await fromReadableStream(entry.value),
     }));
-    pendingWritePromiseMap.set(cacheKey, promiseEntry);
+    const pendingWritePromiseMap = getPendingWritePromiseMap();
+    pendingWritePromiseMap?.set(cacheKey, promiseEntry);
 
     const entry = await promiseEntry.finally(() => {
-      pendingWritePromiseMap.delete(cacheKey);
+      pendingWritePromiseMap?.delete(cacheKey);
     });
     await globalThis.incrementalCache.set(
       cacheKey,
--- a/packages/tests-unit/tests/adapters/composable-cache.test.ts
+++ b/packages/tests-unit/tests/adapters/composable-cache.test.ts
@@ -1,4 +1,5 @@
 import ComposableCache from "@opennextjs/aws/adapters/composable-cache";
+import { RequestCache } from "@opennextjs/aws/utils/requestCache.js";
 import {
   fromReadableStream,
   toReadableStream,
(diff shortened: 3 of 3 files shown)


conico974 (approved):
LGTM Thanks for the PR

Issue #753, opened 2025-02-26
How to handle host headers


vicb:
ref/discussion: see https://github.com/opennextjs/opennextjs-aws/pull/752

When OpenNext is behind a reverse proxy (i.e. aws wrappers) the host can be retrieved from `header["x-forwarded-host"]`.
For Node, it depends how the infra is setup (behind a RP or not).

We also have this:

https://github.com/opennextjs/opennextjs-aws/blob/e48951fcb5882b536e67581feac489cd1689ffe9/packages/open-next/src/core/requestHandler.ts#L46-L48

`header["x-forwarded-host"]` should not be trusted if not behind a reverse proxy as it can be forged.

We should figure out the best way to configure this.

PR #1246, opened 2026-09-11
fix: preserve existing x-forwarded-host in the edge converter

### LINKED ISSUE #1182: [BUG] edge converter overwrites x-forwarded-host with host
Relate: #753,#752

OpenNext currently overwrites `x-forwarded-host` with `result.internalEvent.headers.host`
https://github.com/opennextjs/opennextjs-aws/blob/d35f7cc1e1acfa7fba5130e63495cfa1cda80132/packages/open-next/src/overrides/converters/edge.ts#L62-L65

This means any existing `x-forwarded-host` is discarded and replaced with the current host value.

This causes problems in Cloudflare Worker:

user request header:
```
x-forwarded-host: foo
host: bar
```

plain worker view:
```
x-forwarded-host: foo
host: bar
```

opennextjs/cloudflare view:
```
x-forwarded-host: bar
host: bar
```

my use case:

better-auth use `x-forwarded-host` and `x-forward-proto` to determine baseURL when `trustedProxyHeaders` enabled

https://better-auth.com/docs/reference/security#trusted-proxy-headers

cpruijsen:
## Summary

The edge converter now keeps an existing `x-forwarded-host` when reconstructing a `Request` from a middleware result. If `x-forwarded-host` is missing, the converter still fills it from `host` (previous behavior).

The edge converter previously always replaced `x-forwarded-host` with `host`. Behind a reverse proxy that discarded the original host (Cloudflare Worker: `x-forwarded-host: foo` / `host: bar` became `x-forwarded-host: bar`).

Fixes #1182.

## Decision

Preserve `x-forwarded-host` when present and fall back to `host`, rather than stop setting `x-forwarded-host` entirely. Same precedence as `extractHostFromHeaders`, and the fallback keeps requests that never sent the header working as before. Can drop the fallback if the header should stay unset when it was never sent.

The `foo` / `bar` values are the reporter's on #1182. On #753, when OpenNext is behind a reverse proxy the host should come from `x-forwarded-host`. A trust/config flag for non-proxy setups belongs on #753, not this preserve.

## Test plan

- [x] Unit test: existing `x-forwarded-host` is not overwritten by `host`
- [x] Unit test: missing `x-forwarded-host` is still set from `host`
- [ ] Optional: Worker with better-auth `trustedProxyHeaders` still sees the original host

### DIFF
--- a/.changeset/preserve-x-forwarded-host.md
+++ b/.changeset/preserve-x-forwarded-host.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: preserve existing `x-forwarded-host` in the edge converter
--- a/packages/open-next/src/overrides/converters/edge.ts
+++ b/packages/open-next/src/overrides/converters/edge.ts
@@ -61,7 +61,9 @@ const converter: Converter<InternalEvent, InternalResult | MiddlewareResult> = {
         method: result.internalEvent.method,
         headers: {
           ...result.internalEvent.headers,
-          "x-forwarded-host": result.internalEvent.headers.host,
+          "x-forwarded-host":
+            result.internalEvent.headers["x-forwarded-host"] ??
+            result.internalEvent.headers.host,
         },
       });
 
--- a/packages/tests-unit/tests/converters/edge.test.ts
+++ b/packages/tests-unit/tests/converters/edge.test.ts
@@ -1,0 +1,53 @@
+import converter from "@opennextjs/aws/overrides/converters/edge.js";
+
+function middlewarePassThrough(headers: Record<string, string>) {
+  return {
+    type: "middleware" as const,
+    internalEvent: {
+      type: "core" as const,
+      method: "GET",
+      rawPath: "/",
+      url: "https://bar.example/",
+      headers,
+      query: {},
+      cookies: {},
+      remoteAddress: "::1",
+    },
+    isExternalRewrite: false,
+    origin: false,
+    isISR: false,
+    initialURL: "https://bar.example/",
+    resolvedRoutes: [],
+  };
+}
+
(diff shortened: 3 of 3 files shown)


conico974 (changes requested):
> header["x-forwarded-host"] should not be trusted if not behind a reverse proxy as it can be forged.

You quoted this PR and just discarded it, we should not do that without a proper way of handling it

cpruijsen:
Fair, and my description was too quick to push this to #753. Let me put the facts I checked on the table instead.

You're right that this widens the edge path. The converter's unconditional overwrite was acting as a sanitiser there, and preserving the header removes it, so in a non-proxy deployment a client can now get its own `x-forwarded-host` through where it previously got stomped.

What I do not think it does is introduce the trust decision. Two places already make it:

- `requestHandler.ts` lines 62-64 promote a client-supplied `x-forwarded-host` to `host` with no check. That is the exact code #753 links.
- `extractHostFromHeaders` in `converters/utils.ts` is `headers["x-forwarded-host"] ?? headers.host`, used by the node and apigw converters.

So on every non-edge path the forged header already wins today, and on the edge path it wins again at `requestHandler.ts:62`, whatever this converter sets. That is why I do not think a guard here would buy anything: it would be undone immediately downstream, while the other entry points stayed as they are.

Which points at the same conclusion #753 reaches, that the switch has to be central, read once where `host` is resolved, rather than per converter.

So: your call on this PR. I can gate the preserve behind a config flag here, though I would rather not add that surface before #753 settles its name and default. I can park this behind #753 and fold the `x-forwarded-host` handling into whatever lands there. Or I can close it and open the central fix instead, which would cover #1182 and the node path together.

I have no strong preference between those, and I would rather not guess given it is a security default.

PR #1248, opened 2026-09-13
Enhance cache interceptor for segment prefetching in Next 16

conico974:
Improve the cache interceptor to correctly serve segment prefetches by removing the gating condition on `prefetchInlining`. This change ensures that segment requests return the appropriate segment data instead of the full page payload, addressing an issue where prefetch requests were not being satisfied correctly. Additionally, update the type definition for `prefetchInlining` to reflect its new behavior.

### DIFF
--- a/.changeset/silly-doors-listen.md
+++ b/.changeset/silly-doors-listen.md
@@ -1,0 +1,25 @@
+---
+"@opennextjs/aws": patch
+---
+
+fix: serve segment prefetches from the cache interceptor on Next 16
+
+The segment response branch was gated on `!NextConfig.experimental.prefetchInlining`.
+Next normalizes every truthy `prefetchInlining` into `{ maxSize, maxBundleSize }` and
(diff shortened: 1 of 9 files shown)


eli-3030:
Confirming this against a live cloudflare workers deployment (Next 16.3.5, @opennextjs/cloudflare 1.20.6, opennextjs/opennextjs-cloudflare#1334).

Before: a single tab parked on the home page issued ~200 Next-Router-Segment-Prefetch: /_tree requests per second, indefinitely. Two tabs produced 537k worker requests in the first 24 hours.

With this PR's build (npm i https://pkg.pr.new/@opennextjs/aws@1248, no other changes) in wrangler dev: /_tree answers with the segment payload and x-nextjs-postponed: 2, and the same parked tab makes 5 RSC requests in 6 seconds, then goes quiet. Pages, 404 and route handlers unchanged.

